PR #3, opened 2021-02-04
Update CMake build settings

ewquon:
Tested on NREL Eagle HPC w/ Intel compilers v18.0.5.274


ewquon:
Updated options per @sayerhs's recommendation

PR #5, opened 2021-02-09
Turn on CI 

ewquon:
Tested this setup on my fork
* Build with cmake on ubuntu linux w/ gnu -- @sayerhs, do you think we need other systems for now?
* Only build 3D problem -- @asalmgren, should `ScalarAdvection` compile for DIM < 3? 
* Added problem matchers from PeleC setup (unused for now)

Should hold off on merge until #4 is resolved.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+          # Install MASA
+          cd ${{runner.workspace}}/deps
+          git clone --recursive https://github.com/manufactured-solutions/MASA.git ${{runner.workspace}}/deps/MASA
+          cd ${{runner.workspace}}/deps/MASA
+          ./bootstrap
+          ./configure CXX='mpicxx -std=c++11' CC=mpicc \
+            METAPHYSICL_DIR="${{runner.workspace}}/deps/install/MetaPhysicL" \
+            --prefix="${{runner.workspace}}/deps/install/MASA"
+          make -j ${{env.NPROCS}}
+          make install
```
sayerhs: I suspect these are not needed until you introduce MMS tests. Will speed up CI builds if you remove/comment this out for now.

sayerhs (commented):
I would also recommend you add [CUDA coverage](https://github.com/Exawind/amr-wind/blob/main/.github/workflows/cuda-ci.yml)

ewquon:
Note: Default is DIM=3; DIM < 3 is not supported.

PR #6, opened 2021-02-22
Create documentation for ERF Euler discretization on Arakawa C Ggrid

pkjha-aero:


ewquon (commented):
These figures look really nice and will be a good reference. I'm wondering why the equations weren't directly input as latex in the sphinx doc? That should make maintenance and paper writing in the future easier.

xyuan:
> These figures look really nice and will be a good reference. I'm wondering why the equations weren't directly input as latex in the sphinx doc? That should make maintenance and paper writing in the future easier.

Yes, the document is very well organized. I would suggest to further write the cell-interface/cell-center interpolation formula, so it is more straightforward to code.

pkjha-aero:
@ewquon and @xyuan We started from a PPT and word doc for these figures and equations. As a first pass I wanted to have these documented with equations as images taken from the existing docs. I'll update these equations to proper LaTeX format as in https://erf-pkj.readthedocs.io/en/latest/Euler_Discretization.html#diagnostic-variables.

xyuan (approved):

jmirocha:
Looks correct to me.

PR #7, opened 2021-02-22
Update submodule AMReX to the latest one

pkjha-aero:


ewquon:
We should decide how often we want to update the AMReX submodule, or if it makes sense to freeze it for now. Is there a specific version or bug fix that has been pulled in?

pkjha-aero:
@ewquon, I noticed several changes to AMReX code including some fixes (https://github.com/AMReX-Codes/amrex/commit/c143555e74e4096ec25400369651bf884b8561b3, https://github.com/AMReX-Codes/amrex/commit/eb9b4242a07028a77bd95986f5a5686c81dce428). That's why I wanted to update. Though these fixes may not affect us.
But, I agree that we should decide how often we want to update.

ewquon (approved):

ewquon:
Thanks for the info. I think this is fine, let's plan on freezing the submodule after we pass our next solver milestone.

xyuan (approved):
I agree with you

PR #13, opened 2021-03-25
Implement diffusion for cell-centered equations

zzwz02:
Implement diffusion term for density, temperature and advection scalar equations.


ewquon:
Let's rethink how to incorporate the diffusion terms on top of the latest changes to the code.

PR #20, opened 2021-06-01
Navier-Stokes Code Architecture

### LINKED ISSUE #14: add switches in inputs_ex and 3 benchmarking tests
Add the following switches in inputs_ex for controlling benchmarking tests:
- AdvectionTerm_density
- AdvectionTerm_theta
- AdvectionTerm_scalar
- PressureTerm
- AdvectionTerm_vel
- reverse time step

Add/improve three benchmarking tests:
- scalar advection
- scalar diffusion
- Taylor Green vortex

pkjha-aero:
This PR should work as overall code architecture. Individual capabilities such as Smagorinsky model will need to be updated.


zzwz02:
I would suggest that the codes should not be merged unless the following issues are resolved:
(1) The three new demos `IsentropicVortex`, `ScalarAdvecUniformU` and `ScalarAdvecUniformUV` need README files explaining the problem solved and the corresponding references (if any, such as [this](https://github.com/zzwz02/ERF/blob/BenchmarkExample/Exec/ScalarDiffusion/README) and [this](https://github.com/zzwz02/ERF/blob/BenchmarkExample/Exec/TaylorGreen/README)).
(2) Since the codes are changed quite a lot, the benchmarking test cases (Taylor green vortex, scalar Diffusion+Advection) have to be re-run to make sure the code is doing right.

pkjha-aero:
@zzwz02 , these are good suggestions. However, as @ewquon had mentioned earlier it is better to have these building blocks in place first and then add the ReadMe files and references in separate pull requests. Similarly, Separate PR for adding Pure  Diffusion, Advection Diffusion and Taylor Green test cases. The earlier benchmark test cases (Taylor green vortex, scalar Diffusion+Advection) have some hard coding. Therefore, it may not be easy to merge the latest code into those branches. It'll be better to deal with them separately, otherwise this PR will be extremely big. @ewquon @jmirocha may suggest please.

zzwz02:
My first concern (README in new demos) can be in separate pull requests. But my second concern is important. If these new codes are not validated, nobody would be able to use it -- at least, I can't; I can only work on validated code.

Anyway, it's all up to you.

Source/RK3/EddyViscosity.cpp
```diff
@@ -0,0 +1,115 @@
...
+  // Assuming we already have 'nut' computed for all (i, j, k)
+// TODO: Check exactly where we should compute 'nut'
+  Real turbViscInterpolated = 1.0;
+
+  switch (momentumEqn) {
+  case MomentumEqn::x:
+    switch (diffDir) {
+    case DiffusionDir::x:
+      if (nextOrPrev == NextOrPrev::next)
+        turbViscInterpolated = nut(i, j, k); // K (i, j, k) needed to obtain tau11 (i+1/2)
```
rfhaque, line 42: @pkjha-aero Is it possible to refactor this code and make a subroutine to precalculate the indexes before calling the nut routine at the end? This would allow reuse of the switch case logic at multiple places in the code.

rfhaque (approved):

Source/RK3/EddyViscosity.cpp
pkjha-aero: @rfhaque I'll take care of this in future PR.

Source/Advance.cpp
```diff
@@ -28,7 +28,7 @@ ERF::advance(Real time, Real dt, int amr_iteration, int amr_ncycle)
     getFluxReg(level + 1).reset();
   }
 
-  Real dt_new = dt;
+//  Real dt_new = dt;
```
ewquon, line 31: I think at some point we're due for another code clean-up. But probably after we have testing in place so we can verify that nothing gets broken.

Source/ERF.H
```diff
@@ -51,97 +51,127 @@ grow_box_by_one(const amrex::Box& b)
   return amrex::grow(b, 1);
 }
 
-//
-// AmrLevel-derived class for hyperbolic conservation equations for stellar
-// media
-//
+struct SolverChoice {
```
ewquon, line 54: @pkjha-aero is this going to eventually deprecate `ProbParm`?

Source/Advance.cpp
pkjha-aero: I agree

Source/ERF.H
```diff
@@ -51,97 +51,127 @@ grow_box_by_one(const amrex::Box& b)
...
+  prev = -1,
+  local = 0,
+  next = 1
+};
+
+enum class AdvectedQuantity {
+  unity, u, v, w, theta, scalar
+};
+
+enum class AdvectingQuantity {
```
ewquon, line 82: @pkjha-aero Can we rename this to `ConservedQuantity`? This will make the code clearer and more readable (and have less confusion w/ `AdvectedQuantity`).

Source/ERF.H
pkjha-aero: The variables we define in `ProbParm` in `prob_par.H` are specific to the problem and are utilized in `Exec/*` and not in `Source/*`. But `SolverChoice` is meant for the those things that apply to problems in general and are utilized in `Source/*`

Source/ERF.H
ewquon: Or use "primitive" and "conserved"

Source/RK3/RK3.H
```diff
@@ -66,6 +72,99 @@ void CalcDiffFlux(const MultiFab& cons_in,
...
+AMREX_GPU_DEVICE
+Real ComputeStrainRate(const int &i, const int &j, const int &k,
+                        const Array4<Real>& u, const Array4<Real>& v, const Array4<Real>& w,
+                        const enum NextOrPrev &nextOrPrev,
+                        const enum MomentumEqn &momentumEqn,
+                        const enum DiffusionDir &diffDir,
+                        const Geometry &geom);
+
+// Compute turbVisc, i.e., 'nut' or Kij at cell centres. Initialize it to 1.0
+//TODO: Check if it is better to  fill all Kij and use it or compute on the fly
```
ewquon, line 136: Under a "flops are free" paradigm, I would say just compute on the fly. Something like S_ij might make more sense to precompute and store.

Source/RK3/RK3_driver.cpp
```diff
@@ -67,53 +65,52 @@ void RK3_advance(MultiFab& cons_old,  MultiFab& cons_new,
...
-    const GpuArray<Real, AMREX_SPACEDIM> dx = geom.CellSizeArray();
-    
-    const    Array<Real,AMREX_SPACEDIM> grav{0.0, 0.0, 0.0};
-    const GpuArray<Real,AMREX_SPACEDIM> grav_gpu{grav[0], grav[1], grav[2]};
-
-    // ************************************************************************************** 
-    //
-    // Return update in the cons_upd_1 and *mom_update_1 MultiFabs
-    //
+    //TODO: Do we really need to use setVal again when we have done earlier with values of 0
```
ewquon, line 88: Previously, all vars were initialized to 0; this specifically sets density to rho0, which should probably be some dimensional input value.

ewquon (commented):
> My first concern (README in new demos) can be in separate pull requests. But my second concern is important. If these new codes are not validated, nobody would be able to use it -- at least, I can't; I can only work on validated code.
> 
> Anyway, it's all up to you.

@zzwz02 I share your concern. My plan is to get Raf spun up this week, and put in place regression testing as the first priority. We can always check out the code prior to this PR, generate the reference data, and then re-run the tests with the latest code. The plan is to automate this through github actions, building on top of the minimal compile tests we have right now, to alleviate our worries :)

Source/ERF.H
pkjha-aero: There is a separate treatment of the conserved quantity at cell centers, `cons` and `cons_in` etc. That's why I used this nomenclature. 
Another reason I used this name is the way I see the similarity in the equations. 
In each equation (see the attachment), depending on what the equation is for `{1, u, v, w, theta, scalar}`, these quantities are transported by each of `{rho_u, rho_v, rho_w}`
[ERF Discretization Draft V11_PKJ2.pdf](https://github.com/erf-model/ERF/files/6600717/ERF.Discretization.Draft.V11_PKJ2.pdf)

ewquon (approved):

PR #25, opened 2021-06-09
Implement advection terms as per the new code architecture

pkjha-aero:
Reviewers, please create a new issue if suggesting a modification. That will be taken care of in future PR. This PR is meant to serve as the "to go code" for all the developers to align.


ewquon (approved):
Looks good to me. I'm curious about changes in performance with the new architecture, we should do some benchmarking.

pkjha-aero:
> Looks good to me. I'm curious about changes in performance with the new architecture, we should do some benchmarking.

@ewquon I've made a note as TODO.

PR #28, opened 2021-06-29
Feature boundary conditions

rfhaque:


ewquon:
Glad to see all the tests passing now. I'd like to try to get the reg tests in place prior to merging in. If that doesn't happen in the next day or two I'll go ahead and merge this in.

PR #33, opened 2021-07-09
add fixed value bc functions

xyuan:
Add fixed value bc function to setup the fixed value boundary condition.

### DIFF
--- a/Source/BC/BCBase.H
+++ b/Source/BC/BCBase.H
@@ -138,6 +138,231 @@ namespace math_bcs {
     }
   }
 
+  template<>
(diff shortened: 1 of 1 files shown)


Source/BC/BCBase.H
```diff
@@ -138,6 +139,231 @@ namespace math_bcs {
     }
   }
 
+  template<>
+  inline void extval<BCBound::lower>(const amrex::Geometry& geom, const Box& b, int lo, Array4<Real> data, int idir, int ncomp) {
+    amrex::IntVect type = b.type();
```
pkjha-aero, line 144: @xyuan It would be better to pass variables other than 'data' as const

PR #34, opened 2021-07-09
Sign of application of gravity 

brigaudet:
The value of CONST_GRAV  is a positive constant, but its treatment in the vertical momentum equation assumes that its contribution is negative to the vertical momentum equation in RK3_stage.cpp, causing a bug.
The bug fix is to insert a negative sign when grav_gpu is defined.

### DIFF
--- a/Source/RK3/RK3_stage.cpp
+++ b/Source/RK3/RK3_stage.cpp
@@ -49,7 +49,10 @@ void RK3_stage  (MultiFab& cons_old,  MultiFab& cons_upd,
     // **************************************************************************************
     // Deal with gravity
     Real gravity = solverChoice.use_gravity? CONST_GRAV: 0.0;
-    const    Array<Real,AMREX_SPACEDIM> grav{0.0, 0.0, gravity};
+    // CONST_GRAV is a positive constant, but application of grav_gpu to the vertical momentum
+    // tendency assumes this quantity is negative.
+    //    const    Array<Real,AMREX_SPACEDIM> grav{0.0, 0.0, gravity};
+    const    Array<Real,AMREX_SPACEDIM> grav{0.0, 0.0, -gravity};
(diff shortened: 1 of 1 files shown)


Source/RK3/RK3_stage.cpp
```diff
@@ -49,7 +49,10 @@ void RK3_stage  (MultiFab& cons_old,  MultiFab& cons_upd,
     // **************************************************************************************
     // Deal with gravity
     Real gravity = solverChoice.use_gravity? CONST_GRAV: 0.0;
-    const    Array<Real,AMREX_SPACEDIM> grav{0.0, 0.0, gravity};
+    // CONST_GRAV is a positive constant, but application of grav_gpu to the vertical momentum
+    // tendency assumes this quantity is negative.
+    //    const    Array<Real,AMREX_SPACEDIM> grav{0.0, 0.0, gravity};
+    const    Array<Real,AMREX_SPACEDIM> grav{0.0, 0.0, -gravity};
```
pkjha-aero, line 55: good catch! Alternatively, the way we have negative sign for advection contribution and pressure gradient, we can change the sign in the momentum eqn itself. So, += dt * grav_gpu[idx] can be changed to (-dt) * grav_gpu[idx]

brigaudet: Correct, that would also work.  I proposed changing it where I did because at some point we might have non-horizontal levels, so the grav vector would have non-zero x and y components, which would then contribute to the horizontal momentum equations -- and tracking down sign errors in the horizontal momentum equations downstream seems more error prone then taking care of it all in one place.  And physically it makes sense that CONST_GRAV is a number with a positive value, but 'grav' is a vector to be applied to the equations which has a negative component in the z direction.

pkjha-aero: Agreed.

PR #35, opened 2021-07-09
Feature boundary conditions

rfhaque:


xyuan (approved):
yes, using template parameter for IDIR and upper and lower boundary is better.

brigaudet (approved):

xyuan (approved):
it sounds good.

PR #37, opened 2021-07-13
Initial inclusion of structures to implement NoSlipWall boundary conditions

brigaudet:
This is an initial implementation of code structures to implement the NoSlipWall boundary conditions.


xyuan:
That sounds good to me, it sounds that @rfhaque would like to use idir as template argument, not sure what he thinks about the current version.

PR #44, opened 2021-07-22
Feature boundary conditions

rfhaque:


Source/RK3/RK3_stage.cpp
```diff
@@ -46,8 +46,7 @@ void RK3_stage  (MultiFab& cons_old,  MultiFab& cons_upd,
     yvel.FillBoundary(geom.periodicity());
     zvel.FillBoundary(geom.periodicity());
 
-        amrex::Vector<MultiFab*> vars{&cons_old, &xmom_old, &ymom_old, &zmom_old};
-    //    amrex::Vector<MultiFab*> vars{&cons_old, &xvel, &yvel, &zvel};
+    amrex::Vector<MultiFab*> vars{&cons_old, &xvel, &yvel, &zvel};
```
rfhaque, line 49: @pkjha-aero @brigaudet Is this call to applyBC needed?

brigaudet:
Hi Rijaz,

It depends what RK3_driver.cpp will look like.  But if RK3_driver.cpp looks like what I see in the pull request with only one call to applyBC at the end of RK3_driver - then I would say yes, the call to applyBC in RK3_stage is needed.  BCs need to be applied after every application of the tendencies.

PR #45, opened 2021-07-22
fix the bugs in boundary condition functions

xyuan:
Fix all the bugs that is appeared in the boundary condition functions, and add some comments

### DIFF
--- a/Source/BC/BCBase.H
+++ b/Source/BC/BCBase.H
@@ -161,33 +161,24 @@ namespace math_bcs {
            amrex::ParallelFor(b, ncomp, [=] AMREX_GPU_DEVICE (int i, int j, int k, int n) noexcept
            {
              if (i > hi) {
-               amrex::Real grad = data(hi-1,j,k,n) - data(hi-2,j,k,n);
-               data(i,j,k,n) = data(hi-1,j,k,n) - (hi-1-i)*grad;
-             } else if (i == hi) {
-               amrex::Real grad = data(hi-1,j,k,n) - data(hi-2,j,k,n);
-               data(i,j,k,n) = data(hi-1,j,k,n) - (hi-1-i)*grad;
+               amrex::Real grad = data(hi,j,k,n) - data(hi-1,j,k,n);
+               data(i,j,k,n) = data(hi,j,k,n) - (hi-i)*grad;
              }
            });
         } else if (IDIR == 1) {
            amrex::ParallelFor(b, ncomp, [=] AMREX_GPU_DEVICE (int i, int j, int k, int n) noexcept
            {
              if (j > hi) {
-               amrex::Real grad = data(i,hi-1,k,n) - data(i,hi-2,k,n);
-               data(i,j,k,n) = data(i,hi-1,k,n) - (hi-1-j)*grad; 
-             } else if (j == hi) {
-               amrex::Real grad = data(i,hi-1,k,n) - data(i,hi-2,k,n);
-               data(i,j,k,n) = data(i,hi-1,k,n) - (hi-1-j)*grad;
+               amrex::Real grad = data(i,hi,k,n) - data(i,hi-1,k,n);
+               data(i,j,k,n) = data(i,hi,k,n) - (hi-j)*grad; 
              }
            });
         } else if (IDIR == 2) {
            amrex::ParallelFor(b, ncomp, [=] AMREX_GPU_DEVICE (int i, int j, int k, int n) noexcept
            {
              if (k > hi) {
-               amrex::Real grad = data(i,j,hi-1,n) - data(i,j,hi-2,n);
-               data(i,j,k,n) = data(i,j,hi-1,n) - (hi-1-k)*grad;
-             } else if (k == hi) {
-               amrex::Real grad = data(i,j,hi-1,n) - data(i,j,hi-2,n);
-               data(i,j,k,n) = data(i,j,hi-1,n) - (hi-1-k)*grad; 
+               amrex::Real grad = data(i,j,hi,n) - data(i,j,hi-1,n);
+               data(i,j,k,n) = data(i,j,hi,n) - (hi-k)*grad;
              }
            });
         }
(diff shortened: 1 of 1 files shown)


brigaudet:
Some of these changes I agree with, some I'm still not sure about.

*)  foextrap->apply_cell_based:  I'm fine with these.

*)  foextrap->apply_face_based:  It seems you are implying that, since lo = geom.Domain().smallEnd(IDIR) and hi= geom.Domain().bigEnd(IDIR), lo and hi don't know about the fact that face-based variables (here, the normal velocity components) have one extra grid point than 'lo' and 'hi' would suggest, so this needs to be taken into account when boundary conditions are applied to the data.  I don't think the current code takes this into account, so this would definitely be a bug.  But I think the lo end is unaffected by this; the change needs to be at the hi end (i.e., in the x direction XVEL BCs should be applied outside of lo and hi+1, whereas in the other directions they should be applied outside of lo and hi).  At the hi end, it looks like you use (hi-1) and (hi-2) to form the gradient for face_centered instead of hi and hi-1 -- but I think in for this case the gradient should be formed from (hi+1) and (hi).

*)  reflect_even->apply_cell_based:  I'm fine with these.

*)  reflect_even->apply_face_based:  OK, but note that apply_face_based does not reflect about the same point in space as apply_cell_based (because both mirror around the value at (lo) but (lo) is on a staggered grid for one and non-staggered on the other).  Also, the value at (lo) itself is not necessarily zero for reflect_even (unlike reflect_odd).  Probably best to not change the value at (lo) at all here.  And also -- at (hi) the same comments as in foextrap about hi vs. hi+1 apply.

*)  reflect_odd:  I'm fine with these (again, though note my comments above about the face_based functions at the hi end)

*)  hoextrap:  I'm fine with these (again, with my comments about evaluating face-based functions at the hi end).

*)  brian_reflect_odd:  I'm fine with the comments, look good -- except that the comment for the face_based function at the lo end refers to hi instead of lo.

xyuan:
> Some of these changes I agree with, some I'm still not sure about.
> 
> *) foextrap->apply_cell_based: I'm fine with these.
> 
> *) foextrap->apply_face_based: It seems you are implying that, since lo = geom.Domain().smallEnd(IDIR) and hi= geom.Domain().bigEnd(IDIR), lo and hi don't know about the fact that face-based variables (here, the normal velocity components) have one extra grid point than 'lo' and 'hi' would suggest, so this needs to be taken into account when boundary conditions are applied to the data. I don't think the current code takes this into account, so this would definitely be a bug. But I think the lo end is unaffected by this; the change needs to be at the hi end (i.e., in the x direction XVEL BCs should be applied outside of lo and hi+1, whereas in the other directions they should be applied outside of lo and hi). At the hi end, it looks like you use (hi-1) and (hi-2) to form the gradient for face_centered instead of hi and hi-1 -- but I think in for this case the gradient should be formed from (hi+1) and (hi).
> 


> *) reflect_even->apply_cell_based: I'm fine with these.
> 
> *) reflect_even->apply_face_based: OK, but note that apply_face_based does not reflect about the same point in space as apply_cell_based (because both mirror around the value at (lo) but (lo) is on a staggered grid for one and non-staggered on the other). Also, the value at (lo) itself is not necessarily zero for reflect_even (unlike reflect_odd). Probably best to not change the value at (lo) at all here. And also -- at (hi) the same comments as in foextrap about hi vs. hi+1 apply.
> 
> *) reflect_odd: I'm fine with these (again, though note my comments above about the face_based functions at the hi end)
> 
> *) hoextrap: I'm fine with these (again, with my comments about evaluating face-based functions at the hi end).
> 
> *) brian_reflect_odd: I'm fine with the comments, look good -- except that the comment for the face_based function at the lo end refers to hi instead of lo.



For high end velocity boundary, we need to derive the value at hi+2, hi+1, hi points based on the interior region (i<hi), so the gradient should be constructured at the interior region. Not sure how you think about the gradient from hi+1, and hi which is unknow at the current timestep, only the previous timestep values can be used.  This is same for the other face type variables, please double check it.

brigaudet:
Hi Xingqiu,


I'm not sure I completely understand your reply.

But this is my understanding:  values from lo to hi, inclusive are the 'valid domain' for cell-centered variables.
For face-centered variables, values from lo to hi+1, inclusive are the 'valid domain'.
(This is my understanding based on how lo and hi are defined - if you know I am in error in this, please let me know.)

Assuming the above is correct - valid domain fields are predicted each time step and do not necessarily need to be set according to boundary conditions.  Fields outside of the valid domain, such as the ghost points, do need to be set.  So, for many boundary conditions, this means (< lo) and (> hi) for cell centered, and (<lo) and (> hi+1) for face-centered.

For wall BCs only --  because of the Arakawa C grid staggering - additional points need to be set as BCs - the extreme valid normal velocities need to be set to zero.  The normal velocities follow the face-centered rules, so this mean norm(lo) = 0 and norm(hi+1) = 0.

Does this make sense?

Thanks,

Brian

xyuan:
Yes, we need set lo, and hi+1 which are the boundary interface values via boundary condition routine, it is 1 offset in the code though, I will fix it.


From: brigaudet ***@***.***>
Reply-To: erf-model/ERF ***@***.***>
Date: Monday, July 26, 2021 at 2:58 PM
To: erf-model/ERF ***@***.***>
Cc: "Yuan, Xingqiu" ***@***.***>, Author ***@***.***>
Subject: Re: [erf-model/ERF] fix the bugs in boundary condition functions (#45)

Hi Xingqiu,


I'm not sure I completely understand your reply.

But this is my understanding: values from lo to hi, inclusive are the 'valid domain' for cell-centered variables.
For face-centered variables, values from lo to hi+1, inclusive are the 'valid domain'.
(This is my understanding based on how lo and hi are defined - if you know I am in error in this, please let me know.)

Assuming the above is correct - valid domain fields are predicted each time step and do not necessarily need to be set according to boundary conditions. Fields outside of the valid domain, such as the ghost points, do need to be set. So, for many boundary conditions, this means (< lo) and (> hi) for cell centered, and (<lo) and (> hi+1) for face-centered.

For wall BCs only -- because of the Arakawa C grid staggering - additional points need to be set as BCs - the extreme valid normal velocities need to be set to zero. The normal velocities follow the face-centered rules, so this mean norm(lo) = 0 and norm(hi+1) = 0.

Does this make sense?

Thanks,

Brian

xyuan:
close this PR as this issue is addressed in Brain's PR

PR #49, opened 2021-07-30
fix some index issues in the bc function

xyuan:
This PR fixes some index issues in the foextrap and hoextrap and reflect bc function


pkjha-aero (approved):

brigaudet (approved):
OK!

pkjha-aero (approved):

PR #57, opened 2021-08-24
Fixing indexing errors in NoSlipWall and SlipWall

brigaudet:
1)  fix indexing loop for NoSlipWall at lower boundary; 2) fix indexing for SlipWall at boundaries.


ewquon:
I have verified that this solves an interior boundary issue I was seeing on a single domain. Will test whether it works for larger problems in parallel.

PR #65, opened 2021-09-17
Enable all regression tests in GitHub Actions

rafmudaf:
Currently, the command to run the regression tests only runs the isentropic vortex cases. This change runs all tests with the label "regression".


rafmudaf:
cc @ewquon

rafmudaf:
@ewquon @pkjha-aero Note that the failing tests here are the CUDA tests and are not related to the regression tests that I've enabled.

PR #66, opened 2021-09-17
Disable CUDA tests in GitHub Actions

rafmudaf:
Since the CUDA tests consistently fail in GitHub Actions, this pull request disables those tests until such time as they are prioritized to be fixed. The problem with perpetually failing tests is that it increases the noise in the development process. I'd like to know that the emails from GitHub are related to the things I've recently pushed and not related to some other work.

Before this is merged, a corresponding issue should be opened with a thorough description of the problem with the CUDA tests. Otherwise, there's a high risk that this will be forgotten.


pkjha-aero:
Opened the issue: https://github.com/erf-model/ERF/issues/67. @rfhaque is working on fixing the causes leading to CUDA tests failing (use of rand() function etc.)

pkjha-aero (approved):

Issue #23, opened 2021-06-04
Add READMEs for test problems


ewquon:
The three new demos `IsentropicVortex`, `ScalarAdvecUniformU` and `ScalarAdvecUniformUV` need README files explaining the problem solved and the corresponding references (if any, such as [this](https://github.com/zzwz02/ERF/blob/BenchmarkExample/Exec/ScalarDiffusion/README) and [this](https://github.com/zzwz02/ERF/blob/BenchmarkExample/Exec/TaylorGreen/README)).

_Originally posted by @zzwz02 in https://github.com/erf-model/ERF/issues/20#issuecomment-854260561_

dwillcox:
Looks like the only problems in Exec with Readme's are:

```
TaylorGreenVortex
ScalarDiffusionSine
ScalarAdvecRigidRot
```

So there are still lots more readme's needed to explain the setups.

PR #68, opened 2021-09-21
Regression Tests

pkjha-aero:


rafmudaf (approved):
This generally all looks good to me. The only thing I would add is a brief description of each test. This could be done in a README in the Tests directory, for example.

rafmudaf:
I'd also like to merge #69 first so that we get the information on what additional code was tested here.

rafmudaf:
I just saw issue #23 calling for READMEs for the test problems, so here's an up-vote for that.

ewquon:
> This generally all looks good to me. The only thing I would add is a brief description of each test. This could be done in a README in the Tests directory, for example.

This is a good idea.

Tests/CTestList.cmake
```diff
@@ -38,6 +38,7 @@ function(add_test_r TEST_NAME TEST_EXE PLTFILE)
     set(TEST_EXE ${CMAKE_BINARY_DIR}/Exec/${TEST_EXE})
     set(FCOMPARE_TOLERANCE "-r 1e-10 --abs_tol 1.0e-12")
     set(FCOMPARE_FLAGS "-a ${FCOMPARE_TOLERANCE}")
+    # TODO: Check if we need MPI_COMMANDS for running FCOMPARE_EXE
```
ewquon, line 41: Agree, I don't think this is necessary. @rafmudaf is it better to just leave it in?

rafmudaf: Good point, lets remove that just for our own future clarity.

rafmudaf: Oh it looks like this was merged before this comment was addressed. I'll create a new pull request for this.

PR #71, opened 2021-09-22
Fix CUDA build errors

### LINKED ISSUE #67: Fix failing CUDA issues (rand()...) and re-enable the tests in GitHub Actions
See this PR: https://github.com/erf-model/ERF/pull/66

rfhaque:


pkjha-aero:
I have raised a PR on this branch: https://github.com/rfhaque/ERF/pull/2

ewquon (approved):
I haven't looked at all the changes but for the most part, it looks like we've consolidated the initialization functions into a single, non-inlined function in prob.cpp, and also did away with prob_parm.H. I have faith in the regression tests we've implemented so far.

ewquon:
@rfhaque should we have a call to `InitRandom` somewhere? I currently don't see any way to specify the random seed and the comment at https://github.com/rfhaque/ERF/blob/development/Source/main.cpp#L102 sounds like a to-do item.

rfhaque:
@ewquon I think amrex implicitly calls InitRandom as a part of amrex::Initialize() (Src/Base/AMReX.cpp); that would also imply the same seed every time. Am I missing something?

rfhaque:
> I haven't looked at all the changes but for the most part, it looks like we've consolidated the initialization functions into a single, non-inlined function in prob.cpp, and also did away with prob_parm.H. I have faith in the regression tests we've implemented so far.

@ewquon The other significant change is replacing the managed global variables with a struct that is passed into the device lambda as needed. This ensures code portability across different GPU platforms.

pkjha-aero (approved):

rafmudaf (approved):

ewquon:
@rfhaque I see. https://github.com/AMReX-Codes/amrex/blob/f9216a341e51404e935733a4c953525a6f503d47/Src/Base/AMReX.cpp#L545

PR #72, opened 2021-09-23
Test setup improvements

rafmudaf:
See the commits, but generally
- Remove code coverage from Exec and Submodules
- Update Codecov badge on readme
- Add Codecov CMake flag as an option (for use with ccmake)
- Simplify regression test command


rafmudaf:
The tests here fail now because I've changed the build type to "Debug".

ewquon:
I'll see if we can get the Couette case sorted out today.

rafmudaf:
This passes now with your bug fix @ewquon. Good catch!

ewquon (approved):

PR #75, opened 2021-10-01
Remove global managed variables

rfhaque:


pkjha-aero:
While Github tests are yet to be finished, `ctest` passes all the 12 regression tests on my Mac (CPU only)

pkjha-aero (approved):

Issue #73, opened 2021-09-28
MPI+CUDA CI


ewquon:
Add CI workflow with `-DERF_ENABLE_MPI:BOOL=ON` in addition to `-DERF_ENABLE_CUDA:BOOL=ON`, to replicate @rfhaque's issue on Lassen HPC.

rfhaque:
Found this related issue with AMReX

https://github.com/ECP-WarpX/WarpX/issues/1132

rfhaque:
@ewquon @pkjha-aero Using cmake-3.16 or higher resolves this issue on lassen. We should upgrade our CI builds to use higher cmake-v3.16+.

dwillcox:
As of #1051 we now build our CUDA CI test on Github with MPI enabled, i.e. with `-DERF_ENABLE_MPI:BOOL=ON` as @ewquon mentioned.

PR #76, opened 2021-10-05
Add CI workflow for MPI+CUDA

ewquon:
To address #73

### DIFF
--- a/.github/workflows/mpi-cuda-ci.yml
+++ b/.github/workflows/mpi-cuda-ci.yml
@@ -1,0 +1,55 @@
+name: ERF CUDA CI
+
+on:
+  push:
+    branches: [development]
+  pull_request:
+    branches: [development]
+
+jobs:
+  cuda-build:
+    runs-on: ubuntu-18.04
+    name: CUDA v${{matrix.cuda_ver}}
+    strategy:
+      matrix:
+        cuda_pkg: [11-0, 10-2, 10-1, 10-0]
+        include:
+          - cuda_ver: "11.0"
+            cuda_pkg: 11-0
+            cuda_extra: libcurand-dev-11-0 cuda-cupti-dev-11-0
+          - cuda_ver: "10.2"
+            cuda_pkg: 10-2
+            cuda_extra: cuda-curand-dev-10-2 cuda-cupti-dev-10-2
+          - cuda_ver: "10.1"
+            cuda_pkg: 10-1
+            cuda_extra: cuda-curand-dev-10-1 cuda-cupti-10-1
+          - cuda_ver: "10.0"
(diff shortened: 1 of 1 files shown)


ewquon:
@rfhaque All the CI workflows finish without changing the cmake version.

.github/workflows/mpi-cuda-ci.yml
```diff
@@ -0,0 +1,55 @@
...
+          echo "deb https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1804/x86_64 /" | sudo tee /etc/apt/sources.list.d/nvidia-ml.list
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+              libopenmpi-dev cuda-command-line-tools-${{matrix.cuda_pkg}} cuda-compiler-${{matrix.cuda_pkg}} cuda-minimal-build-${{matrix.cuda_pkg}} cuda-nvml-dev-${{matrix.cuda_pkg}} cuda-nvtx-${{matrix.cuda_pkg}} ${{matrix.cuda_extra}}
+      - name: Configure and build
+        run: |
+          export PATH=/usr/local/nvidia/bin:/usr/local/cuda-${{matrix.cuda_ver}}/bin:${PATH}
+          export LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/local/cuda-${{matrix.cuda_ver}}/lib:${LD_LIBRARY_PATH}
+          cmake -Bbuild-${{matrix.cuda_pkg}} \
+            -DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo \
```
pkjha-aero, line 51: @ewquon @rafmudaf should we use `Debug` for all the three (`ci`, `cuda-ci`, `mpi-cuda-ci`) workflows?

ewquon: I'm debating the importance of the following:
* Being consistent vs thorough
* Using `Debug` to catch bugs involving uninitialized vars that might be hidden by optimization
* Using `RelWithDebug` or `Release` to represent the more likely use case--sometimes optimization can introduce unexpected segfaults and it would be good know about them

rfhaque: @ewquon @pkjha-aero Can we split the test cases into two categories: regression and performance. For regression, we test with Debug to catch errors, and for performance we run with Release to catch optimization errors etc. The sets of regression and performance tests can (an will) overlap. Suggestions?

asalmgren:
With the combination of our CI tests on every PR plus our nightly CPU-only and CUDA regression tests, i think our coverage is reasonable for now.  Closing this PR until we decide we need more.

PR #80, opened 2021-10-15
Github Actions: copy whitespace and tab style from AMReX

dwillcox:
This adds an automated check via GitHub Actions for whitespace and tab linting.

These style checks make sure there is no trailing whitespace in source files and that we are using spaces instead of tabs.


dwillcox:
The latest two commits apply the recommendations from our tab and trailing whitespace linter to replace tabs with spaces and remove trailing whitespaces in all source files.

With these changes, the style checks now pass.

pkjha-aero (approved):

PR #107, opened 2021-10-27
HIP CI Testing

dwillcox:
Adds HIP configuration to the ERF CI tests.


.github/workflows/hip.yml
```diff
@@ -0,0 +1,121 @@
...
+          -DCODECOV:BOOL=ON \
+          -DCMAKE_C_COMPILER=$(which clang)             \
+          -DCMAKE_CXX_COMPILER=$(which clang++)         \
+          -DCMAKE_CXX_STANDARD=17 \
+          ${{github.workspace}};
+        # ${{matrix.mpipreflags}} \
+
+        # for some reason this cmake command fails to build the code,
+        # and we need to use the make command that follows instead ...
+        # cmake --build ${{runner.workspace}}/ERF/build-${{matrix.os}} --parallel ${{env.NPROCS}};
```
etpalmer63, line 79: The reason could be that the environment variable `NPROCS` wasn't set. In some of the other runs there is a stage called setup where they set it. Here:

```
    - name: Setup
      run: |
        # echo "::add-matcher::.github/problem-matchers/gcc.json"
        echo "NPROCS=${{matrix.procs}}" >> $GITHUB_ENV
```

dwillcox: Oh, interesting!! Thanks @etpalmer63!

PR #119, opened 2021-10-28
Replace several macros by const

pkjha-aero:


asalmgren (approved):

pkjha-aero:
Looks like it may be better to abort this PR.
See this: https://stackoverflow.com/questions/16119923/using-constants-with-cuda
@rfhaque @ewquon @asalmgren opinion?

ewquon:
Sounds like from that stackoverflow thread that there's no compelling reason to use global const over preprocessor def, perhaps it's just a matter of style?

PR #120, opened 2021-10-28
Start a to-do based on discussion

pkjha-aero:
I have several "TODO:" in the code. Some of them are just some coding thoughts, some are physics stuff. 
Don suggested to create github issues corresponding to the TODOs of interest. Once these are done, we can close this PR. 
No need to merge!


asalmgren:
Should this be an "issue" rather than a PR?

PR #171, opened 2021-11-05
Doxygen startup

etpalmer63:
This PR was to get the Doxygen comments going again. I first went back through and changed several configuration options with a bias towards including more. If something unwanted is showing up in the doxygen documentation, the style can be changed. The configuration is stored in `Doxygen.in` if your curious, or you can just tell me and I'll be happy to adjust it to your liking. 

I also wrote a script that builds both the doxygen comments and sphinx documents. This will be helpful later for automatically updating the docs on the webpage via CI. It also makes it easy to remember the commands to build the documentation. :smile:  If you can't wait to see them, run the script and point your browser at the appropriate location of `index.html`.

Finally, I went through several files and added or modified comments so they could be included in Doxygen. Doxygen accepts multiple comment styles. Since a variety were already in place, I left them as is. If later, we desire to have a uniform look, it is not difficult to grep the doxygen comments once they are in place and modify them as desired.


etpalmer63:
Also, I should point out that some doxygen comments might not show up. I believe the issue is that Doxygen has a difficult time sorting out the AMReX syntax without processing AMReX source code files. It seems there is a right way to include those files so that things work as expected. But at the moment it is a work in progress.

Issue #191, opened 2021-11-09
Inconsistency between geostrophic forcing and coriolis terms


ewquon:
In steady state for `abl_driver_type == "GeostrophicWind"`, the added forcing term should balance out the Coriolis terms. There appears to be missing sine/cosine terms as well as a vertical forcing component based on U_g.

ewquon:
Addressed by https://github.com/erf-model/ERF/pull/196/commits/ea6d8ca51e240591faef6ad2554ac514eb87257e, included in https://github.com/erf-model/ERF/pull/196

PR #196, opened 2021-11-11
Setup for Ekman Spiral case

ewquon:


ewquon:
Preliminary results, matching the AMR-Wind input (running n=10 steps) looked pretty good. The ERF error compared to the analytical solution was surprisingly a few orders of magnitude higher than AMR-Wind. After n=100 the error increases; at n=1000, the errors are large enough to be noticeable (~0.1 m/s). Will continue to investigate, in relation to https://github.com/erf-model/ERF/issues/191...
ERF_Ekman_uv
ERF_Ekman_uverr

ewquon:
Verified that Coriolis terms and geostrophic forcing implemented in ERF matches AMR-Wind.

asalmgren:
It will be interesting to see if the time substepping -- then running at a larger time step -- helps this.

ewquon:
Commit https://github.com/erf-model/ERF/pull/196/commits/ea6d8ca51e240591faef6ad2554ac514eb87257e addresses https://github.com/erf-model/ERF/issues/191. 

Notes:
* Solution after 10 steps (following https://amr-wind.readthedocs.io/en/latest/developer/verification.html#ekman-spiral) matches the analytical solution (absolute errors < 1e-4) -- note that the solution was initialized to the analytical solution.
* A convergence study may be run with the provided `run_convergence_study.sh` script (requires an appropriate python environment, see beginning of `check_convergence.py`).
* Second-order convergence of errors was _not_ observed (for reference: https://amr-wind.readthedocs.io/en/latest/_images/ekman_spiral_error.pdf), the error was approximately constant for dz ranging from 6.25 m to 25 m. 
* Cases with latitude < 90 were not tested.

Holding off on creating an rtest until the aforementioned convergence issue is resolved.

PR #197, opened 2021-11-11
We need to impose bc's on density before converting velocity to momentum

asalmgren:


ewquon:
This fixes the density nan error associated with a hi-z slip wall that I saw in my Ekman-spiral case under development.

PR #198, opened 2021-11-11
Generalize Integration

dwillcox:
This PR will generalize the time integrator from our hard-coded RK3 to a framework where we only have to specify how to calculate the RHS and then where to fill BCs, etc.

The diffs for the ScalarAdvecDiffUniformU problem look pretty good, but of course the CI will run all the tests ...

```
            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 density                                          0                         0
 adv_0                               1.33226763e-15           1.403066252e-15
 pressure                                         0                         0
 temp                                             0                         0
 theta                                            0                         0
 x_velocity                         1.421085472e-14           1.421085472e-16
 y_velocity                                       0                         0
 z_velocity                                       0                         0
```


asalmgren (commented):
Nice job!  Much more general than I was expecting:

1) I'm thinking the generic stuff should go into AMReX itself... what do you think?  (Looks like you're heading that way since they're AMReX_... names :-) )
2) I'm wondering if we should use "TimeIntegration" instead of "Integration" just to be more clear?

dwillcox:
I've renamed "Integration" to "TimeIntegration" like you said.

And yeah! Moving the AMReX_* files into AMReX is definitely what I'd like to do. For now so this can get merged, I'll leave it as is, since it works. Then when I PR those files into AMReX, I'll update our AMReX submodule to use that instead, since I would like to write some extensive documentation for how to use them for the AMReX docs. But ERF is running with it now already.

All tests pass. Attached is the ctest output.

[test.log](https://github.com/erf-model/ERF/files/7531117/test.log)

PR #208, opened 2021-11-12
Combine the problems of advection by uniform velocity in in 1 and 2D

pkjha-aero:
…and update corresponding reg tests


pkjha-aero:
@asalmgren I have merged the two test problems in `Exec`, but left them separate in the `Tests/test_files` to make use of the existing cmake setup. Both the tests make use of the common executable for scalar advection by a given velocity whether in 1D or 2D.

PR #209, opened 2021-11-12
LES Equations in Theory Section

baperry2:
I thought writing out the different equation sets for different LES models might help us nail down exactly what we are trying to do.

This is a draft for now. Several questions popped up for me while looking things over and I highlighted things with notes. Some are just things I don't understand yet, some are additional things I think might be worth implementing. There are also some inconsistencies with notation that I know will need to be cleaned up.


baperry2:
Cleaned this up so it should be ready for integration now.

A few notes:
- It would great if someone could add a sentence explaining why we solve for theta instead of some other form of the energy equation, to help the uninitiated in atmospheric science (like me)
- I re-arranged the discussion a bit to try to maintain parallelism in how we discuss viscous and other molecular transport terms, as they all come from the same underlying physics.
- There some pieces that will need modification for consistency in the thermal conduction term once we nail down what the form should be.
- The form of the TKE equation should be verified, might be missing a couple terms (e.g. bouyancy related), but I just wanted to get something in there for now.
- We still need to clarify on where ABL/PBL models fit into these equations and/or their boundary conditions.
- Do we want to implement dynamic models for the Smagorinsky coefficients? That is standard practice in combustion but I'm not sure about ERF applications and if we're planning to primarily use Deardorff anyway it might not be worth the effort.

Issue #223, opened 2021-11-15
Check diffusive term for scalar evolution equation


asalmgren:
is the code currently computing the gradient of C or of (rho C)

baperry2:
Updated code has diffusive term as div(rho*alpha grad C).

Issue #219, opened 2021-11-14
Scalar Transport SGS Models


baperry2:
We should add Smagorinsky-like SGS models for theta and scalars.

baperry2:
Done in the linked PR.

PR #232, opened 2021-11-18
Smagorinsky fluxes for scalars/temperature

baperry2:
Addresses open issues relating to scalar transport:

https://github.com/erf-model/ERF/issues/219 Smagorinsky terms for scalars
https://github.com/erf-model/ERF/issues/223 Check form of diffusive terms for scalars

Also slightly modifies molecular transport, moving `rhoAlpha` inside the divergence so it is in the same place as `mu_t/Pr_t = rhoAlpha_t`. But constant `rhoAlpha` is still assumed (user specifies an `alpha` and now also a corresponding `rho`, then the product is computed and used throughout the simulation).

Still needs cleaning up/testing, some updates in the docs on discretization, and I'm sure this will fail under clang-tidy.


baperry2:
This should be good to go now.

This PR also allows molecular and turbulent diffusive/viscous processes to be turned on and off independently, rather than assuming that only one is active. This is done through a new input file option `erf.molec_diff_type` which can be `None` or `Constant`. The default behavior remains unchanged: if an LES model is specified only the turbulent viscosity from the Smagorinsky model is used, if no LES model is specified then a constant dynamic viscosity is used.

asalmgren (approved):

PR #237, opened 2021-11-20
1) More code clean-up

asalmgren:
2) Add erf_init_hse to set rhobar in the problem directory.
We still need to create pbar from rhobar, and write/read it in checkpoints


dwillcox:
This PR into your fork's branch should fix it, the issue was the hse vector sizes were defined in the `post_init` function, so this moves that definition into `initData`.

https://github.com/asalmgren/ERF/pull/1

PR #253, opened 2021-11-23
Deviatoric part of strain-rate tensor

### LINKED ISSUE #61: DNS strain
In DNS mode, we appear to be neglecting the div(U) term in the strain (that's not necessarily zero for compressible flow). My understanding is the strain-rate tensor should be:
image
`\epsilon_{ij} = \frac{1}{2}\left( \frac{\partial U_i}{\partial x_j} + \frac{\partial U_j}{\partial x_i} + \frac{2}{3}\nabla \cdot U \delta_{ij} \right)`

Priority: low

pkjha-aero:


pkjha-aero:
@asalmgren @baperry2 With the consideration of ``D_ij`` terms one of tests failing and change in error for others is not surprising. We can possibly update the reference results at this stage.

asalmgren:
I'm a little concerned about the error in the Ekman spiral test -- that velocity field is divergence-free so I wouldn't expect any change.

Source/TimeIntegration/StrainRate.cpp
```diff
@@ -136,3 +134,93 @@ ComputeStrainRate(const int &i, const int &j, const int &k,
...
+                     const enum MomentumEqn &momentumEqn,
+                     const enum DiffusionDir &diffDir,
+                     const GpuArray<Real, AMREX_SPACEDIM>& cellSize) {
+    Real dx = cellSize[0];
+    Real dy = cellSize[1];
+    Real dz = cellSize[2];
+
+    Real expansionRate = 0;
+
+    //TODO: Check if it is more efficient to store computed divergence at cell-centers rather than
```
baperry2, line 150: I'm pretty sure it will be more efficient to precompute the divergence. Right now it takes 6 divergence computations to compute a single stress, so precomputing the divergence should reduce the cost associated with calls to this function by 6x. I think that can wait though. There are several places that could be similarly revised.

Source/TimeIntegration/StrainRate.cpp
```diff
@@ -136,3 +134,93 @@ ComputeStrainRate(const int &i, const int &j, const int &k,
...
+
+AMREX_GPU_DEVICE
+Real
+ComputeExpansionRate(const int &i, const int &j, const int &k,
+                     const Array4<Real>& u, const Array4<Real>& v, const Array4<Real>& w,
+                     const enum NextOrPrev &nextOrPrev,
+                     const enum MomentumEqn &momentumEqn,
+                     const enum DiffusionDir &diffDir,
+                     const GpuArray<Real, AMREX_SPACEDIM>& cellSize) {
+    Real dx = cellSize[0];
```
baperry2: Another potential performance optimization would be to pass invCellSize here instead of cellSize to avoid all the divides (and similarly in other parts of the code).

Source/TimeIntegration/StrainRate.cpp
```diff
@@ -136,3 +134,93 @@ ComputeStrainRate(const int &i, const int &j, const int &k,
...
+    Real dy = cellSize[1];
+    Real dz = cellSize[2];
+
+    Real expansionRate = 0;
+
+    //TODO: Check if it is more efficient to store computed divergence at cell-centers rather than
+    // computing on the fly
+
+    // Divergence at cell (i, j, k)
+    Real divergence = (u(i+1, j, k) - u(i, j, k))/dx +
```
baperry2: Move inside the switch so effort is not wasted computing this when it is not used. Even better, for readability, make a function (and force inlining for efficiency) to compute divergence, which could replace the other divergence computations below as well. 

Doing the above would also allow replacing all the `switch` conditionals with logic operations to compute the indices where divergence is computed (similar to here: https://github.com/erf-model/ERF/blob/724c3f50f46758c173326b52552a9c3a2aeb70d1/Source/TimeIntegration/Diffusion.cpp#L142), which I find to be more readable - but that's just my opinion, so it's not necessary.

baperry2 (changes requested):
Looks good. There are a couple performance optimizations that may make sense down the road. The only thing essential for now would be eliminating the divergence computation when it is not used.

PR #365, opened 2022-01-24
add buoyance source term to KE equation

xyuan:
add buoyance source term to KE equation


asalmgren (approved):

asalmgren:
The code looks good.  Could you go ahead and update the documentation as well?  I see some typos in the equations as written, specifically I think dt should multiply all the source terms.  Also can you check the equations for correctness?  I think we must be missing a factor of rho somewhere ...?

asalmgren:
In fact I'm guessing all those terms on the last line may be missing a factor of both dt and rho ...?

PR #384, opened 2022-02-27
Xyuan/io

xyuan:
The NetCDF file reader should work based on my test on Summit, I also added some example code to build MultiFab from the NDArray


asalmgren:
We want to have the option of writing checkpoint files in native format even if we build with USE_NETCDF -- let's make that choice a run-time parameter not a compile-time choice.   Do you want to fix that before merging?  Or we can correct in a follow-up PR.

xyuan:
Let do it in the next PR, I will create a new PR on this task tomorrow.

PR #393, opened 2022-03-17
Initializing ERF with idealized meteorological data

pkjha-aero:
- Initialize ERF with idealized (horizontally homogeneous) meteorologic data provided in a NetCDF file


xyuan (commented):
I would suggest to reorganize the code to read the NetCDF file once, instead of open and read the NetCDF file multiple times for each variables. 

The geometrical grid data hasn't been read and used when you intialize the new ERF run, but I am not sure that is critical or it is normalized to (0,1).

PR #401, opened 2022-03-29
fix bugs in indexing for MOST BC application

baperry2:
Some out of bounds indices were being used around the boundaries, causing the code to crash in DEBUG mode. This is an attempt to fix that. @xyuan and @dwillcox - can you check and make sure none of the changes affect the intended functionality?


dwillcox (approved):
Looks great, good catches @baperry2 !

PR #402, opened 2022-03-30
Implementation of MYNN2.5 PBL Model

baperry2:
Covers implementation of the MYNN PBL Model.

Will add documentation before merging.

### DIFF
--- a/Source/SpatialStencils/PBLModels.H
+++ b/Source/SpatialStencils/PBLModels.H
@@ -1,0 +1,119 @@
+#ifndef _PBLMODELS_H_
+#define _PBLMODELS_H_
+
+#include <TimeIntegration.H>
+#include "DirectionSelector.H"
+
+inline
+void ComputeTurbulentViscosityPBL(const MultiFab& xvel, const MultiFab& yvel, const MultiFab& zvel,
+                                  const MultiFab& cons_in, MultiFab& eddyViscosity,
+                                  const Geometry& geom,
+                                  const SolverChoice& solverChoice,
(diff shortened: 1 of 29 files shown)


Source/SpatialStencils/PBLModels.H
```diff
@@ -0,0 +1,203 @@
...
+      const Box &dbx = geom.Domain();
+      Box sbx(bx.smallEnd(), bx.bigEnd());
+      sbx.grow(2,-1);
+      AMREX_ALWAYS_ASSERT(sbx.smallEnd(2) == dbx.smallEnd(2) && sbx.bigEnd(2) == dbx.bigEnd(2));
+
+      const GeometryData gdata = geom.data();
+
+      const Box xybx = PerpendicularBox<ZDir>(bx, IntVect{0,0,0});
+      FArrayBox qintegral(xybx,2);
+      qintegral.setVal(0.0);
```
dwillcox: `qintegral.setVal<RunOn::Device>(0.0);`

dwillcox:
Looking at the failed CUDA runs, I don't think that's our fault.

It seems like in the Github Action setup, our `apt-key` command is failing when we try to retrieve repository data from NVIDIA.

Possibly this is either NVIDIA or Github's fault, as this worked earlier today and you haven't changed anything in the Github Action setup script.

The ROCm HIP tests passed though, and those have always been trickier than the CUDA tests, so it's likely the CUDA tests would pass if they were able to run.

Hopefully this problem goes away soon, I'll re-run the failed jobs later today to see if it does ...

baperry2:
Looks like whatever was happening to the CUDA runs resolved itself. Going to merge now that I've put in some documentation.

PR #422, opened 2022-04-15
set default for AMREX_HOME path in Make.ERF

baperry2:
When compiling with GNUmake, the path to the AMReX submodule will automatically be used by default. 

If the AMREX_HOME environment variable is defined, that takes precedence, so this does not affect the ability to compile using AMReX downloaded in other locations if the user so desires.


dwillcox:
nice, that saves some typing :)

PR #444, opened 2022-04-26
WIP: MOST cleanup and documentation

baperry2:
Draft PR for now as this is ongoing, but creating to allow discussion. @xyuan and @dwillcox, please chime in if anything doesn't seem right to you.

Changes so far:

- [x] Currently, the computed theta_mean is actually rhoTheta_mean. I modified it to be the planar_average(theta), but I'm not sure whether that or the mass-weighted version [planar_average(rhoTheta)/planar_average(rho)] would be more physically appropriate.
- [x] Use the theta turbulent diffusivity instead of viscosity when applying MOST for theta
- [x] Compute and use vertical turbulent diffusivity components when the PBL model is active

Other outstanding issues planned to be part of this PR:

- [x] Documentation of how shear stress/flux boundary conditions are applied (why turbulent visc/diff shows up)
- [x] Verify dimensional consistency when applying boundary conditions
- [ ] I think it might be possible to avoid pre-computing turbulent visc/diff - more on this later
- [ ] the `setupABLMost` function is called one per timestep at the beginning of `ERF::Advance()`, but I believe it should be called at each slow substep

### DIFF
--- a/Docs/sphinx_doc/BoundaryConditions.rst
+++ b/Docs/sphinx_doc/BoundaryConditions.rst
@@ -167,54 +167,54 @@ boundary values using ``FillPatch``.
 
 MOST Boundaries
 -------------------
-Monin-Obukhov similarity theory (MOST) is used to describe the atmospheric surface layer (ASL), The MOST theory assumes that the ASL is in a steady state and horizontally homogenous, and the turbulent stresses :math:`\overline{u^{'}w^{'}}` and :math:`\overline{w^{'}v^{'}}` are assumed to be constant with height. Based on these assumptions,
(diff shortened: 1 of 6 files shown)


xyuan (commented):
the changes look nice!

baperry2:
Ok, I rebased onto development to capture the refactor of `ERF_PhysBCFunct.{H,cpp}`. But it looks like I might need to do some debugging - the modified code crashes on the second timestep for the ABL `inputs_most` case

baperry2:
I'm going to close this and create separate PRs for these fixes due to the changes in the underlying code

PR #640, opened 2022-08-09
Read in and store SST

mmsanders:
Read in SST from WRF input


mmsanders:
Failed test is residual from previous PR

PR #788, opened 2022-11-03
Verified MOST BC

AMLattanzi:
This passed analytical verification for the MOST BC.

Please review the factor of 2 changes with turbulent viscosities. 

Once approved, I'll run checks on Biollante before completing the merge.

### DIFF
--- a/Source/Diffusion/ComputeStress_N.cpp
+++ b/Source/Diffusion/ComputeStress_N.cpp
@@ -35,7 +35,7 @@ ComputeStressConsVisc_N(Box& bxcc, Box& tbxxy, Box& tbxxz, Box& tbxyz, Real mu_e
 
 void
 ComputeStressVarVisc_N(Box& bxcc, Box& tbxxy, Box& tbxxz, Box& tbxyz, Real mu_eff,
-                       const Array4<const Real>& K_turb,
+                       const Array4<const Real>& mu_turb,
                        Array4<Real>& tau11, Array4<Real>& tau22, Array4<Real>& tau33,
                        Array4<Real>& tau12, Array4<Real>& tau13, Array4<Real>& tau23,
                        const Array4<const Real>& er_arr)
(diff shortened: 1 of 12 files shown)


AMLattanzi:
This branch just passed all tests (except ABL-MOST because that benchmark needs updating) on Biollante.

Source/Diffusion/ComputeStress_N.cpp
```diff
@@ -44,9 +44,9 @@ ComputeStressVarVisc_N(Box& bxcc, Box& tbxxy, Box& tbxxz, Box& tbxyz, Real mu_ef
 
     // Cell centered strains
     amrex::ParallelFor(bxcc, [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept {
-        Real mu_11 = mu_eff + K_turb(i, j, k, EddyDiff::Mom_h);
+        Real mu_11 = mu_eff + 2.0 * K_turb(i, j, k, EddyDiff::Mom_h);
```
baperry2: I think this code is correct, but the variable names could potentially cause confusion down the road. Here the variables called `mu` have the factor of two already included but the variable called `K` does not, while the definition of K is the opposite (`K= 2*mu_t`). Ideally, we'd have a variable name with `K` in it whenever the factor of two multiplication has already been performed, and a variable name with `mu` whenever it has not.

Source/Diffusion/ComputeTurbulentViscosity.cpp
```diff
@@ -73,7 +73,7 @@ void ComputeTurbulentViscosityLES (const amrex::MultiFab& Tau11, const amrex::Mu
           Real SmnSmn = s11bar*s11bar + s22bar*s22bar + s33bar*s33bar
                       + 2.0*s12bar*s12bar + 2.0*s13bar*s13bar + 2.0*s23bar*s23bar;
 
-          K_turb(i, j, k, EddyDiff::Mom_h) = 2.0 * CsDeltaSqr * cell_data(i, j, k, Rho_comp) * std::sqrt(2.0*SmnSmn);
+          K_turb(i, j, k, EddyDiff::Mom_h) = CsDeltaSqr * cell_data(i, j, k, Rho_comp) * std::sqrt(2.0*SmnSmn);
```
baperry2: Here `K_turb` should be renamed `mu_turb` or maybe something like `turb_trans_coeff` because it does not have the factor of two and contains the turbulent diffusivities as well

Source/ERF.cpp
```diff
@@ -1198,17 +1196,17 @@ ERF::define_grids_to_evolve (int lev)
 void
 ERF::setupABLMost (int lev)
 {
-    MultiFab& S_new = vars_new[lev][Vars::cons];
-    MultiFab& U_new = vars_new[lev][Vars::xvel];
-    MultiFab& V_new = vars_new[lev][Vars::yvel];
-    MultiFab& W_new = vars_new[lev][Vars::zvel];
+    MultiFab& S_old = vars_old[lev][Vars::cons];
```
baperry2, line 1199: Because this now uses the old data, can this comment above where the function is called be removed? 

h1tps://github.com/erf-model/ERF/blob/3d409b7de3dd9ff450f60a36d9109b1745b37dbb/Source/TimeIntegration/ERF_TimeStepping.cpp#L97

baperry2 (dismissed):
I think this is all correct, my only comments are on variable naming, to avoid potential confusion when someone else looks at the code later.

PR #813, opened 2022-12-06
solver use total density, and add some aux moisture variables

xyuan:
solver use total density, and add some aux moisture variables


xyuan:
close this PR as we agreed that the (1+q) factor is needed. so I will make another PR to merge the microphysics model

PR #852, opened 2022-12-19
Allow user-specified c_p

ewquon:


ewquon:
@asalmgren FYI this WILL break the rtests for the density current and bubble problems. Note:
1. Ran ctest on our HPC, pass rate was 73%. I expect that this is due to the inconsistency that you mentioned, between the initialized values with the user-specified c_p vs the const c_p=1004.5 used in various places throughout the code (e.g., ERF_make_fast_coeffs.cpp). The MovingTerrain* cases are also failing, but I think this has been a persistent issue...
2. Sanity checked the fcompare output from DensityCurrent and Bubble_DensityCurrent (same case, but with more general problem) — these are identical.
3. Verified that the output before and after these changes gives identical results with specified `c_p=1004.5`.

PR #919, opened 2023-01-26
ABL updates

ewquon:
* Correct behavior of init_type==input_sounding
* Add updated Rayleigh damping options
* Add option to apply non-random divergence-free (sinusoidal) initial perts
* Add option to apply random rho*theta initial perts


ewquon:
This also includes the fix in #918

PR #967, opened 2023-02-18
TKE fix

AMLattanzi:
1. Dissipation should have a minus sign for accumulation into the RHS.

2. Production should have 2*mu_t

3. Modify the logic when computing turbulent length scale to avoid a divide by zero error.


AMLattanzi:
My changes are based upon the following equations taken from Steve Pope's book on turbulent flows. The first gives the TKE balance and the dissipation closure below that while the second image gives the production closure:
image
image (1)

PR #989, opened 2023-03-10
Fix Deardorff implementation, update docs, add additional sampling outputs/options

ewquon:
@AMLattanzi, see commits 19d19531e0ced97051d84888695bb2f7f42d572a through e7e6b5e0065bc96b17fc9ec70f9dc2e4e108d3b6 from today (March 10) for the Deardorff fixes we had been discussing prior to your trip. I've updated the docs to reflect what I expect to be currently implemented in the code but an extra set of eyes would be nice.


ewquon:
Also, @AMLattanzi, is there any reason to effectively recalculate `mu_turb` here https://github.com/ewquon/ERF/blob/aba4111b6be54d1f54688d5a35871365f4d29362/Source/Diffusion/DiffusionSrcForState_N.cpp#L365-L367, or could we just use the snippet I have in the comment?

AMLattanzi:
I don't see an obvious reason to recalculate mu_turb at that point. We compute mu_turb at the start of the time step and hold it constant over the RK stages. If you do update it at that point, you will be changing mu_turb at each RK stage, which would be inconsistent with how everything else interacts with mu_turb.

Source/Diffusion/DiffusionSrcForState_N.cpp
```diff
@@ -343,30 +348,42 @@ DiffusionSrcForState_N (const amrex::Box& bx, const amrex::Box& domain, int n_st
...
+            // eddy diffusivity for heat, KH = (1 + 2*l/delta) * mu_turb
+            Real KH = cell_data(i,j,k,Rho_comp) * l_C_k * length * (1.+2.*length/l_Delta) * std::sqrt(E);
+
+            // Add Buoyancy Source
+            // where the SGS buoyancy flux tau_{theta,i} = -KH * dtheta/dx_i,
+            // such that for dtheta/dz < 0, there is a positive (upward) heat flux;
+            // the TKE buoyancy production is then g/theta_0 * tau_{theta,w}
+            hfx_x(i,j,k) = 0.0;
+            hfx_y(i,j,k) = 0.0;
+            hfx_z(i,j,k) = -KH * dtheta_dz;
```
AMLattanzi, line 375: I'm not sure this is the best place to store the heat flux? Specifically, this part of the code is for RhoKE sources and only gets triggered for Deardorff. By contrast, we compute the diffusive fluxes up above and that code is triggered for any turbulence model. If there is an issue with the diffusivity with theta, we may need to modify ComputeTurbulentViscosity.

AMLattanzi (commented):
This looks good. Just a minor comment on the heat flux storage.

AMLattanzi (approved):

PR #1019, opened 2023-03-21
Numerous adjustments to Deardorff model, change sign convention for tau

ewquon:
Deardorff:
* fix sign error on buoyancy term
* add limit on mixing length under stable stratification -- should be less than or equal to the cube root of cell volume used for unstable conditions and, following WRF, set a lower bound of 1e-8
* add limit on `rho*e` (> eps) after slow step
* add "wall effect" to the dissipation coefficient, enable by setting `erf.Ce_wall = 3.9`
* initialize SFS TKE field (`e` rather than `rho*e`) to a constant by setting `erf.KE_0`, default is to 0.1 (match SOWFA / AMR-Wind) -- for backwards compatibility, can still set `erf.RhoKE_0`
* update eddy diffusivities in first RK stage -- _work in progress..._

General turbulence modeling consistency:
* flip sign on tau so that `tau_ij = R_ij - R_kk delta_ij/3`, `tau_ij = -2 mu_t S_ij`, and shear production `P = -tau_ij S_ij`
* rename "SGS" to "SFS" in code


Source/Diffusion/ComputeStress_T.cpp
AMLattanzi, line 1: tau_33 at L51 and L276 should have a *= -mu_tot to be consistent.

ewquon: Thanks for catching this!

AMLattanzi (approved):
These rounds of edits pass all the Biollante reg tests (except ABL-Deardorff since it changes the Deardorff turbulence model) so I'm good with moving it into develop and remaking that benchmark.

PR #1028, opened 2023-03-30
Changes that enable tiling across z

mukul1992:
This required a new function `TileNoZ() `that can replace `TilingIfNotGPU()`. It returns a tile size that prevents tiling across z-direction for MFIter loops that break due to the z-tiling. Several MFIter loops are then changed to use this as needed for running the program successfully. 
I was able to make the terrain version work for DensityCurrent.
The ABL-Smag-OMP also checks out, the plotfiles differ with less than e-7 errors.
The ABL-MOST-OMP is not working yet, and other executables need to be checked.


mukul1992:
just added a commit with more changes for z-tiling in WPS_Test.

mukul1992:
corrected TileNoZ to be a host function for correct CUDA compile

mukul1992:
added z-tiling in MOSTaveraging.cpp

PR #1037, opened 2023-04-03
Fix fast rhs OMP tiling issues

mukul1992:
Fixed by putting x/y-mom RHS calculation in a separate MFIter loop.
I have not implemented this for `fast_rhs_MT`: the `MFIter` loop has its tiling set to `false` for now.


mukul1992:
I am not able to figure out how to remove the submodules from this PR. I was hoping this can be done by the moderator while approving.

PR #1115, opened 2023-06-07
Fix bubble input

ewquon:
Need to set `prob.T_0 = 0` if an `erf.init_type` is specified


AMLattanzi:
This was wrapped up in MR 1117.

PR #1116, opened 2023-06-07
add option erf.init_sounding_ideal = 1

xyuan:
add option erf.init_sounding_ideal=1 to fix the difference between pres, and pres_hse


AMLattanzi:
This was wrapped up in MR 1117.

PR #1134, opened 2023-06-29
don't run style checks on Submodules

baperry2:


asalmgren (approved):

asalmgren:
I'm moving these commits into a PR that I can force to auto-merge ... need to figure out why the HIP tests are failing.

baperry2:
Thanks. The HIP tests are weird, seems like they fail most of the time but sometimes randomly pass on re-running.

PR #1204, opened 2023-08-10
Efficient Scalar Advection / New Prob Type

eabogle:
Added in capability for users to erf.use_efficient_advection as an option. 
Added in problem type 14 in Scalar Advection that is just a cos function.


eabogle:
duplicated In another pull request, deleting this one

PR #1217, opened 2023-08-22
Turn some CUDA warnings into errors

WeiqunZhang:


WeiqunZhang:
The CUDA CIs are supposed to fail.

PR #1223, opened 2023-08-29
MOST w/ Charnock

AMLattanzi:
Add Charnock. Not integrated into heat flux calculation yet. Also need to consider using structs to contain the iteration code and template the update_fluxes function to pass the struct in.

@ewquon we will want to chat about correctness and interpretation with a heat flux component.

@asalmgren we will want to look at modularizing the functions in ABLMost.cpp (I think we can ask the type question once, instantiate a struct that has the source code, and template in the same manner as the advection operator). This can be connected to requests made by Yorgos.


asalmgren:
See line 254 in MOSTAverage.cpp -- you have - + there ... otherwise looks good

asalmgren (approved):

PR #1224, opened 2023-08-30
Modular MOST

AMLattanzi:
Only a CPU compile test. Need to verify solution doesn't change, GPU compile test, and speed test.

### DIFF
--- a/Source/BoundaryConditions/ABLMost.H
+++ b/Source/BoundaryConditions/ABLMost.H
@@ -38,13 +38,28 @@ public:
         amrex::ParmParse pp("erf");
         pp.query("most.z0", z0_const);
 
+        // Specify how to compute the flux
+        std::string flux_string{"moeng"};
+        pp.query("most.flux_type", flux_string);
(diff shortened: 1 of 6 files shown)


AMLattanzi:
Successful comparison in serial against the current development branch with **ABL/inputs_most** after 1 step. Verified compile and run on GPU.

@mukul1992 could we do a quick performance test on this to verify there isn't a significant hit? The above mentioned inputs would be a good option.

@asalmgren we can put the computation for Yorgos into the Donelan flux type when the time comes.

mukul1992:
Comparing the **dev branch** as of today, last PR being #1227, with the **old state** #1222 just before this PR, using the input file `inputs_most` with no I/O and 100 time steps:

### 1 CPU test
old: 20.38 s, dev: 20.5 s

### 1 GPU test
old: 0.88 s, new: 0.89 s

There does not seem to be any significant performance hit. The plotfile comparison shows relative errors $< 10^{-12}$ between the two.

PR #1232, opened 2023-09-06
Refactor initialization routines, problem definition

ewquon:
Major overhaul of the Exec source directory. Problems should now be consistently defined as perturbations applied to a base state.

Updates: 
* Create new `ProblemBase` class, which has virtual function definitions for: `erf_init_dens_hse`, `init_custom_pert` (formerly `init_custom_prob`), `init_custom_terrain`, and `erf_init_rayleigh`
* Problem definition in Exec should take place in a derived `Problem` class, instantiated by `amrex_probinit`
* "Recipes" for common Problem member functions are provided in Prob/ and should be included within the Problem class definition in prob.H; these include:
  * `Prob/init_constant_density_hse.H` — base state from a uniform density field, probably makes sense only if gravity is turned off
  * `Prob/init_density_hse_dry*.H` — base state from integrating vertically through a dry air column, with variants for flat (no) terrain, terrain, or both
  * `Prob/init_rayleigh_damping.H` — add a WRF-style rayleigh damping layer with reference values defined by problem params
* Initialization behavior is defined by `erf.init_type` which defaults to "" (use the calculated HSE as the background state); a new `erf.init_type="uniform"` option has also been implemented for backwards compatibility, if the problem calls for fixed reference density and temperature values that differ from the HSE (e.g., the scalar advection/diffusion problem)


ewquon:
I think the DensityCurrent problem not passing is to be expected and is acceptable.

This is the fcompare output of plt00000 before and after:
```            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 density                            7.506992141e-05           6.489842887e-05
 x_velocity                                       0                         0
 y_velocity                                       0                         0
 z_velocity                                       0                         0
 pressure                           3.186329559e-05           3.204511179e-10
 theta                                0.03459916559           0.0001153205362
 pres_hse                                         0                         0
 dens_hse                                         0                         0
```
The base states are identical. The problem is set up to perturb rho, not rho*theta, i.e. the initial pressure perturbation should be 0. In the new case, the maximum initial pressure perturbation is O(1e-10) whereas in the previous plt00000, the max initial perturbation was ~O(1e-5), so the new implementation should be more exact.

The old density field range: `[-0.0000750699214138173,  0.05267775140949682]`
The new density field range: `[-0.00006500796407871157, 0.05267775142007136]`

The difference stems from the previous implementation calling `init_isentropic_hse` within `init_custom_prob` and then using the values that come from the Newton iteration. After the final iteration, the resulting `r` and `p` arrays come from evaluating the EOS. The current implementation skips the call to `init_isentropic_hse` and uses the `r_hse` and `p_hse` that are already provided to `init_custom_pert` (renamed from `init_custom_prob`). This is calculated using the saved hydrostatically balanced density `r` from the Newton iterations and then evaluating the hydrostatic equation rather than the EOS. These are only approximately equal because the tolerance for the Newton iterations = 1e-8 > 0.

ewquon:
@AMLattanzi here are the parts of the code relevant to our discussion from earlier today. I'm open to suggestions for how to improve this code structure!

`ProbParmDefaults`: https://github.com/erf-model/ERF/blob/6cdfc0847afd750ffa91d6ba963a2de31e8ca8c4/Source/prob_common.H#L10-L13

`ProblemBase.base_parms` update: https://github.com/erf-model/ERF/blob/6cdfc0847afd750ffa91d6ba963a2de31e8ca8c4/Source/prob_common.H#L148-L155

Problem-specific `ProbParm`, which I ended up inheriting from `ProbParmDefaults` to guarantee that it contains rho_0 and T_0: https://github.com/erf-model/ERF/blob/6cdfc0847afd750ffa91d6ba963a2de31e8ca8c4/Source/prob_common.H#L148-L155

Call to `ProblemBase::init_base_parms()` (note that this is only needed when `init_type="uniform"`, but included in all problems for generality): https://github.com/erf-model/ERF/blob/6cdfc0847afd750ffa91d6ba963a2de31e8ca8c4/Exec/RegTests/DensityCurrent/prob.cpp#L16-L29

ewquon:
Digging into this some more... Here is a detailed look into the initial differences (plt00000 from running the test suite) between the DensityCurrent cases. The fcompare output is the same for all 4 DensityCurrent tests; DensityCurrent_J2_* (not shown) have a different mesh than the baseline.

Perturbation pressure, before (top) / after (bottom) -- should be 0
DensityCurrent--pres_pert_plt00000

Perturbation density, before (top) / after (bottom)
DensityCurrent--dens_pert_plt00000

Difference (after - below), density (top) / theta (bottom)
DensityCurrent--diff_dens_theta_plt00000

The perturbation is identical but the magnitude of the background differences increase with height, consistent with the slightly different hydrostatic balance used in `init_custom_prob` from before. Furthermore, the differences show up in both density and pressure/theta because the previous implementation used the `init_isentropic_hse` vars (`r` and `p`) to calculate `Tbar_hse` and `theta_perturbed` rather than with `r_hse` and `p_hse` as in the new implementation. https://github.com/erf-model/ERF/blob/6cdfc0847afd750ffa91d6ba963a2de31e8ca8c4/Exec/RegTests/DensityCurrent/prob.cpp#L188-L218
(Note: In the previous version of the code, the inputs `r_hse` and `p_hse` are overwritten, but will have been reinitialized again with a final call to initHSE().)

ewquon:
The initial condition for the moving terrain cases actually are actually in agreement (MovingTerrain_sub and MovingTerrain_nosub are identical, as expected):
```
            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 density                                          0                         0
 x_velocity                                       0                         0
 y_velocity                                       0                         0
 z_velocity                                       0                         0
 pressure                           5.560286809e-08           5.561708359e-13
 theta                               1.24714461e-10           4.157103108e-13
 pres_hse                                         0                         0
 dens_hse                                         0                         0
 z_phys                                           0                         0
 detJ                                             0                         0
 PLOTFILE AGREE to specified tolerances: absolute = 1e-12 relative = 1e-12
```

The differences are only in theta, and mimic the differences seen with DensityCurrent. Because rho was previously initialized to r_hse, it is identical to the current implementation.
MovingTerrain--diff_theta_plt00000

ewquon:
For TaylorGreen, the plt00000 agree to machine precision. The Advecting and Advecting-Diffusing cases have identical ICs.
```
            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 density                                          0                         0
 x_velocity                                       0                         0
 y_velocity                                       0                         0
 z_velocity                                       0                         0
 pressure                           3.890221478e-13           3.698105732e-13
 temp                               1.355252716e-15           3.697491458e-13
 theta                              2.764455331e-14           2.852392779e-13
 scalar                                           0                         0
 PLOTFILE AGREE to specified tolerances: absolute = 1e-12 relative = 1e-12
```

Check theta error (rho_0=1.0)
```python
from erftools.EOS import getRhoThetagivenP
# rho_0 = V_0 = M_0 = 1
pmax = 1./1.4 + 6./16
orig = getRhoThetagivenP(pmax)
new = getRhoThetagivenP(p_0)
new += getRhoThetagivenP(pmax) - getRhoThetagivenP(p_0)
print(new - orig)
# result: 2.5146551507759796e-14
```

ewquon:
Shout to @dwillcox for helping sort out the GPU issues!

PR #1250, opened 2023-09-18
Hurricane QKE fixes

AMLattanzi:
1. Change IC to be homogeneous QKE (not homogeneous rho*QKE)
2. Fix double counting of 3 factor in SQ for PBL model


baperry2:
Good catch here.

PR #1252, opened 2023-09-19
QKE src fixes

AMLattanzi:
1. Dissipation must be subtracted from the RHS
3. Buoyancy source is \sim <w' \theta'> but -<w' \theta'> \sim k*dthetadz, so the buoyancy term should be negative
4. The GH term should have (g/theta_0) from [Nakanishi & Niino](https://link.springer.com/article/10.1023/B:BOUN.0000020164.04146.98); see Eq. (23).

This ran and was stable for many time steps and the QKE field looked reasonable.

### DIFF
--- a/Exec/DevTests/MiguelDev/inputs
+++ b/Exec/DevTests/MiguelDev/inputs
@@ -23,7 +23,7 @@ zhi.type = "SlipWall"
 
 # TIME STEP CONTROL
 erf.no_substepping = 0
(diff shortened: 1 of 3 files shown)


baperry2 (approved):
Looks good to me. Can add one more fix here (also runs stably for many timesteps): https://github.com/AMLattanzi/ERF/pull/1

PR #1258, opened 2023-09-28
Clean up box defs in Vel to Mom.

AMLattanzi:
When converting from velocity to momentum, we by default need one lateral ghost cell to be filled. With numerical diffusion, we need all of them filled (due to the stencil width). I also don't see why we should be growing a tilebox; this should lead to overlap in the thread operations (presumably safe since it's just writing but inefficient I think).


dwillcox:
Given that explanation, this makes sense to me and is much simpler!

dwillcox (approved):

PR #1261, opened 2023-10-05
Fix for failing WPS test.

AMLattanzi:
Summary of changes and rationale:

1. The indexing in **fill_from_wrfbdy** needed revamping. This is due to the fact that FillIntermediatePatch may call this routine under conditions when only cons vars should be filled or a subset of cons vars (see set_bcs after fast rhs). The new indexing makes the operations only occur on what we want to set.
2. The corner regions in **bx_xlo** and **bx_xhi** of **fill_from_wrfbdy** will incorrectly reach into a set region when the variable is NOT read. We want to copy from a valid interior cell.
3.  The deletion of **grids_to_evolve** will make the microphysics call **cloud()** in the set region. So the diagnosed quantities (qv, qc) will never match because we now operate on cells we didn't use to. We can remake the benchmark or we can exclude the qv and qc quantities.

### DIFF
--- a/Source/BoundaryConditions/BoundaryConditions_wrfbdy.cpp
+++ b/Source/BoundaryConditions/BoundaryConditions_wrfbdy.cpp
@@ -53,8 +53,7 @@ ERF::fill_from_wrfbdy (const Vector<MultiFab*>& mfs,
     Vector<int> comp_var = {ncomp_cons, 1, 1, 1};
 
     // End of vars loop
-    int var_idx_end = Vars::NumTypes;
-    if (cons_only) var_idx_end = Vars::cons + 1;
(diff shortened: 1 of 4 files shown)


asalmgren:
Just for conciseness let's consolidate lines like 

    int var_idx_end = Vars::NumTypes;
    if (cons_only) var_idx_end = Vars::cons + 1;

into 

int var_idx_end = cons_only ? Vars::cons+1 : Vars::NumTypes;

AMLattanzi:
Changes adopted in two places in [commit 37a67a8](https://github.com/erf-model/ERF/pull/1261/commits/37a67a8a9ae1ebcb740ec5136530179d813031f0)

asalmgren (approved):

PR #1272, opened 2023-10-22
Add GABLS1 case

ewquon:
- Define new problem to initialize QKE following the prescribed initial TKE in Cuxart et al. 2006
- Add `erf.most.surf_heating_rate` input parameter (units of K/hr), to be used with `erf.most.surf_temp` (as the initial value)
- Add output of eddy diffusivities to compute the PBL modeled covariances
- Numerically unstable if low Mach dt = 10 s (from Cuxart 2006) is used; the largest stable dt is 1.0 s w/ 6 substeps
- Need latest bug fixes from #1270, #1271


ewquon:
After 9h of simulation time, results insensitive to dt (as long as it's small enough), vertical spatial order, or Sq=2 vs 3 (default).
download-6

PR #1275, opened 2023-10-25
Fix MOST with grid stretching in z.

AMLattanzi:
When filling ghost cells with MOST, one must account for vertical grid stretching with terrain. The MOENG and Donelan flux computations are now passed a dz value that will either come from the geometry container (no terrain) or from the difference in z_phs at the surface (with terrain; knows about grid stretching).

### DIFF
--- a/Source/BoundaryConditions/ABLMost.H
+++ b/Source/BoundaryConditions/ABLMost.H
@@ -132,13 +132,16 @@ public:
     void
     impose_most_bcs (const int lev,
                      const amrex::Vector<amrex::MultiFab*>& mfs,
-                     amrex::MultiFab* eddyDiffs);
+                     amrex::MultiFab* eddyDiffs,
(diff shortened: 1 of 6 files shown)


ewquon (approved):

ewquon:
Verified using #1276 (https://github.com/erf-model/ERF/blob/e8150488fe71342ab728b18f5e74f63ab78d1cd6/Exec/ABL/inputs_most_pbl), with and without grid stretching, and different nz in the stretched case.

check_PR1275
* **blue** and **black** are expected to be _very nearly equal_, with the same surface dz (1024/64 = 16 m)
* the original bug is manifested in **orange**, which applies MOST with the wrong dz (1024/48 = 21.333 m)
* the fix from this PR is in **green**, which has the _exact same mesh as orange_ and agrees with the blue solution.

PR #1299, opened 2023-11-15
Adding qp for buoyancy type 1 

nataraj2:
Buoyancy type 1 should include the precipitation mass mixing ratio as well. Type 2 already does.


asalmgren:
@nataraj2 -- let's avoid the extra divide -- just move the RhoQp part outside of the parenthesis so you don't have to divide and multiply ...

PR #1348, opened 2023-12-15
Add terrain/stretched grid capability for PBL models

baperry2:
No diffs if there is no terrain. Still need to test for the case @ewquon was working on with PBL models + stretched grids.


baperry2:
going to merge so this doesn't fall out of date - doesn't change the behavior that currently works and is at least an improvement for what doesn't.

PR #1350, opened 2023-12-18
Add more moisture state variables

AMLattanzi:
First pass. Took a step with Bubble in debug.

@nataraj2 please continue to double check the microphysics classes and the hand off of variables to them.

Things to be completed:

1. The plotfile routine will be broken. This is due to the fact that **qt = qv + qc** with some moisture models but will be **qt = qv + qc + qi** with other moisture models. I propose that we make **qt** a quantity owned by the micro model and then we grab that to plot or output.
2. A moisture model with precipitation will need additional scalars? Does **qp** undergo a similar decomposition as **qt** or do we not advect/diffuse the precipitating components?
3. We should use state variable **q1 = qv**  for the Pert press computations. These will be updated each RK stage.
4. We should use state variables **q1, q2, and q3** for the buoyancy calcs (same rational).


AMLattanzi:
@nataraj2 I believe the perturbational pressure and the buoyancy has been taken care of. If you have the bandwidth, please double check the buoyancy.

**_Things left to do:_**
1. Verify the hand off of data to and from FE and Kessler classes
2. Verify the updates for FE and Kessler classes
3. Fix **Plotfile.cpp** --- the **qt** var should be owned by the micro class, so just grab the qt comp from mic_fab_vars and plot that. Double check for other logical issues.
4. Ice was not accounted for in the buoyancy routines nor anywhere else (not a priority but a note)

AMLattanzi:
@nataraj2 Everything should be done. Just test the plotfile with each model, review the code, and run a few test cases. Thanks in advance!

PR #1361, opened 2024-01-04
Updates for terrain / grid stretching

ewquon:
* Fix bug w/ higher-order advection + terrain (https://github.com/erf-model/ERF/pull/1361/commits/ed6ae8c59ed1ba786dee49cdb6a49c9e22ecc3f6)
* Clean up problem-specific terrain code
* Warn if mismatched `prob_hi[2]` and `zlevels_stag[nz]` -- the latter is the stretched grid levels that are used to populate `z_phys_nd` and may differ from the input geometry if "grid_stretching_ratio" or "terrain_z_levels" is specified
* Use `zlevels_stag[nz]` in `init_terrain_grid()` instead of `ProbHiArray[2]` -- only has an effect if surface elevation `z_arr(ii,jj,k0) != 0`
* Enable terrain for idealized case with input sounding


ewquon:
Draft for now, need to verify that the new default `init_custom_terrain()` doesn't break any reg tests...

ewquon:
Note: regtests fail for DensityCurrent_detJ2 and DensityCurrent_detJ2_nosub (both with terrain, default 3rd order) fail starting from [ed6ae8c](https://github.com/erf-model/ERF/pull/1361/commits/ed6ae8c59ed1ba786dee49cdb6a49c9e22ecc3f6), i.e., after the bug fix.

ewquon:
https://github.com/erf-model/ERF/pull/1361/commits/302d57e414418ca77dd3694293d147d0bce7f066 enables ideal input soundings to be used with terrain, resulting in comparable pressure and density base states with and without grid stretching. The commit message describes the implementation. Some additional comments:
* The nominal grid levels, either supplied by the user through `terrain_z_levels` -- or `grid_stretching_ratio` and `initial_dz` -- are the "coordinate surface heights" ($\zeta$) from Klemp 2011
* $\zeta$ is expected to range from 0 (at the surface )to $z_{top}$
* With BTF and Sullivan TF terrain grids, the physical height of the highest grid nodes is _exactly equal_ to $z_{top}$ at k=geom[0].bigEnd(2)+1, (where $\zeta=z_{top}$).
* With STF terrain grids (smoothing following Klemp 2011), the grid should be deformed near the surface only (see https://github.com/erf-model/ERF/issues/1368). The current implementation assumes the domain top is far enough away from the surface to be outside of the deformed grid region, in which case, the physical height of the highest grid nodes is also equal to $z_{top}$.

ewquon:
This PR looks bigger than it actually is. I consolidated all the duplicated terrain initialization code (that set the surface z_phys_nd = 0) and cleaned up most of the prob source files.

@asalmgren a heads up that the gold data were updated!

PR #1377, opened 2024-01-11
Fix issue with STF smoothing.

AMLattanzi:
The max height was omitted from the z_h term when implementing STF smoothing for terrain-following height coordinates.


AMLattanzi:
This will be reopened after the LSM data structures are in place and I have time to address the divide by 0 case it creates.

PR #1378, opened 2024-01-12
Particle WoA GPU fix

AMLattanzi:
Promote member functions to public to avoid extended lambda capture error.

### DIFF
--- a/Source/Particles/ERFPC.H
+++ b/Source/Particles/ERFPC.H
@@ -131,6 +131,17 @@ class ERFPC : public amrex::ParticleContainer<  ERFParticlesRealIdx::ncomps,
             m_advect_w_gravity = a_flag;
         }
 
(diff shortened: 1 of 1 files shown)


debog:
@AMLattanzi What does this mean, and why do these functions need to be public? I am just starting to learn about lambda expressions and the capture rules.

(Also tagging @gardner48 since we were trying to figure this out)

WeiqunZhang:
This is a CUDA (not HIP or SYCL) specific restriction on extended lambda.  See https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#extended-lambda-restrictions No. 4.

debog:
Thanks, @WeiqunZhang! So extended lambda just means the extension of the lambda syntax in CUDA to allow specifying host and/or device?

WeiqunZhang:
That's correct.

PR #1383, opened 2024-01-16
Fix oops

asalmgren:


Source/Particles/ERFPCEvolve.cpp
```diff
@@ -130,10 +129,16 @@ void ERFPC::AdvectWithFlow ( MultiFab*                           a_umac,
                     iv[1] += domain.smallEnd()[1];
```
debog, line 129: @AMLattanzi @asalmgren Can lines 124-129 be replaced by `ERFParticlesAssignor(p, plo, dxi, domain)` (defined in ERFPC.H)? Same with lines 226-231 below?

Source/Particles/ERFPCEvolve.cpp
```diff
@@ -98,8 +98,7 @@ void ERFPC::AdvectWithFlow ( MultiFab*                           a_umac,
             bool use_terrain = (a_z_height != nullptr);
             auto zheight = use_terrain ? (*a_z_height)[grid].array() : Array4<Real>{};
```
debog, line 99: @asalmgren @AMLattanzi how does this make `zheight` available on the GPU?

Source/Particles/ERFPCEvolve.cpp
```diff
@@ -130,10 +129,16 @@ void ERFPC::AdvectWithFlow ( MultiFab*                           a_umac,
...
+                               zheight(iv[0]+1,iv[1]  ,iv[2]  ) *      lx  * (1.0-ly) +
+                               zheight(iv[0]  ,iv[1]+1,iv[2]  ) * (1.0-lx) * ly +
+                               zheight(iv[0]+1,iv[1]+1,iv[2]  ) *      lx  * ly;
+                        zhi =  zheight(iv[0]  ,iv[1]  ,iv[2]+1) * (1.0-lx) * (1.0-ly) +
+                               zheight(iv[0]+1,iv[1]  ,iv[2]+1) *      lx  * (1.0-ly) +
+                               zheight(iv[0]  ,iv[1]+1,iv[2]+1) * (1.0-lx) * ly +
+                               zheight(iv[0]+1,iv[1]+1,iv[2]+1) *      lx  * ly;
                     } else {
                         zlo =  iv[2]    * dx[2];
                         zhi = (iv[2]+1) * dx[2];
```
debog, line 144: @AMLattanzi @asalmgren Can lines 130-145 (or 130-151) be written as a separate function/functor? If functor, it could store `zheight` and `use_terrain` as member data? And we could use it for lines 232-252 as well.

debog:
@asalmgren @AMLattanzi If you agree with my comments above, then I can try to implement them and send a pull request.

@gardner48 Tagging you so that you get notified as well.

Source/Particles/ERFPCEvolve.cpp
WeiqunZhang: This is due to C++ lambda capture. If a C++ lambda uses `[=]`, it will make a copy of all the variables used in the lambda's function body. These variables are part of the lambda function object. In the case of GPU, the CUDA runtime will copy the lambda function to GPU at kernel launch.

Source/Particles/ERFPCEvolve.cpp
asalmgren: Actually I want to move this functionality into a routine in amrex -- it is used by multiple codes.   I was just going to make a new function called something like update_particle_idata that takes one particle and the zheight array.   What do you think?

Source/Particles/ERFPCEvolve.cpp
debog: Thanks, @WeiqunZhang! Makes sense now.

Source/Particles/ERFPCEvolve.cpp
debog: That would be nice since this is a particle functionality, not an ERF-specific thing. But terrain and zheight are specific to atmospheric flows. To make it more general, could we have xcoord, ycoord, zcoord (defaulted to nullptr?) instead of just zheight that allows mapping in all dimensions? ERF would then call this function with only zheight.

asalmgren: We're working on the generalized functionality ...

AMLattanzi: I think we could parallel what is done in the [BinMapper](https://github.com/AMLattanzi/amrex/blob/2fc8a34ced0c0acae04298c69ec23803ec05e9cd/Src/Particle/AMReX_ParticleUtil.H#L295-L338) functor. This would move the functionality to AMReX, clean up the ERF code since the above snippet now lives in a functor, and would be similar to what we do without curvilinear coordinates.

PR #1391, opened 2024-01-19
Refactor ConstantAlpha to allow running with turbulence model

ewquon:
Previous approach:
* Calculate shear stress (tau), which in general is strain rate multiplied by 2*(mu + mu_t) -- here mu and mu_t refer to the molecular and turbulent viscosities
* When calculating the diffusion contribution, multiply by rho/rho0 if using ConstantAlpha -- here, rho is the instantaneous local value, _averaged from cell centers to faces_
* The issue arises because mu_t also ends up being scaled by this density ratio

New approach
* Calculate shear stress with appropriate mu_eff, based on whether we're using the Constant or ConstantAlpha solver path -- here, tau is scaled by the _cell-centered or face-averaged_ value of rho, depending on the stress quantity
* Diffusion contribution calculations are agnostic to the molec_diff_type
* **Because the density used to scale the diffusion contribution is different, the gold data for regtests with ConstantAlpha will change**

Tests based on DensityCurrent
* ConstantAlpha with `les_type=none`: `dynamicViscosity=75`, `rho0_trans=1` (default) vs `dynamicViscosity=150`, `rho0_trans=2` -- verified results were identical
* ConstantAlpha with `les_type=Smagorinsky`: `dynamicViscosity=75`, `rho0_trans=1` (default) vs `dynamicViscosity=150`, `rho0_trans=2` -- verified results were identical
* very small differences observed with and without Smagorinsky in this case


ewquon:
Density current prob (coarse dx=dz=100 m) at 900 s, before/after this PR
compare_before_after_PR1391

Source/Diffusion/ComputeStress_N.cpp
```diff
@@ -20,32 +21,60 @@ using namespace amrex;
...
+        ParallelFor(tbxxy,tbxxz,tbxyz,
+        [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept {
+            tau12(i,j,k) *= -cell_data(i, j, k, Rho_comp) * mu_eff;
+        },
+        [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept {
+            tau13(i,j,k) *= -cell_data(i, j, k, Rho_comp) * mu_eff;
+        },
+        [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept {
+            tau23(i,j,k) *= -cell_data(i, j, k, Rho_comp) * mu_eff;
+        });
```
AMLattanzi, line 59: For the off-diagonal stresses, shouldn't the **cell_data** be averaged to the location of the strain (this would hold for other places in the code as well)?
image

AMLattanzi (changes requested):
Fairly certain the cell_data must be averaged to the strain locations (as is done in **DiffusionSrcForState_N/T.cpp**. Once that is completed, I am good with this PR.

Source/Diffusion/ComputeStress_N.cpp
ewquon: Yes, you're right. I have a fix coming.

ewquon:
@AMLattanzi, [23d55a5](https://github.com/erf-model/ERF/pull/1391/commits/23d55a50e6f9ca3ae393f319f1eab6055a027799) fixes the segfault that showed up in the CI and [1834373](https://github.com/erf-model/ERF/pull/1391/commits/18343734fc5d21f155740303fd5c8ab04ee75357) does the averaging that's needed.

ewquon:
Actually hold on, I only updated the terrain solver path.

ewquon:
Ok, @AMLattanzi, this latest commit makes the rest of the changes. I've verified that the rho averaging for the ConstantAlpha path matches the original mu_turb averaging.

AMLattanzi (approved):

PR #1399, opened 2024-01-25
Terrain with Rayleigh damping

ewquon:
Add `z_phys_cc` to parameter list for `erf_init_rayleigh`; update default implementation to handle cases with and without terrain. Caveats:
1. Previously, the damping strength was determined by the distance from the top of the domain; now the damping strength is determined by the distance from the top-most cell height. This simplifies the implementation with terrain and also ensures the maximum damping value is attained within the computational domain.
2. Damping is still applied uniformly at each k-level, even with terrain. The implicitly assumes that where the damping is active, the grid is approximately uniformly spaced. Because the terrain heights at each k are not necessarily constant, the _maximum_ value is taken to represent that level.

The Witch of Agnesi hill example has been updated to allow damping. Example results (hill height=1, zhi=2) without damping:
WoA_undamped
and with damping (zdamp=1, noting that the resulting damping will increase gradually from z=1 to z=2; dampcoef is chosen to have a time constant equal to the simulation length):
WoA_damped

### DIFF
--- a/Exec/DevTests/MovingTerrain/prob.H
+++ b/Exec/DevTests/MovingTerrain/prob.H
@@ -49,7 +49,8 @@ public:
         amrex::Vector<amrex::Real>& vbar,
         amrex::Vector<amrex::Real>& wbar,
         amrex::Vector<amrex::Real>& thetabar,
-        amrex::Geometry      const& geom) override;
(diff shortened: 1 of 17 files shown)


ewquon:
Argh, accidentally fubard the merge.

PR #1408, opened 2024-01-28
Specified Lateral BC Reformat

AMLattanzi:
This encompasses a detailed unit test to verify that the lateral specified and relax zones are properly computed. The paradigm is then extended to the QV variable in the post update. This remained well bounded in z velocity for 1000 steps with the FLOWMAS simulation set up. It does NOT use a forward prediction for the relaxation operations; this avoids the need for a fillboundary in the RHS.


AMLattanzi:
Inputs file and read from metgrid need to be returned to original state.

AMLattanzi:
New formulation results with FLOWMAS simulation (1000 steps; 5 relax cells; Null moisture). The z velocity remains well bounded, but the banding is present just like in WRF. The banding is not prevalent in the other components.
image


New formulation results with FLOWMAS simulation (200 steps; 5 relax cells; Kessler moisture). The z velocity remains well bounded, but the banding is highly accentuated with moisture. However, the banding is not observable in the Qv variable it self. Work remains to be done with moisture and lateral BCs.

image
image

AMLattanzi:
@nataraj2 This heavily follows what you have been utilizing. It just doesn't zero out the RHS in the set region (specifies the true tendency) and it doesn't use the forward prediction for relaxation (this avoids a fillboundary call for the RHS, which would be cumbersome with moisture).

PR #1397, opened 2024-01-25
Particle container fixes

debog:
+ Made the particle advection functions virtual.
+ Added the new particle files to BuildERFExe.cmake (this was missed earlier)
+ Temporary bugfix in updating particle k-index

PR #1410, opened 2024-01-31
Bugfix for particle advection without terrain

debog:
The changes in pull request #1397 caused a segfault when advecting particles with no terrain.


asalmgren:
@debog -- one of the amrex style guidelines we try to follow is that every if or else use brackets to enclose whatever follows -- could you make that fix in the Evolve routine?  Thx!

debog:
@asalmgren Done!

PR #1411, opened 2024-01-31
MetGrid Moisture

AMLattanzi:
The banding with moisture and lateral forcing is corrected with this PR. The origin of the banding was due to an imbalance between the pressure and rho*g with moisture. Correcting that balance yields the following result with moisture and 5 relax cells over 150 steps with the FLOWMAS simulation:
image (1)


AMLattanzi:
@wiersema1 I would like you to review these changes and we can discuss the clean up that will be needed to finish this up.

PR #1413, opened 2024-01-31
WIP: Fitch wind farm parametrization

nataraj2:
This PR introduces the Fitch parametrization for wind farms 
[Fitch model 2012](https://journals.ametsoc.org/view/journals/mwre/140/9/mwr-d-11-00352.1.xml)
This parametrization for wind farms involves adding source terms to update the x and y velocities as well the turbulent kinetic energy in the mesh cells where the wind turbines are located. `Source/WindFarmParametrization/Fitch` contains the Fitch model source terms. The paramterization is a runtime option - `windfarm_type`. Additional parametrizations will be added later.

An example is provided in `Exec/Fitch`. The turbine locations are in the attached figure and the movie shows the evolution of the velocity magnitude at the hub height of the turbines of `z=100 m`. All other features such as dissipation and turbulence have been turned off in `inputs_Fitch` to demonstrate the feature.

To do:
1. Read in the parameters from the WRF tbl file.
2. Introduce drag coefficient for the turbines as in WRF.
Fitch_WindTurbineLocations
Fitch_vel_abs


asalmgren:
in fitch_update it would be good to put all those loops into a single ParallelFor (over multiple boxes).

asalmgren:
Also -- we will want to implement other models that do similar things to Fitch but are somewhat different models -- do you want to go ahead and put that generality in now so the choice of parametrization can be a run-time parameter?

PR #1424, opened 2024-02-03
MetGrid balance HSE with moisture.

AMLattanzi:
Balance the HSE with moisture and MetGrid initialization. @wiersema1 please review this code for our meeting Monday.

The metgrid init (top) is now well balanced but the WRF init (bottom) shows finite perturbational pressure. **This needs to be understood.**
image

Removing the effect of moisture (qv = 0 in plotfile.cpp) does not explain the non-zero perturbational pressure with wrf init:
image

Retaining the moisture effect (qv value used in plotfile.cpp) does not show strong correlation between qv and pert pressure:
image


AMLattanzi:
Here is the pert pressure with MetGrid BEFORE these changes (does not balance with moisture):
image

PR #1427, opened 2024-02-06
Balance pressure.

AMLattanzi:
The base state from wrf input does not yield well behave perturbational pressures. We correct that here.


AMLattanzi:
Comparison of MetGrid init versus WRF init (with this draft PR) (100 steps) `Z vel min/max: -1.14/1.5 -0.95/1.2`
NOTE: The wrf bdy does not exactly match the initial state like it does with MetGrid initialization. This is due to the RhoTheta variable, which has ~0.23 diff in the set/nudging regions when compared to the initial field in those regions. The largest errors between the bdy and initial state are in the large red region in the right figure (pert pressure jump to ~ 36 in that region).
image (4)

Comparison of WRF init (development) versus WRF init (with this draft PR) (100 steps) `Z vel min/max: -3.30/40.01 -0.95/1.2`
image (5)

AMLattanzi:
By forcing the BDY and initial state to match with the WRF input case, a direct comparison with MetGrid initialization can be approximately drawn. The same larger scale structures are observed in both simulations. The MetGrid (left) versus the WRF Init w/ IC=BDY (right) `200 steps & Zvel min/max: -0.76/0.71 -0.46/0.52`

NOTE: The MetGrid case linearly extrapolates all the data above 18000 m since that is the top level in the MetEm file. What the WRF pre-processor does with these points is not clear right now.
image

AMLattanzi:
Coalescing the WRF and MetGrid BC operations and width vars allows one to restart the `WRF init simulation` from a chk file created by a `MetGrid init simulation`. This ensures the bdy and init states match. Using the following run `./ERF3d.gnu.TEST.MPI.ex inputs_flowmas_wrfinput erf.restart=chk00000 max_step=25` (where chk00000 is generated by the MetGrid init simulation). We get the following fcompare:

image

AMLattanzi:
Using MOST at the bottom boundary yields the following results compared to WRF, for the FLOWMAS simulation after 15 minutes of simulation time.

WRF (left) and ERF (right) at k==1
image
image

WRF (left) and ERF (right) at k==20 (midline)
image
image

PR #1460, opened 2024-02-27
Modifications for BOMEX

AMLattanzi:
This adds a `custom` MOST option where constant fluxes can be specified, which allow the BOMEX simulations to be completed. This should be generalized to allow for moisture fluxes but a similarity theory or Cd closure is needed for that.


AMLattanzi:
@chandru320 A test of the code was completed with 4 processors in DEBUG mode for 10 steps with `ABL/inputs_most_bomex`. The ghost cells were printed at `(0,0,k<=0)` to confirm appropriate values. The plt files confirm u velocity is decreasing while theta and qv are injected into the system:
image
image
image

PR #1466, opened 2024-02-28
Moisture source term profiles for BOMEX

chandru320:
This creates a BOMEX test case, copied from TemperatureSource, with a modified input file and input_sounding file. 
Moisture source terms are added based on a profile (using cutoff and cutoff transition height)


chandru320:
The temperature and moisture source terms work as expected. The source terms are cooling rate and drying rate, and would reduce temperature and moisture in the domain.

qv_1min
Tempertaure_1min

PR #1469, opened 2024-02-29
Moist Input Sounding

AMLattanzi:
The PR modifies the input sounding creation of the base state. Namely, we must balance with moisture when initializing from an input sounding. This ties into the virtual potential temperature vs moist potential temperature. We use the moist potential temperature in the EOS, so we must use that to balance. Furthermore, the rho_hse needs to include the moisture.

We can use `abl_mwe_moist.i` with this PR to see the pert pressure is 0 while with the development branch it was not:
image (2)
image (1)


ewquon:
I believe the root cause of this issue is my implementation of the input sounding initialization based on _virtual_ rather than moist potential temperature. The EOS functions were later updated to handle qv (e.g., `getRhogivenThetaPress`), and those are based on _moist_ potential temperature.

AMLattanzi:
Reducing `qv` to prevent condensation and setting `u/v/w = 0.0` for the IC shows the following z velocity profile after 2 steps. Printing the buoyancy and pert pressure confirms that the system is balanced (both are 0 for a all k at a fixed i and j)
image

PR #1471, opened 2024-03-01
fix some bugs in radiation code

xyuan:
this is the first fixes for the radiation code, more pr will be created later.


xyuan:
my argonne email is not allowed to access the slack, please use github issues for the radiation code.

PR #1472, opened 2024-03-04
Provide alternate solution path for imposing MOST

ewquon:
The new "explicit MOST" solution path is activated with the `ERF_ENABLE_EXPLICIT_MOST_STRESS` option (or `-DERF_EXPLICIT_MOST_STRESS` compile flag for gmake). This eliminates the dependence of the boundary condition on the surface eddy diffusivity and consequently eliminates the circular dependence when using the Deardorff SFS model (wherein the surface BC depends on the the eddy diffusivity and the eddy diffusivity also depends on surface conditions). This is also expected to improve log-law agreement (e.g., non-dimensional shear) near the surface.

This new option is closer to the Moeng 1984 formulation. Instead of using the effective viscosity and modeled surface stress to set the velocity BC, we set Neumann conditions based on the velocity gradient determined at the first cell face (i.e., between klo and klo+1). Moeng goes one step further and interpolates between this resolved velocity gradient and the modeled velocity gradient (from the similarity formulas) evaluated at a location below the first cell height. The MOST stress is then directly applied by setting the surface values of Tau13 and Tau23. We effectively have a shear-stress wall BC and not a velocity wall BC. Ultimately, the velocity gradient that is imposed dictates the first-cell velocity gradient that determines TKE production.


ewquon:
@AMLattanzi, can you double check the fixes in https://github.com/erf-model/ERF/pull/1472/commits/f865355f9ce29d01cc98417cd69c677e98cc0996? Thanks!

ewquon:
Verified:

- [x] MOST calculated stresses and heat flux are untouched when used in DiffusionSrcForState/Mom
- [x] Velocity gradients used to extrapolate to ghost cells match du/dz and dv/dz in ComputeStrain (both between RK stages and from one timestep to the next)
- [x] Temperature gradient used to extrapolate to ghost cells matches dtheta/dz in ComputeTurbulentViscosity for Deardorff
- [x] Surface strains S13, S23 match strains at k=1
- [x] du/dz and dv/dz are equal at the first two levels; dw/dx = dw/dy = 0 on the surface
- [x] SmnSmn is using the correct strain

ewquon:
Adding this here for reference (FYI @hgopalan, @mchurchf). Uniform ~10m resolution, Smagorinsky. Moeng and Sullivan 1994 shear-driven case. Averaging between 7200-10800 s (MS94 ran for 15225 s, and that's the ref data here, but I averaged where there was the best agreement between simulated ustar).

Shown are the 2 advection schemes, WRF-like (5th-order/3rd-order upwind in the horiz/vert directions) and 2nd-order central (with 10% 6th-order numerical diffusion thrown in for grins -- this wasn't actually needed to stabilize the solution and doesn't actually clean up the numerical noise in the free atmosphere... something like #1482 is still needed). 

* "set vel" -- MOST BC sets the ghost velocity based on the stress-strain relationship with modeled eddy viscosity
* "set stress" -- the new solution path here, along the lines of Moeng 1984. Surface stress and strain are decoupled. Drag from x,y-momentum diffusion uses the surface MOST stress; the SGS model uses the near-surface strain-rate. 

MS94_Smag_dx10_winds

[inputs_MS94_S.txt](https://github.com/erf-model/ERF/files/14541781/inputs_MS94_S.txt)
[input_sounding.txt](https://github.com/erf-model/ERF/files/14541782/input_sounding.txt)

ewquon:
@AMLattanzi I think this is good to merge. I say we bring it in as is and not try to refactor at this point. (I cleaned up some of the ifdefs, if that helps)

Without more sophisticated SGS modeling, results appear reasonable in comparison with other codes, e.g., AMR-Wind (not shown) or SOWFA (Churchfield et al 2010, AMS BLT, shown below).
image

AMLattanzi (approved):

PR #1476, opened 2024-03-05
Hydrometeors in PlotFile.cpp

nataraj2:
This PR changes the writing of hydrometeors in `PlotFile.cpp` by taking the conservative variables corresponding to the hydrometeors and dividing by density (ie. `rhoQs/Rho`). We follow the following pattern in arranging hydrometeors:

`rhoQ1` - water vapor
`rhoQ2` - cloud water
`rhoQ3` - rain
`rhoQ4` - ice
`rhoQ5` - snow
`rhoQ6` - graupel

For no precipitation, only first 2 components in the list exists. With rain, the first 3 components exist, and with cold microphysics, all the components exist. Some change is needed in the `Init` routines of the microphysics classes and `prob.cpp` of the test cases to reflect this change, and will be done in another PR soon.

### DIFF
--- a/Exec/SquallLine_2D/inputs_moisture
+++ b/Exec/SquallLine_2D/inputs_moisture
@@ -38,7 +38,7 @@ amr.check_int       = 1000       # number of timesteps between checkpoints
 # PLOTFILES
 erf.plot_file_1         = plt        # root name of plotfile
 erf.plot_int_1          = 60         # number of timesteps between plotfiles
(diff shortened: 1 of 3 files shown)


debog:
@asalmgren This merge broke the [moist bubble](https://github.com/erf-model/ERF/blob/development/Exec/RegTests/Bubble/inputs_BF02_moist_bubble) case. I was able to run it by removing `qt` from the plot variables. I guess all input files with `qt` need to be updated.

asalmgren:
@nataraj2 -- I'm assuming we should be able to plot qt still?   Can you fix?

AMLattanzi:
This should be corrected in [PR 1480](https://github.com/erf-model/ERF/pull/1480). There was also an index error corrected that would only impact SAM.

PR #1477, opened 2024-03-05
Extending microphysics interface for Lagrangian moisture models

debog:
+ Added a Lagrangian moisture model base class.
+ Number of `qmoist` multifabs can now vary by AMR level; for Eulerian moisture models, they should be the same at all levels; for a Lagrangian model, they exist only for the base level.
+ Microphysics `advance` function needs the state variables and terrain (to pass on to Lagrangian models).


Source/Microphysics/LagrangianMicrophysics.H
```diff
@@ -0,0 +1,141 @@
...
+#ifdef ERF_USE_PARTICLES
+
+#include <utility>
+#include <string>
+
+#include "NullMoistLagrangian.H"
+#include "Microphysics.H"
+
+/* forward declaration */
+class ERFPC;
```
debog, line 17: @asalmgren @AMLattanzi Instead of forward declaring, should I define `LagrangianMicrophysics` as a templated class with the particle container type as the template parameter? Is that better coding?

Source/Microphysics/Null/NullMoistLagrangian.H
```diff
@@ -0,0 +1,54 @@
...
+
+#ifndef NULLMOISTLAGRANGIAN_H
+#define NULLMOISTLAGRANGIAN_H
+
+#ifdef ERF_USE_PARTICLES
+
+#include "NullMoist.H"
+
+/* forward declaration */
+class ERFPC;
```
debog, line 12: @asalmgren @AMLattanzi same question here: should I make `NullMoistLagrangian` templated on the particle container type?

PR #1485, opened 2024-03-08
Bomex Subsidence Src

AMLattanzi:
This PR adds custom subsidence source terms for the `\phi = {Theta; u; v; qv}` variables. The approach is to specify a field for `\bar{w}` and then then planar average `\bar{\phi}`. The source terms become `\bar{w} \partial \bar{\phi} / \partial z`.

The z derivatives cause a bit of an issue with the PlaneAverage class since we must have `\bar{\phi}` inside the ghost cells --- e.g. `src_{k=0} = \bar{w}_{k=0} * 0.5 * (\bar{phi}_{k=1} - \bar{\phi}_{k=-1}) * dzInv`. The solution to this was to allow PlaneAverage class to operate on the ghost cells but to include an offset when doing so since PlaneAverage uses a Host and Device vector (index starts at 0). We then transfer the Vector data to TableData so we can access at k<0 and k>N.

Finally, modifications were made to PlanAverage to make sure the average is correct with data that is NOT cell centered. This involved correctly computing the number of points in the domain and using a mask to ensure we do no double count face/node data with multiple processors.


AMLattanzi:
@asalmgren please focus on PlaneAverage.H and particularly the modification of `m_ncell_line/plane` as well as the use of the `OwnerMask`.

AMLattanzi:
@chandru320 Please review this code and test to make sure it does what you expect. Please post results to this PR like before.

asalmgren (approved):

chandru320:
qv_10min
Temperature_10min
u_10min

PR #1492, opened 2024-03-12
Nudging Mods

AMLattanzi:
WRF (left at 20 min) vs ERF (right at 15 min) at the bottom plane z==0.  Limiting CC var width to 1 less than FC var width allows pressure gradients to propagate through the nudging region. The exponential decay also ensures the nudging has minor effect at the edge of the relaxation zone.
image (1)


AMLattanzi:
Analogous simulation with Metgrid and 10 relaxation cells:
Screenshot from 2024-03-13 13-44-52

PR #1497, opened 2024-03-14
Lat/Lon & Real bounds check

AMLattanzi:
1. Get latitude and longitude of the lower left corner of the domain 
2. Bounds check with met and wrf init in the x and y direction
3. Bounds check the user specified z_top is close to min/max of z_hi from wrf input

- Need to verify on GPU since I capture by reference with the ParFor


AMLattanzi:
@nataraj2 Please verify this suits your needs for the wind farm locations.

PR #1514, opened 2024-03-20
fix particle initialization so all particles are actually in the doma…

asalmgren:
…in; also add the ability to specify where particles are created by defining a box in physical space


Source/Particles/ERFPC.H
```diff
@@ -215,12 +216,17 @@ class ERFPC : public amrex::ParticleContainer<  ERFParticlesRealIdxAoS::ncomps,
...
         void initializeParticlesUniformDistribution (const std::unique_ptr<amrex::MultiFab>& a_ptr = nullptr);
+        void initializeParticlesUniformDistributionInBox (const std::unique_ptr<amrex::MultiFab>& a_ptr,
+                                                          const amrex::RealBox& particle_box);
 
     protected:
 
         bool m_advect_w_flow;               /*!< advect with flow velocity */
         bool m_advect_w_gravity;            /*!< advect under gravitational force */
 
+        amrex::RealBox particle_box;
```
debog, line 227: Can you rename this to `init_particle_box` or `m_particle_box`? I imagine `particle_box` will get used as a local variable name at various places.

Source/Particles/ERFPCInitializations.cpp
```diff
@@ -35,6 +59,8 @@ void ERFPC::InitializeParticles (const std::unique_ptr<MultiFab>& a_height_ptr)
 
     if (m_initialization_type == ERFParticleInitializations::init_default) {
         initializeParticlesDefault( a_height_ptr );
+    } else if (m_initialization_type == ERFParticleInitializations::init_box) {
```
debog, line 62: I think we can get rid of `initializeParticlesDefault()` and consequently, `initializeParticlesDefaultTracersWoA()` and `initializeParticlesDefaultHydro()`. They can now be realized by `initializeParticlesUniformDistributionInBox()` (and they will anyway not be used). 

The default initialization can then be `initializeParticlesUniformDistributionInBox()`. If needed, the initial `particle_box` can be set to the appropriate values if the particle species name is `tracer_particles` or `hydro_particles` (but I wouldn't bother).

asalmgren: I'm going to leave this to you to clean up as you see fit ... do whatever makes most sense for your targets :-)

PR #1520, opened 2024-03-21
Geostrophic wind profile

chandru320:
Creates a geostrophic wind profile as a function of altitude and implement as source terms to momentum.


asalmgren:
Just want to make sure you know about the commands that make it easy to eliminate tabs and traiing white space ... 

ERF/.github/workflows/style/check_tabs.sh

and

ERF/.github/workflows/style/check_trailing_whitespaces.sh

:-)

PR #1548, opened 2024-04-02
Boundary Register Only Inflow

AMLattanzi:
The boundary register data should only be used on a face if that face is defined as an external Dirichlet ingested `ext_dir_ingested`. Also, I've cleaned up some async copies to the GPU; there is no reason to have the compile macros switch between an async gpu copy and memcpy because the AMReX gpu container already does this for us.

Results **before** change (perturbed flow fed into uniform u=10 IC):
image (3)

Results **after** change (perturbed flow fed into uniform u=10 IC):
image (2)

### DIFF
--- a/Exec/ABL/inputs.read
+++ b/Exec/ABL/inputs.read
@@ -41,7 +41,7 @@ erf.check_int       = -1        # number of timesteps between checkpoints
 
 # PLOTFILES
 erf.plot_file_1     = plt       # prefix of plotfile name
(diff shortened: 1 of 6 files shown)


asalmgren:
@AMLattanzi -- would you be willing to clean up the remaining Gpu::htod_memcpy_async calls in BoundaryConditions/BoundaryConditions*cpp routines?  Might as well do them all!

PR #1556, opened 2024-04-05
fix minor MOST zref bug

baperry2:
Remove displacement of 1/2 cell when computing index of zref that I believe to be unnecessary.

Example: if current zref requested is 14 and dz is 10, the returned index to average over would be 0 rather than 1.

### DIFF
--- a/Source/BoundaryConditions/MOSTAverage.cpp
+++ b/Source/BoundaryConditions/MOSTAverage.cpp
@@ -270,7 +270,7 @@ MOSTAverage::set_k_indices_N()
             AMREX_ASSERT_WITH_MESSAGE(m_zref <= m_zhi - 0.5 * m_dz,
                                       "Query point must be below the last z-cell!");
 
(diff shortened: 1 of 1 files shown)


AMLattanzi (commented):
All of the data relevant to the MOST BC ({theta, u, v, umag}) is aliased off of the bottom boundary by 0.5 dz. Therefore, the relation between `zref` and the `k` index is given by `L275`. Note that we assert that `zref > 0.5 dz`. 

Consider if a user specifies a `zref` that is `1.01dz` in height with `zlo = 0`. For the current implementation, we would arrive at `k=1`, which shifts the reference point up to `zref=1.5dz`. For the old implementation we would round down to `k=0` and have `zref=0.5dz`.  The old implementation is not incorrect but is overly conservative in that you have to exceed the new CC point before we accept it as the k index.    

https://github.com/erf-model/ERF/blob/8d26e2c356dce4af035931ab78adcf4b53e59a42/Source/BoundaryConditions/MOSTAverage.cpp#L273-L275

AMLattanzi (approved):

PR #1562, opened 2024-04-05
EXPERIMENTAL: Implement thin immersed boundary forcing, clean up incompressible path

ewquon:
Thin body modeling is for infinitesimally thin surfaces in a flow (e.g., solar collectors), verified in 2D for a thin airfoil under the compressible path. The incompressible path works without the thin-body forcing, verified for the advecting isentropic vortex problem.


ewquon:
FYI @tonyinme

PR #1593, opened 2024-04-30
Rad pressure units

AMLattanzi:
Use pressure units of `Pa` since RRTMGP converts for us. See [E3SM-Project/rte-rrtmgp minor](https://github.com/E3SM-Project/rte-rrtmgp/blob/ddf82a25a3a59a8f0fe904b69181cb7bd99881fb/cpp/rrtmgp/kernels/mo_gas_optics_kernels.cpp#L263). The long wavelength [example case](https://github.com/E3SM-Project/rte-rrtmgp/blob/ddf82a25a3a59a8f0fe904b69181cb7bd99881fb/examples/rfmip-clear-sky/rrtmgp_rfmip_lw.F90#L266) was used to verify this through print statements. Note that in the example case the state (pressure, temp, composition) is read in from a netCDF file `multiple_input4MIPs_radiation_RFMIP_UColorado-RFMIP-1-2_none.nc` and `ncdump` also confirms the units of pressure are `Pa`. Finally, hacking the ERF code and example case allows the long wavelength results to agree well between the two codes. Short wavelength still needs to be addressed.

### DIFF
--- a/CMake/BuildERFExe.cmake
+++ b/CMake/BuildERFExe.cmake
@@ -77,6 +77,7 @@ function(build_erf_lib erf_lib_name)
 
   if(ERF_ENABLE_RRTMGP)
     target_sources(${erf_lib_name} PRIVATE
+                   ${SRC_DIR}/Utils/Orbit.cpp
                    ${SRC_DIR}/Radiation/Init_rrtmgp.cpp
                    ${SRC_DIR}/Radiation/Finalize_rrtmgp.cpp
                    ${SRC_DIR}/Radiation/Run_longwave_rrtmgp.cpp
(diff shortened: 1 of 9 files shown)


AMLattanzi:
Added plugins for the eccentricity and other factors missing from the code. The pseudo unit test still needs to be used to verify the SW and LW before merging. In a hacked branch, things were looking OK. 

Also, we will need to plug in a calendar day computation to the radiation module; likely coming from the time data structure I'd imagine.

AMLattanzi:
No moisture with a prescribed linear pressure profile and constant temperature.

SW:
image

LW (follow up why flux_dn in ERF, left, is half RRTMGP example):
image

PR #1621, opened 2024-05-14
Refactor cloud work in Kessler

AMLattanzi:
The saturation calculation in Kessler utilizes an analytical closure with a linearized approximation for `qsat`. This derivation is redone in the moisture overleaf in section 8.8:
image

However, the function used for this derivation does not match the one used to compute `qsat` . Alternatively, we may use the `erf_dtqsatw` function for consistency. The derivation with this function instead is given here:
Deriv_Qsat

The second method did not give the desired results for a designated unit test. Specifically, by setting `qv=1.01*qsat` we should induce condensation. If we re-compute `qsat` with the final `theta/T/P` we should see that `qv_{final} = qsat_{final}`. We do not pass this test for the current development branch or an implementation that uses the call to `dqsat`. I've thus added the Newton iterations to Kessler and this does pass the unit test (to the precision of the tolerance used for iterations).

This PR still needs testing with BOMEX and other cases to verify that it is correct.


AMLattanzi:
Using the development branch and modifying `Kessler.cpp` , gives the following output. Note that the final states do not match and the error is not negligible, this is the motivation for the current PR.
```
...
Real fac = qsat*4093.0/*L_v/(Cp_d/*std::pow(tabs_array(i,j,k)-36.0,2));

 // DEBUG
qv_array(i,j,k) = 1.1/*qsat;                
...
...
theta_array(i,j,k) = theta_array(i,j,k) + theta_array(i,j,k)/tabs_array(i,j,k)*d_fac_cond * (dq_vapor_to_clwater - dq_clwater_to_vapor - dq_rain_to_vapor);
tabs_array(i,j,k)  = getTgivenRandRTh(rho_array(i,j,k), rho_array(i,j,k)*theta_array(i,j,k), qv_array(i,j,k));
pres_array(i,j,k)  = getPgivenRTh(rho_array(i,j,k)*theta_array(i,j,k), qv_array(i,j,k)) * 0.01;

if (i==0 && j==0) {
     Real qsatf;
      erf_qsatw(tabs_array(i,j,k), pres_array(i,j,k), qsatf);
     Print() << "START: " << qsat << ' ' << qv << "\n";
     Print() << "END  : " << qsatf << ' ' << qv_array(i,j,k) << "\n";
     Print() << "\n";
 }
```
image

AMLattanzi:
The source of the mismatch has been identified. The `gamma` parameter in the following snippet is a function of time. We currently only use the value of `gamma` at the start of the step. By updating temperature and then converting to theta, the mismatch is removed (temperature update only depends upon `L_v/Cp`)!

image

AMLattanzi:
The use of `dtqsat` is still more consistent but will hold off with this update until the BOMEX simulation is ironed out.

PR #1623, opened 2024-05-15
SAM Cloud & Precip

AMLattanzi:
Various fixes to thermal sourcing for the SAM model; there were sign errors with the directionality of the change in q variables. This also needs testing.


AMLattanzi:
@nataraj2 the SAM results now look much more similar to the Kessler results after this modification. I believe the cloud and evaporation thermal sources had a sign error (feel free to review the code there).

 These changes significantly modify the Bomex results (~4-7k steps):
image

These are the Kessler results at (10k -18k steps):
image

AMLattanzi:
Recent results show that the poor agreement between Kessler and PINNACLES may be due to the `mu_turb` closure. The PR to bring in the Brunt-Vaisala scaling is in place. If that is verified to be the culprit, we can merge changes into this PR, rerun, and confirm if things are corrected for BOMEX.

asalmgren:
The fixes from this PR have been incorporated into the PR that added the SAM_NoRainNoIce option.

PR #1633, opened 2024-05-29
SAM only rain option

nataraj2:
This PR adds an option `SAM_OnlyRain` which implements the `SAM` microphysics model with only cloud water and rain ie. no ice, snow and graupel. 

An input file - `inputs_BF02_moist_bubble_SAM_OnlyRain` is added to `RegTests/Bubble` for the moist bubble test case with `SAM_OnlyRain`.

The gif shows the comparison of the rain for `Kessler` and `SAM_OnlyRain`. The contour min/max, though not exactly same, is comparable for both the models. The difference is because the rain parametrization is different for `Kessler` and `SAM` models.

MoistBubble_Kessler_SAM_Rain-ezgif com-video-to-gif-converter


nataraj2:
Redundant

PR #1641, opened 2024-06-11
Fix Deardorff overwrite of theta_v component.

AMLattanzi:
The Deardorff turbulence model was specifying the `theta_v` component as a function of the turbulence length scale. However, the `Pr_t` scaling after that specification was applied to horizontal and vertical components. Therefore, we were overwriting the `theta_v` component when using the Deardorff model. This PR corrects that and adds a bomex simulation input deck for Kessler.


AMLattanzi:
To demonstrate the error, one must have a positive stratification `\partial_{z} \theta`. The `ABL_input_sounding` case can be used to specify the stratification and print statements at the set of `theta_v` versus the end of the `ComputeTurbulentViscosity` show the deviation (with development).

PR #1654, opened 2024-06-20
Template land and sea MOST iters

AMLattanzi:
This avoids the need to do two loops over the MOST iters for each step. Instead, we define a base struct that has the iteration signature (used by all other derived structs). We then find the struct type we want and pass that through the templated wrapper. This was only compile tested.


AMLattanzi:
@WeiqunZhang can you provide some insight on the SYCL failure and whether this PR can be modified to circumnavigate said issue?

WeiqunZhang:
Virtual functions are not supported on GPUs. This is not just a SYCL issue. It's a compile time issue for SYCL, but a run time issue for CUDA. The latter is arguably worse. Even for CPU, the current approach of using virtual functions inside loops over cells is bad for performance. You have to go back to the old approach.

PR #1678, opened 2024-07-09
Modify polynomial coeffs for qsat

AMLattanzi:
Fix polynomial coeffs to match [Flatau 1992](https://rams.atmos.colostate.edu/cotton/vita/71.pdf).


asalmgren:
@AMLattanzi -- we have both 273.15 and 273.16 in the routine ... I feel like we should pick one or the other of those

PR #1695, opened 2024-07-17
Turbulent inflow generation logic patch

dustinma324:
This patch modifies the logic in two functions, `calc_tpi_update()` and `apply_tpi()`. In this version, a `MultiFab` is created within `DataStructs/TurbPertStruct.H` to store the random perturbation amplitude per cell within the perturbation box. The randomization step is now pulled out of `apply_tpi()`, which was incorrectly being called at every iteration, and placed within `calc_tpi_update()`. 

With this change, procedures such as calculating mean box velocity, perturbation amplitude, and assigning random perturbations are only triggered when the update interval is surpassed by the local elapsed time per box. This ensures that the randomization of perturbation assignment within the perturbation box occurs only once per update interval. The `Array4` that stores the random perturbation is then used in `apply_tpi()` per iteration as a direct addition to the source term of the desired field.

- This patch does not alter the behavior of acoustic waves when applying perturbations to the source term of the `rhoTheta` field.
- Incompressible box perturbation now working with movie provided

https://github.com/user-attachments/assets/c2f6f26b-a89d-4d9d-a575-c0e57691adaa


dustinma324:
Not sure why "https://iopscience.iop.org/article/10.1088/1748-9326/aa5d86" created a problem for me during this PR. Should I comment out this link?

PR #1707, opened 2024-07-23
Monotonic Advection

AMLattanzi:
This PR extends the hack that allowed ERF's Bomex simulations to match Siebesma (2003). The current formulation is applied to all scalars and needs verifying that it preserves the properties of the hack (which was only applied to `qc`). Furthermore, the choice to forward predict with `cur_cons` instead of `old_cons` is open to discussion/interpretation.

Before merging, a run out to 7200 [s] on the BOMEX case will be completed and the results will be posted here. The cloud cover and `qc` fields will be considered).


AMLattanzi:
Local BOMEX run yielded the following results (simulation terminated at ~9k [s]). Note, blue line is last instantaneous plane average, red is 1 hr time average of plane averages, and colored lines are each instantaneous plane average.
image
image

PR #1711, opened 2024-07-25
Mono Interp

AMLattanzi:
Rather than order reduction for boundness breaking, we could consider limiting the interpolated primitive variable to the `min/max` (this has to include ghost cells).


asalmgren:
If we believe that (rho s)_t + div (rho u s) = 0 is equivalent to s_t + u dot grad s = 0, then I think we believe that the value on the face should in fact lie between the values at cells on the left and right, so wouldn't those be the appropriate values to limit with?  This would differ from 2nd order in that we wouldn't necessarily take the average of those two, we'd just limit it to be between them.

PR #1720, opened 2024-08-02
Add option to input a constant geostrophic wind profile, update docs

ewquon:
This is a simpler alternative to creating a new problem and modifying `update_geostrophic_profile()`, which is a function of time and height. This approach takes a simple text file table, specified with `erf.abl_geo_wind_profile` that is constant with time.


ewquon:
For BOMEX, we can replace 
```
erf.add_custom_geostrophic_profile = true
```
with
```
erf.abl_driver_type = "GeostrophicWind"
erf.abl_geo_wind_table = "geostrophic_profile"
```
where the "geostrophic_profile" input file contains
```
0    -10.0  0
4000  -2.8  0
```

This eliminates the need to define the custom function in prob.cpp that is called every timestep.

PR #1736, opened 2024-08-13
Most Rotation Intermediate Clean Up

AMLattanzi:
Before completing the MOST stress rotations, this PR passes all the stress tensor components into the `impost_most_bcs` routine and unifies the definition of the diffusive flux. More specifically, we now include the negative sign: `tau = -k d \phi d x_{i}`. Prior to this PR, the negative was only done for the momentum stress tensor components while the scalars did not have the negative sign. This change impacts the MOST BCs routines, the stored fluxes `hfx & qfx`, and the diffusion source term. Finally, note that the output SGS fluxes `hfx3 & qfx3` will be `-k d \phi dz`.

### DIFF
--- a/Source/BoundaryConditions/MOSTStress.H
+++ b/Source/BoundaryConditions/MOSTStress.H
@@ -850,7 +850,7 @@ struct moeng_flux
                     const amrex::Array4<const amrex::Real>& /*t_surf_arr*/,
                     const amrex::Array4<amrex::Real>& dest_arr) const
     {
-        amrex::Real rho, eta, qv;
+        amrex::Real rho, eta;
 
         int ic, jc;
         ic = i  < lbound(cons_arr).x ? lbound(cons_arr).x : i;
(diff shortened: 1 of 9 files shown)


Source/BoundaryConditions/MOSTStress.H
```diff
@@ -957,7 +959,7 @@ struct moeng_flux
             // Note: Kh = eta/rho
             //      hfx = -Kh dT/dz  ==> +ve hfx corresponds to heating from the surface
             // Extrapolate from klo to ghost cell a distance of -deltaz; negative signs cancel
-            dest_arr(i,j,k,icomp+n) = rho*(theta + moflux*rho/eta*deltaz);
+            dest_arr(i,j,k,icomp+n) = dest_arr(i,j,zlo,icomp+n) + moflux*rho/eta*deltaz;
```
ewquon, line 962: I like that this gets rid of the multiple factors of `rho`, which always makes me do a double take.

ewquon (approved):
This actually fixes an inconsistency when writing out SFS profiles -- before this, `exp_most` would e.g., write out <u'w'> whereas the default path would write out `rho<u'w'>`. I've looked this over as carefully as I can but, ultimately, there's a lot of sign flips and in CI we trust.

PR #1738, opened 2024-08-14
MYNN Update

ewquon:
Major changes:

* Updated notation to match NN09
* Calculate Level-2 QKE from Richardson number and Level-2 stability functions -- this gives larger values of `qe`, which **leads to more aggressive limiting and successfully enforced SM>=0 and SH>=0** for a single-column test problem starting with a real sounding
* Add `Tv` to MOSTAverage -- this is equal to `T` if there's no moisture
* Update MYNN to be moisture aware: use qstar from ABLMost to calculate buoyancy flux (i.e., virtual temperature flux); calculate gradient of thetav if we have qv -- @baperry2, see updated `ComputeVerticalDerivativesPBL()`
* Create new MYNN data struct -- we could move the rest of the MYNN vars in here if we want @baperry2
* Placeholders added for the MYNN partial condensation scheme, which would affect the QKE buoyancy source
* Placeholders added in MOSTStress to include qstar when calculating the Obukhov length

**TBD**: Right now, the implementation matches NN09 if there is no rain. It's unclear to me if the partial condensation scheme (NN09, Appendix B) is needed; also, whether our microphysics schemes will play nicely with our MYNN implementation. To generalize, I suppose NN09 Eqns.20–22 could be used to calculate the buoyancy flux; then, the modeled kinematic heat flux could be back calculated (e.g., Stull 4.4.5d) and an equivalent `EddyDiff::Theta_v` (where "_v" is vertical, _not_ virtual) could be calculated.


ewquon:
Re-ran GABLS -- looks reasonable to me, will update golds.
GABLS1_hvelmag_rerun20240814
GABLS1_temp_rerun20240814
GABLS1_flux_rerun20240814
GABLS1_var_rerun20240814

Source/BoundaryConditions/MOSTStress.H
```diff
@@ -316,6 +316,7 @@ struct surface_flux
         u_star_arr(i,j,k) = mdata.kappa * umm_arr(i,j,k) / std::log(mdata.zref / z0_arr(i,j,k));
         do {
             ustar = u_star_arr(i,j,k);
+            // TODO: calculate virtual temperature flux if we have q*
```
AMLattanzi, line 319: Could we use the `Tv` mean you now compute in MOSTAverage here? I.e., if a valid moisture model exists, when we call the iterator we just pass the `Tv` plane average instead of the `T` plane average. This would be clean and efficient; just not sure if it's perfectly consistent with the theory?

AMLattanzi (approved):

PR #1749, opened 2024-08-20
Bugfix: populate terrain arrays before initializing microphysics object

debog:
Terrain arrays must be populated before the microphysics object is initialized because they are used by the `micro->Init()` function.

### DIFF
--- a/Source/ERF_make_new_level.cpp
+++ b/Source/ERF_make_new_level.cpp
@@ -192,27 +192,6 @@ void ERF::MakeNewLevelFromScratch (int lev, Real time, const BoxArray& ba,
         zflux_imask[lev] = nullptr;
     }
 
-    //********************************************************************************************
-    // Microphysics
-    // *******************************************************************************************
-    int q_size  = micro->Get_Qmoist_Size(lev);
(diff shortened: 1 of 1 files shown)


debog:
@AMLattanzi @asalmgren Is this fix correct?

asalmgren:
Hey @debog -- I'm actually in the middle of doing some additional fixes -- by definition if it works it's correct :-)   I'll merge this then incorporate it in the changes I'm doing now, which needed fixing when we are doing regridding with terrain ...

debog:
> Hey @debog -- I'm actually in the middle of doing some additional fixes -- by definition if it works it's correct :-) I'll merge this then incorporate it in the changes I'm doing now, which needed fixing when we are doing regridding with terrain ...

Sure! It works for one test that we are trying (it fixed initializing our particle-based microphysics model with terrain). I hope it did not break something else that depended on that particular order of initialization.

asalmgren:
@debog - would you mind if I actually close this PR and submit my PR with this changes and a number of others?

debog:
This fix will be addressed by @asalmgren's PR.

PR #1756, opened 2024-08-23
Improvements to generally help with idealized hurricane simulations

ewquon:
This PR includes:
* PBL height calculation — following WRF MYNN-EDMF — based on a hybrid of theta increase (CBL) and TKE (SBL) methods; selected with `erf.most.pblh_calc`
* **Beljaars free convection correction** based on wstar
* Mahrt & Sun subgrid velocity scale correction to account for underresolved mesoscale motions (only relevant for dx > 5 km)
* Updated thetav calc to account for liquid water in addition to water vapor
* Fix for Donelan flux type (39d772e9f4d9af5af76de96c05e7a2af5a7c9d12) in the default MOST path that sets ghost cells using eddy diffusivity
* Add **option to simulate idealized simulations over water** (with `erf.is_land=0`)
* Add **Donelan roughness option**
* Add SM,SH clipping for MYNN and surface windspeed clipping following WRF
* **Enable QKE advection to be turned off** (`advect_QKE` is true by default)
* **Fix restarts for MOST with variable surface roughness**
* Restore TropicalCyclone problem
* Updated docs

FYI @baperry2, the PBLH is updated just before calling `m_most->update_fluxes()` and stored in the `ABLMost` object. There are placeholders for the YSU PBLH calc. 

The Beljaars correction helps to stabilize the solution — verified that it tends to increase |L|. Also, verified that the solution without terrain is exactly equal to the solution with terrain (zsurf==0, grid_stretching_ratio=1.0).


ewquon:
Open to suggestions on which class `w_star` and `pblh` should live in; where PBLHeight.H should live; and how to amrex-ify `MYNNPBLH.compute_pblh()`...

baperry2:
How about creating a new `PBL` directory in `Source` to contain all the PBL stuff? Or maybe it should be a subdirectory in `Diffusion`?

asalmgren:
I'd be open to a new subdirectory

ewquon:
@baperry2 and @asalmgren I opted to create a Source/PBL directory. With the interfaces to MOST and possibly other parts of the code, I think it was general enough to have it's own high-level dir. 

FYI @baperry2, I renamed `ComputeTurbulentViscosityPBL` to `ComputeDiffusivityYSU` and stripped out the MYNN part but didn't touch anything in the YSU section -- hopefully, this should avoid massive merge conflicts.

baperry2:
Sounds good, I was planning to split out the turbulent viscosity routines as well! 

Should MYNNStruct live in DataStructs or PBL?

ewquon:
> Sounds good, I was planning to split out the turbulent viscosity routines as well!
> 
> Should MYNNStruct live in DataStructs or PBL?

I think it makes sense for DataStructs to have structs of all types.

baperry2:
Sure. Having it in with the rest of the PBL stuff would be more intuitive to me, but leaving it in DataStructs is more consistent with other parts of the code, so that's good too.

ewquon:
Well after thinking about it more, I don't disagree with you... As another example, the MOSTStress structs all live inside BoundaryConditions. And I think it makes sense for it to be close to the PBL code since I do some calculations inside of MYNNStruct, it's not plain old data.

PR #1791, opened 2024-09-05
General Newton Raphson iteration for HSE

AMLattanzi:
DO NOT MERGE BEFORE REVIEW.

This modification follows the derivation just completed in the JAMES paper (below). It attempts to unify how we compute HSE and get rid of the iterations without tolerance pathways. This method often converges in `2-3` iterations and is below `1e-12` tolerance.

image


Source/HSE_utils.H
```diff
@@ -13,6 +13,59 @@ namespace HSEutils
...
+                        const Real& T,
+                        const Real& qt,
+                        const Real& qv,
+                        Real& P,
+                        Real& rd,
+                        Real& F)
+    {
+        int iter=0;
+        Real eps  = 1.0e-6;
+        Real ieps = 1.0e6;
```
ewquon, line 49: Did you want this to be 1e-10 as in the paper?

ewquon (approved):
I've verified the derivation and the corresponding code, looks good to me.

Source/HSE_utils.H
```diff
@@ -13,6 +13,59 @@ namespace HSEutils
...
+
+            // Diagnose density and residual
+            rd = getRhogivenThetaPress(T, P, RdOCp, qv);
+            AMREX_ALWAYS_ASSERT(rd > 0.0);
+            Real r_tot = rd * (1. + qt);
+            F = P + 0.5*r_tot*g*dz + C;
+            ++iter;
+        }
+        while (std::abs(F)>m_tol && iter<25);
+        if (iter>25) printf("WARNING: HSE Newton iterations did not converge to tolerance!\n");
```
ewquon: I don't think `if (iter>25)` will ever be true.

Source/HSE_utils.H
AMLattanzi: No, the paper has some typos that I am still cleaning up.

Source/HSE_utils.H
AMLattanzi: Good call that should be a `>=` instead.

PR #1797, opened 2024-09-10
MYNN refactor to follow manuscript

AMLattanzi:
This should break the CTEST and regtests with MYNN. This code requires detailed review before merging.

The Deardorff model transports the LES TKE `k_{L]` while the PBL model transports ` 2 X` the PBL TKE. We first want to unify this so both models simply transport TKE; this removes the factor `2` for PBL source terms and collapses those terms to the same form as Deardorff. 

Additionally, there was a factor of `1/2` on the `mu_{t}` computation that I introduced due to the differences between `mu` and `K=2mu`. Inspecting [MY82](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/RG020i004p00851) we start with the following excerpt that explains the `boundary layer approximation`:
image

We may now examine the governing equations and closures:
image

If one simplifies the strain rate tensor to only involve vertical gradients, then we have off-diagonal components like `S_{31} = S_{13} = 1/2 \partial_z u`. The stress tensor in the u momentum equation then simplifies to `\tau_{31} = -2 mu_{t} S_{31} = mu_{t} \partial_z u`. Matching this with Eq. (22a) and Eq. (30a) above shows that `K_{M} = mu_{t}`. So the factor of `1/2` in the MYNN module is not correct. 

Extending these concepts to the source terms shows that the PBL TKE should have the following sources:
image

And this result is consistent with the Deardorff model once one simplifies the shear production by accounting for null lateral derivatives; where it is noted the Frobenius product `<S,S>_{F}^2` is the sum of the squares of the elements of strain rate tensor. 
image


baperry2:
How much does this impact the results for the GABLS1 case?

AMLattanzi:
> How much does this impact the results for the GABLS1 case?

@baperry2 I do not know yet. I wanted to get eyes on it first to see if I am correct. If we can all agree, then I think this warrants a re-run of GABLS and verifying that it does not corrupt the agreement. If we agree this is correct and GABLS looks horrible, we will have to have a whiteboard session to get to the bottom of things.

ewquon:
Abandoning "QKE" is a departure from literature but I'm in support of this because I do think it's an unnecessary and potentially confusing artifact. Can we do away with `RhoQKE_comp` throughout as well?

I think I agree that the factor of two only makes sense when you're working with the stress tensor and shouldn't be there if we're explicitly multiplying du/dz. Good catch!

AMLattanzi:
@ewquon The removal of `RhoQKE_comp` is a very interesting idea and what I was working towards. If you run with Deardorff and MYNN, do you want separate TKEs for each model since they have different dissipation rates (due to length scales and coefficients)?

baperry2:
I did the original implementation with the separate QKE variable so the difference in use and interpretation between TKE (for Deardorff) and QKE (for MYNN) would be clear and consistent with WRF, but I also see benefits to merging them into a single variable. 

I agree with Eliot that if we do this we should completely get rid of `RhoQKE_comp` and just using the TKE variable for MYNN, applying the same options for its advection/diffusion as we currently do for QKE. 

When I implemented this I did not intend that people should ever use the combination of Deardorff and MYNN due to the different competing TKE variables (my understanding from Branko is that people don't do that with WRF, and the MYNN + Smagorinsky for horizontal combination is only sometimes used). Right now, we error out if users attempt this combination, so there should be no issue with just completely disallowing it.

I've been going back and forth on removing the factor of 0.5 from the K computation, but right now I'm leaning toward that being correct.

AMLattanzi:
@baperry2 Thank you for that info, it helps to know the back story. 

If  we aren't allowing the combination of Deardorff + MYNN then I think we should remove the `RhoQKE_comp` .  If the Smag + MYNN is only sometimes used, are most WRF users left with zero lateral diffusion and just a vertical diffusivity?

Do you currently think the `1/2` factor should in fact be in MYNN module? Can you provide some mathematics that contradict my findings in the MY82 paper?

baperry2:
Sorry if it wasn't clear. I meant that I currently agree with your findings in this PR. In this work, it appears that they are using `K = mu_t` rather than `K = 2*mu_t` that appears elsewhere.

baperry2:
Ok, so I think this does need a bit more analysis. With the code changes in this PR, the GABLS1 test fails with negative temperature after ~125 iterations. 

This PR is doing two things (1) Redefining `QKE` to be `TKE = 2*QKE`, which otherwise should not impact the solution if propagated through everywhere, (2) making the correction of removing the 0.5 factor from K_m,v. To test the effect of only (2), I created a separate branch: https://github.com/baperry2/ERF/tree/mynn-KvsMu

I can complete a GABLS1 run with that branch. It does have some impact on the solution (interestingly moving more toward our YSU results, but somewhat away from the LES results). Based on this, I think the K change is probably correct, although we may need to assess why there is now slight disagreement for both YSU and MYNN. However, it seems like we've missed something that needs to be modified for things to stay consistent when going from solving for QKE to solving for TKE.

fluxes
temperature
windspeed

ewquon:
> @ewquon The removal of `RhoQKE_comp` is a very interesting idea and what I was working towards. If you run with Deardorff and MYNN, do you want separate TKEs for each model since they have different dissipation rates (due to length scales and coefficients)?

That's a valid point. In most cases, I think you would either use the PBL scheme --or-- the Deardorff turbulence model at a particular grid level. The outlier use case is if you wanted to add horizontal diffusion in addition to vertical diffusion on a mesoscale domain. While I've seen this done with PBL + Smag, I'm not sure why you would want to put in more effort to transport 3D k when you're underresolving the strain rate. Maybe better to leave in both for generality...

ewquon:
I should've read ahead -- Bruce & Branko corroborated my thinking.

AMLattanzi:
> Ok, so I think this does need a bit more analysis. With the code changes in this PR, the GABLS1 test fails with negative temperature after ~125 iterations.
> 
> This PR is doing two things (1) Redefining `QKE` to be `TKE = 2*QKE`, which otherwise should not impact the solution if propagated through everywhere, (2) making the correction of removing the 0.5 factor from K_m,v. To test the effect of only (2), I created a separate branch: https://github.com/baperry2/ERF/tree/mynn-KvsMu
> 
> I can complete a GABLS1 run with that branch. It does have some impact on the solution (interestingly moving more toward our YSU results, but somewhat away from the LES results). Based on this, I think the K change is probably correct, although we may need to assess why there is now slight disagreement for both YSU and MYNN. However, it seems like we've missed something that needs to be modified for things to stay consistent when going from solving for QKE to solving for TKE.
> 
> fluxes temperature windspeed

@baperry2 I observed the same thing today. RE (1): I think we would have the wrong IC now since the values need to be halved everywhere in the domain since `TKE= 1/2 QKE` . But it does seem something else is at play. I'll focus on this tomorrow and thanks for running the GABLS case!

baperry2:
Agreed on the IC and that there still must be something else missing for the QKE->TKE conversion. The QKE and TKE ICs should be made consistent and a benefit of moving to just TKE is that would automatically be the case.

AMLattanzi:
This PR now only encompasses a change where `QKE -> TKE`. The factor of `2` on the `mu` computation was removed. To complete this work for code optimization, we would delete the `RhoQKE_comp` and then for dissipation we would make `PBL_length -> Turb_length` and put `C_e/L` for Deardorff and `B1/ell` for MYNN. We would also add a `SmnSmn_1d` that ignores lateral gradients. The Deardorff and MYNN pathway would then converge in a clean manner.

AMLattanzi:
As expected, no change to GABLS results:
image
image
image

baperry2 (approved):

baperry2:
@AMLattanzi can this be updated and merged?

Also do you intend to merge the QKE and TKE variables soon after thus, or push that off until later?

AMLattanzi:
@baperry2 I'll close this PR and then refactor to remove QKE when I get the time. In a bit of a hold up with the JAMES paper and the Q4 report.

PR #1802, opened 2024-09-12
Fix IC for GABLS 

AMLattanzi:
This PR is part of the broader effort to collapse `QKE` and `KE` vars and sort out small issues with the `PBL` schemes.

For the GABLS case, there was a bug with the initial condition. It was found that the IC issue causes instability in the PR that goes from transporting `QKE -> KE`. 

Re running the GABLS case with the correct IC in this PR gives the following results (use last two `plt32400` and `plt32100` as inputs to the validation repo):
image
image
image

For reference, using `profile.nc` input (generated with the IC in development):
image
image
image


Exec/ABL/inputs_GABLS1_ysu
```diff
@@ -66,15 +66,11 @@ erf.abl_driver_type = "GeostrophicWind"
...
-#erf.theta_ref       = 263.5 # from Cuxart et al. 2006
-
+erf.les_type    = "None"
 erf.pbl_type    = "YSU"
 
 # Initial conditions from Beare et al. 2006
 prob.KE_0            = 0.4 # [m2/s2]
 prob.KE_decay_height = 250. # [m]
-prob.KE_decay_order  = 1
+prob.KE_decay_order  = 3
```
baperry2: Actually can you just delete the prob.KE_0 inputs for the YSU input file? These go unused because it doesn't use either QKE or TKE.

baperry2 (approved):
Looks good to me. Surprising that this impacts stability for the change when the profiles look almost identical.

Exec/ABL/inputs_GABLS1_ysu
AMLattanzi: Will do

PR #1805, opened 2024-09-13
MYNN mu_t fix

AMLattanzi:
This PR addresses the factor of `1/2` issue raised in [PR 1797](https://github.com/erf-model/ERF/pull/1797).

The salient change here is that we need to use one-sided differences with MOST. This makes the derivatives consistent with what would be obtained with the explicit MOST pathway and also alleviates an error. Namely, `K_turb` must be calculated before MOST populates the ghost cells, this ensures the non-explicit pathway works. Therefore, when inside the MYNN module, the ghost cells at `k==-1` are not valid (they have been filled with the default BC --- e.g. NoSlip, FOEXTRAP, etc.). The one-sided diffs match how explicit MOST populates the vertical ghost cells. The effect of this change is to drastically modify the GABLS results by DECREASING the stresses. When we correct the factor of `1/2` on `mu_t` we bring those curves back in line with benchmark results; this also IMPROVES the heat flux predictions from the current development branch:

image
image
image


Source/PBL/ERF_PBLModels.H
```diff
@@ -94,7 +94,8 @@ ComputeVerticalDerivativesPBL (int i, int j, int k,
                                amrex::Real& dudz,
                                amrex::Real& dvdz,
                                const int RhoQv_comp,
-                               const int RhoQr_comp)
+                               const int RhoQr_comp,
+                               const bool use_most=false)
```
baperry2: default to true or simply remove and always treat as being true (both PBL schemes require MOST and we error out if the user tries to use either without MOST). If we remove, we could actually simplify this function a good bit applying the requirement that the bottom boundary is always MOST.

baperry2 (approved):
Nice catch and the fix looks good to me. Just the note that we could remove all the `use_most` tracking through PBL stuff because both models require MOST anyway. But also it doesn't hurt in case it becomes necessary at some point in the future.

I tried testing whether this also improves YSU and unfortunately it does not (vertical derivatives are not directly used at the bottom surface), but it may be something similar regarding the order BCs are applied.

PR #1818, opened 2024-09-18
Submodules: dependabot updates and shallow clones

baperry2:
This PR does two things. There are plusses and minuses to both so I'm not going to merge just yet:

- Set up dependabot updates for the submodules. Dependabot will automatically create a PR once per month to update each submodule if more recent versions are available. These PRs must still be approved and can be discarded if undesired, and submodules can still be manually updated whenever desired. See amr-wind example: https://github.com/Exawind/amr-wind/pull/1241 (for this to work, a rule exception would have to be added to allow dependabot to create branches in the github repo). This ensures the code stays up to date, but does create extra noise in terms of pull reuqests.

- Recommend shallow clones for submodules. Somewhat reduces the download size of ERF, but not a huge impact because the tip of all active branches still gets cloned for each submodule. Users can unshallow if they desire the full git history, but that does create an extra step.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,3 @@ updates:
     directory: /
     allow:
       - dependency-name: "Submodules/AMReX"
-      - dependency-name: "Submodules/RRTMGP"
-      - dependency-name: "Submodules/WW3"
-      - dependency-name: "Submodules/YAKL"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+
+updates:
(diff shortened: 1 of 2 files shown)


AMLattanzi (commented):
I like this idea. However, I can't remember if the YAKL and RRTMGP hashes are pinned for a certain reason (I think they are). For that reason, I'd like to roll this out in stages. This would let us sort out the kinks and let me get back into radiation to check things.

That said, for now, I think we should limit dependabot to only update AMReX each month.

baperry2:
Ok, modified so dependabot only touches AMReX

AMLattanzi (approved):

PR #1841, opened 2024-10-02
Real BDY Re-factor

AMLattanzi:
This code refactors the BDY set and nudging based upon a simplified test case with constant `theta=300` and `u=10` in the BDY region but `theta=300` and `u=v=w=0` in the IC region. The flow should propagate into the domain with the advective velocity. The changes made herein allow that to occur by NOT setting `rho` but allowing it to evolve by the continuity equation. Other set/nudge vars are augmented with the `primitive` vars from metgrid or WRF. The test case result is much better now, but more vetting is needed:
image (7)

### DIFF
--- a/Exec/DevTests/MetGrid/inputs
+++ b/Exec/DevTests/MetGrid/inputs
@@ -7,7 +7,7 @@ amrex.fpe_trap_overflow = 1
 
 # PROBLEM SIZE & GEOMETRY
 geometry.prob_extent =  28000 16000 8000
-amr.n_cell           =   140 80 100
+amr.n_cell           =   140     80  100
 
 geometry.is_periodic = 0 0 0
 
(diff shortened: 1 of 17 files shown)


AMLattanzi:
The Flowmas MetGrid case with `NullMoist` at `1000 steps`. The harsh banding in velocity at x_lo and x_hi faces is not readily apparent:
image (8)

AMLattanzi:
Dave W's test case with `substepping` , `dt=0.5`,  `4 procs`,  `set_width = 1`, and `relax_width = 5`. Due to the larger time step and the relaxation zones, the propagation of the bdy velocity into the domain should be MUCH faster.

Note, the system does not rapidly converge to the homogeneous solution of `u=10`. Rather, pulsing occurs with a characteristic frequency. While not tested, more dissipation and extended run time may alleviate this; also note that this case is abruptly driven by the BDY where real cases will match the IC at t=0.

10 steps
image

50 steps
image

100 steps
image

200 steps
image

AMLattanzi:
Using just set zones with `set_width = 5` does readily converge towards the homogeneous solution (200 steps):
image

asalmgren (approved):

PR #1898, opened 2024-10-18
add get_dt accessor function for multiblock

mukul1992:


asalmgren:
@mukul1992 -- while you're here, how about making those functions take "int lev" as an argument and return the value at that level -- it will be useful down the line ...

mukul1992:
Should I still keep them under the ifdef or make them available in general?

asalmgren:
No strong feeling about that for now -- do whatever is convenient

PR #1905, opened 2024-10-23
Remove qke 

AMLattanzi:
This PR follows the equations in the JAMES paper and compresses the `KE` and `QKE` vars. This reduces the memory footprint in terms of the dycore variables and the eddy diffusivities. 

Source terms:
Deardorff:
image
MYNN 2.5:
image

This is a work in progress and results will be posted to this thread.


asalmgren (approved):

Source/SourceTerms/ERF_make_sources.cpp
```diff
@@ -60,8 +60,8 @@ void make_sources (int level,
     const bool use_terrain      = solverChoice.use_terrain;
 
     TurbChoice tc = solverChoice.turbChoice[level];
-    const bool l_use_deardorff  = (tc.les_type == LESType::Deardorff);
-    const bool l_use_QKE        = tc.use_QKE && tc.diffuse_QKE_3D;
```
ewquon, line 64: Looks like we lost the option toggle the 6th-order numerical diffusion term for KE/QKE alone. I think this was an experimental option and, if I recall correctly, didn't make much difference as far as numerical stability of MYNN was concerned.

AMLattanzi: Let me put that back in just for consistency. I'd like to change as few things as possible with this PR.

ewquon (approved):

AMLattanzi:
There are some small differences between this new implementation and the current development:

1. The use of `ComputeSmnSmn` averages the strains to CC, rather than using a central difference at the CC.
2. The source terms are computed once per time step rather than once per RK stage.
3. The buoyancy source follows the Deardorff model and thus is `\partial_{z} \theta_d` whereas the MYNN module used `\theta_v`. It's unclear if Deardorff should also use `\theta_v`.

Despite these differences (which means the CTest will never pass) the GABLS case looks decent:
image
image
image

ewquon:
> There are some small differences between this new implementation and the current development:
> 
> 1. The use of `ComputeSmnSmn` averages the strains to CC, rather than using a central difference at the CC.
> 2. The source terms are computed once per time step rather than once per RK stage.
> 3. The buoyancy source follows the Deardorff model and thus is `\partial_{z} \theta_d` whereas the MYNN module used `\theta_v`. It's unclear if Deardorff should also use `\theta_v`.
> 

The buoyancy source should depend on the kinematic virtual temperature flux, so we need to use virtual temperature. I believe we have the unsaturated path coded up but it gets more complicated under saturated conditions. From Deardorff1980: 

image

AMLattanzi:
I have reverted the source terms to conform to what was originally in the development branch. We can sort out the `theta_v` and frequency with which we compute source terms in a subsequent PR so things don't get too convoluted.

The benchmark remake is due to the fact that `KE` is clipped at machine precision but `QKE` was clipped at `1.0e-12`.

PR #1937, opened 2024-11-07
Add Source/IO/ERF_SampleData.H to CMakeLists.txt

asalmgren:


mukul1992:
Hi @asalmgren unfortunately it has to be ERF_SampleData.H instead of just SampleData.H

PR #2007, opened 2024-12-06
Base number of substeps on a theoretical max.

AMLattanzi:
This attempts to remove the hardcoded `6` substeps with a computed value based upon a theoretical limit for the substep max time.

### DIFF
--- a/Source/TimeIntegration/ERF_ComputeTimestep.cpp
+++ b/Source/TimeIntegration/ERF_ComputeTimestep.cpp
@@ -201,10 +201,9 @@ ERF::estTimeStep (int level, long& dt_fast_ratio) const
          if (fixed_dt[level] > 0. && fixed_fast_dt[level] > 0.) {
              dt_fast_ratio = static_cast<long>( fixed_dt[level] / fixed_fast_dt[level] );
          } else if (fixed_dt[level] > 0.) {
-             // Max CFL_c = 2.0 and 6 substeps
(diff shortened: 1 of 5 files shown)


ewquon:
For reference, from WRF: https://github.com/wrf-model/WRF/blob/master/dyn_em/solve_em.F#L441-L458

The number of substeps is based on a speed of sound of 300 m/s and a CFL limit of 1, although the standard WRF rule of thumb (dt = 6 dx [km]) combined with enforcing a minimum number of substeps=4 will in practice make the time stepping more conservative.

Example 1:
* dt/dx = 0.006
* speed of sound ~ 300 m/s
* acoustic CFL = 300 * 0.006 = 1.8
* WRF uses the minimum substep ratio = 4
* effective CFL = 0.45

Example 2 (slightly larger time step):
* dt/dx = 0.007
* acoustic CFL = 2.1
* WRF rounds up to the next even ratio = 4
* effective CFL = 0.53

In all cases, the effective CFL is less than the theoretical limit (from the WRF manual, based on Wicker & Skamarock 2002) of 1.42 / sqrt(3) = 0.820 for a 5th order scheme.

ewquon (approved):

asalmgren:
Those are the same four tests that failed before when we changed the default dt behavior -- we could either re-make the benchmarks or fix the tests so they have specified values rather than using defaults?  Either way ok with me.

asalmgren (approved):

PR #2011, opened 2024-12-06
Setup EB Scalar DevTest.

skang67:
1. Setup a scalar advection problem in DevTest/EB_Test.
2. Added ERF_redistribute to Make.package and the ERF class. (This function is not yet added to ERF_TI_to_substep_fun.H.)


asalmgren:
Three things:
1) remove the line xlo.rhotheta = 1 -- that was a mistake on my part.  The correct entry is xlo.theta which is already there.
2) Instead of changing Make.package, please "git mv ERF_redistribute.cpp ERF_Redistribute.cpp"
3) I'm not sure I understand what you're doing in ERF_Redistribute.cpp -- it seems to me you could remove the function in lines 10-31 then just call the function currently at line 34 inside the MFIter loop in ERF_TI_no_substep.H.  Is that where you're going?

asalmgren (approved):

PR #2020, opened 2024-12-12
Saturation Adjustment Micro Model

AMLattanzi:
This PR introduces the simple Saturation Adjustment microphysics model that only involves `qv, qc` vars and uses Newton iterations to converge to a saturated state.

This was tested on the moist bubble rise and gave reasonable results:
image (1)


pressel (approved):
Thanks, @AMLattanzi.  This looks great to me.

PR #2021, opened 2024-12-12
Minor changes for deterministic simulations with particles

debog:
- Added an input parameter to choose "stable" `Redistribute()` for GPU sims (use [this](https://github.com/AMReX-Codes/amrex/pull/4200)).
- Moved the fixing of random seed to the beginning of the ERF constructor so that it is set before anything else (eg. particle initialization).

### DIFF
--- a/Source/ERF.cpp
+++ b/Source/ERF.cpp
@@ -10,6 +10,7 @@
 #include <ERF_EOS.H>
 #include <ERF.H>
 #include <AMReX_buildInfo.H>
+#include <AMReX_Random.H>
 #include <ERF_Utils.H>
 #include <ERF_TerrainMetrics.H>
 #include <memory>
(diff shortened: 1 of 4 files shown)


Source/ERF.cpp
```diff
@@ -84,6 +85,15 @@ Vector<std::string> BCNames = {"xlo", "ylo", "zlo", "xhi", "yhi", "zhi"};
...
 {
+    int fix_random_seed = 0;
+    ParmParse pp("erf"); pp.query("fix_random_seed", fix_random_seed);
+    // Note that the value of 1024UL is not significant -- the point here is just to set the
+    // same seed for all MPI processes for the purpose of regression testing
+    if (fix_random_seed) {
+        Print() << "Fixing the random seed" << std::endl;
+        InitRandom(1024UL);
+    }
+
```
debog, line 96: @asalmgren Is this okay? I moved this piece of code from `Initialization/ERF_InitCustom.cpp` to the beginning of the ERF constructor.

asalmgren: Yes -- happy to have it there.    I'm pushing through some PRs today to do live regression so I'll merge this and some others very soon.

PR #2031, opened 2024-12-18
RANS Part 1: Poisson wall distance

ewquon:
Wall distance is needed to estimate RANS length scales. Currently implemented is the trivial case of using the Poisson equation ([Tucker 2003, JCP](https://doi.org/10.1016/S0021-9991(03)00272-9)) over flat terrain. 

To activate this code path, set `erf.rans_type = "kEqn"`. At the moment, the code will simply fill a walldist MultiFab (which can be included as an output in the plot_vars list) with the distance from the zlo boundary. To activate the Poisson solver, the appropriate comment blocks in `ERF::poisson_wall_dist()` should be removed.

Here, the Poisson approach recovers the exact wall-normal distance with an error of 0.2% the grid spacing at cell centers.
image

Also implemented is blanking for thin immersed bodies. While results appear qualitatively correct, the Poisson wall distances are inaccurate away from the body. I've verified that the error is reduced as the size of the immersed body increases relative to the domain extent.
walldist_thinbody_2d

---
Input files (need to comment out the `#if 0` blocks to test):
[inputs_flat_abl_walldist.txt](https://github.com/user-attachments/files/18186162/inputs_flat_abl_walldist.txt)
[inputs_thinaf_walldist.txt](https://github.com/user-attachments/files/18186161/inputs_thinaf_walldist.txt)


ewquon:
FYI, @hgopalan. I think I've got a handle on this poisson solve and I think it should work for general terrain. Just need to plug in the appropriate terrain metrics, following the pressure solver code in the anelastic path.

PR #2062, opened 2025-01-13
RRTMGP Port

AMLattanzi:
This adds radiation via `RRTMGP` and `YAKL`. The `RRTMGP` submodule is set to coincide with what is currently used by the **EAMXX** component of **E3SM**; see [here](https://github.com/E3SM-Project/rte-rrtmgp/tree/b24ca1f616e45659b334dbd7297017cb7927367e). The files added to ERF are modeled after and ported from those used in **EAMXX**; see [here](https://github.com/E3SM-Project/E3SM/tree/master/components/eamxx/src/physics/rrtmgp).

TODO action items:
====================
- [x] Remove unnecessary code 
- [x] O3 volume mixing ratio is function of height
- [x] Add doc info for inputs and outputs
- [ ] Hook up to LSM (`mu0, albedo, zenith`)
- [x] Hook up to orbital calcs (commented out for now)
- [x] Sort out `lat/lon` availability (only with real init?)
- [ ] Allow for multi-level calcs (only level 0 calls update right now)
- [x] Convert rad src for application to dry pot temp
- [x] Replicate simple [unit test](https://github.com/E3SM-Project/E3SM/discussions/6878) and compare output


Source/Radiation/ERF_Radiation.cpp
AMLattanzi, line 1: @pressel can you take a look at the following:

https://github.com/erf-model/ERF/blob/e531a885f409854cb139e9d6abc5ef30525bf40a/Source/Radiation/ERF_Radiation.cpp#L425

https://github.com/erf-model/ERF/blob/e531a885f409854cb139e9d6abc5ef30525bf40a/Source/Radiation/ERF_Radiation.cpp#L449

https://github.com/erf-model/ERF/blob/e531a885f409854cb139e9d6abc5ef30525bf40a/Source/Radiation/ERF_Radiation.cpp#L455

https://github.com/erf-model/ERF/blob/e531a885f409854cb139e9d6abc5ef30525bf40a/Source/Radiation/ERF_Radiation.cpp#L595

AMLattanzi: `lwp` and `iwp` are filled in a specific spot. I did away with `nc_lay` since it was only used for output data we don't have or use. I think the h2o vmr is computed correctly, but a second set of eyes would be useful.

The `eff_radius` quantities are inputs that we must compute; any insight there would be helpful.

AMLattanzi:
Successful unit test. To replicate results, the `ifdef 0` in `ERF_Radiation.cpp` must be set to `1` and the files in `Exec/DevTest/Radiation` must be used.

image
image
image
image

PR #2069, opened 2025-01-15
Init WRFInput Memory

AMLattanzi:
TODO List
===================

- [x] Divide out the mapfactors  in velocity
- [x] Port the convert wrfbdy routines to use MFs
- [x] Address general compilation issues
- [x] Compare plt0000 for WPS_Test

### DIFF
--- a/Source/ERF.cpp
+++ b/Source/ERF.cpp
@@ -878,13 +878,13 @@ ERF::InitData_post ()
         IntVect ngvect_cons = vars_new[lev][Vars::cons].nGrowVect();
         IntVect ngvect_vels = vars_new[lev][Vars::xvel].nGrowVect();
 
-        int icomp_cons = 0;
         int ncomp_cons = lev_new[Vars::cons].nComp();
         bool do_fb     = true;
 
 #ifdef ERF_USE_NETCDF
         // We call this here because it is an ERF routine
         if (use_real_bcs && (lev==0)) {
+            int icomp_cons = 0;
             bool cons_only = false;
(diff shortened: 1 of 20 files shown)


Source/ERF.cpp
```diff
@@ -872,24 +872,41 @@ ERF::InitData_post ()
...
         // ***************************************************************************
         // Physical bc's at domain boundary
         // ***************************************************************************
         IntVect ngvect_cons = vars_new[lev][Vars::cons].nGrowVect();
         IntVect ngvect_vels = vars_new[lev][Vars::xvel].nGrowVect();
 
-        (*physbcs_cons[lev])(lev_new[Vars::cons],0,ncomp,ngvect_cons,t_new[lev],BCVars::cons_bc,true);
-        (   *physbcs_u[lev])(lev_new[Vars::xvel],0,1    ,ngvect_vels,t_new[lev],BCVars::xvel_bc,true);
-        (   *physbcs_v[lev])(lev_new[Vars::yvel],0,1    ,ngvect_vels,t_new[lev],BCVars::yvel_bc,true);
+        int icomp_cons = 0;
```
AMLattanzi: @asalmgren I added the call to the realbcs here since I was seeing incorrect values in the ghost cells. I don't think it changes anything but let me know your thoughts.

asalmgren: Looks ok to me.   Do we know yet if we match plt00001 from before the changes?

asalmgren (approved):

AMLattanzi:
This has an issue with `WPS_Restart` that needs sorted out. Erroneous values for `RhoTheta`.

PR #2072, opened 2025-01-17
Adding control of adaptive time-stepping.

pressel:
This PR is meant to address an issue in which adaptive time-stepping computes an erroneously large time-step `dt` when using the anelasitc solver under low (no) windspeed conditions. The fix is to limit the length of the initial time-step and subsequent time-steps, which are controllable though the input file variables  `erf.dt_max_initial` and erf.dt_max` respectively.


asalmgren:
@pressel -- it looks like you accidentally deleted change_max from the table in Inputs.rst?  Or was it duplicated?

pressel:
> @pressel -- it looks like you accidentally deleted change_max from the table in Inputs.rst? Or was it duplicated?

Oops, I've fixed it.

PR #2095, opened 2025-02-03
fix fine ghost cells of eddyDiffusivity that overlie coarse grids

asalmgren:


Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp
```diff
@@ -858,4 +858,84 @@ void ComputeTurbulentViscosity (const MultiFab& xvel , const MultiFab& yvel ,
...
+            if (geom.isPeriodic(i)) {
+                domain.grow(i,1);
+            }
+        }
+
+        for (MFIter mfi(eddyViscosity); mfi.isValid(); ++mfi)
+        {
+            Box bxcc   = mfi.tilebox();
+
+            Box planex = bxcc; planex.setSmall(0, 1); planex.setBig(0, ng[0]); planex.grow(1,ng[1]); planex.grow(2,ng[2]);
```
AMLattanzi: Defining the boxes this way seems like it could lead to a lot of extra work (particularly on CPU with tiling) and maybe an out of bounds reference. If we take a `tilebox` of (10,0,0) (20,10,10), which is inside the valid box (0,0,0) (40,40,40),  then directly setting its xlo and xhi would force us to check  for a notcovered cell even though we really don't want to because it doesn't coincided with a proc boundary --- i.e., this would be picked up by tbx (0,0,0)(10,10,10). Could we instead follow what is done in the BC routines and operate on a `growntilebox` and then set the hi/lo side (see [here](https://github.com/erf-model/ERF/blob/development/Source/BoundaryConditions/ERF_BoundaryConditionsCons.cpp))? I think this optimization would create null boxes when no work is to be done and we wouldn't cycle every `i,j,k` then.

AMLattanzi (approved):

PR #2143, opened 2025-02-24
Update AMReX submodule

gardner48:
Update AMReX submodule to pick up rpath fix in AMReX-Codes/amrex#4343


gardner48:
@skang67 this will fix your build issues with FFTW on Mac

PR #2161, opened 2025-02-28
Make initial fields for `init_type=wrfinput` more consistent with EOS

ewquon:
**This fixes the EOS check at https://github.com/erf-model/ERF/blob/04633f7c7d3fd7d1fe75bca90f964a7afd407a22/Source/Initialization/ERF_InitFromWRFInput.cpp#L788-L793**

### Background
The diagnostic equation for total pressure (due to dry air and water vapor) is given by the ARW v4 technical note Eqn. 2.16:
image

This is derived from the EOS with total pressure equal to the sum of partial pressures:
image
and the definition of potential temperature assuming the moist gas constant is approximately equal to the gas constant for dry air (i.e., R ≈ Rd).

Compare with https://github.com/wrf-model/WRF/blob/master/dyn_em/module_big_step_utilities_em.F#L1060-L1076 (in general, `use_theta_m=1`), which gives:
```
p + pb = p0 * (r_d*(t0+thm))/(p0*(al+alb))**(Cp/Cv)
```
Therefore, for the EOS to be consistent within WRF, 
image.

There is a bit of confusion that remains. In WRF, the code calls `AL` and `ALB` the inverse density (for the perturbational and base-state, respectively) but does not explicitly say it is the _dry_ inverse density. Based on the implementation, I have to conclude that it is dry.

**tl;dr**: There was nothing wrong with the initial implementation, per se, it's just unclear to me how the perturbation potential temperature "T" in the wrfinput—which should be the correct quantity, if taken at face value—is actually used in WRF. So I have rewritten the data ingestion to be in terms of moist theta "THM" (confusingly called `t` in the WRF source code), which is used throughout the WRF dycore..


AMLattanzi:
Running the `WPS_Test` case with development (left) and this branch (right) show the error between `P_eos` from our equation of state and `P_tot` obtained from wrfinput is reduced near the surface (where qv will be greatest). The print was at `i==100 && j==100` and are in favor of this approach.
image

PR #2166, opened 2025-03-02
Add datetime options for time control

ewquon:
E.g.,
```
start_datetime = "2021-07-20 00:00:00"
stop_datetime  = "2021-07-21 00:00:00"
```
instead of
```
start_time  = 1626739200.0 
stop_time   = 1626825600.0
```

Specifying `start_datetime` or both (`start_datetime` and `stop_datetime`) will also trigger a timestamp print at the beginning of each coarse step.

### DIFF
--- a/Source/ERF.cpp
+++ b/Source/ERF.cpp
@@ -395,8 +395,9 @@ ERF::Evolve ()
     for (int step = istep[0]; step < max_step && cur_time < stop_time; ++step)
     {
         if (use_datetime) {
-            Print() << "\n" << getTimestamp(cur_time, datetime_format)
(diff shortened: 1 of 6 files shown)


Source/ERF.cpp
```diff
@@ -393,6 +394,11 @@ ERF::Evolve ()
     //      for finer levels (with or without subcycling)
     for (int step = istep[0]; step < max_step && cur_time < stop_time; ++step)
     {
+        if (use_datetime) {
+            Print() << "\n" << getTimestamp(cur_time, datetime_format)
```
AMLattanzi: To pass the HIP failure I think we need to convert the Real values (cur_time, start_time) to type time_t when passing through these functions.

PR #2171, opened 2025-03-06
Dirichlet inflow file with theta

AMLattanzi:
This PR extends the inflow file option to include a `theta_d` input that is optional. The code was tested with the `TurbulentInflow` pathway and the test can be replicated by running the `generateLogLawProfile.py` in two modes. 

1. Running the python code as-is will generate a Dirichlet inflow data that includes `z, u, v, w`
2. Modifying the python code to generate a Dirichlet inflow data that includes `z, u, v, w, theta_d`

The files can be read at each level but no time interpolation exists yet.


ewquon (approved):

hgopalan:
Can we consider adding TKE in future if we want to run RANS using the 1-Equation model ?

PR #2184, opened 2025-03-12
Fix for 'jQuery is not defined' when using search

ewquon:
(popup in browser developer console)

https://github.com/readthedocs/sphinx_rtd_theme/issues/1452


ewquon:
Hallelujah
image

ewquon:
@mukul1992

PR #2216, opened 2025-03-25
MOST --> SurfaceLayer

AMLattanzi:
**NOTE: THIS PR BREAKS BACKWARDS COMPATIBILITY FOR MOST BCs!**

This PR overhauls the `ERF_ABLMost.H/cpp` implementation and renames it `ERF_SurfaceLayer.H/cpp`. 

The following paradigm is adopted:

1. `SurfaceLayer` class only provides subgrid diffusive fluxes, it does NOT touch ghost cells
2. With `surface_layer` BC type, the lateral velocities and scalars are set to `HOEXTRAP` so as to preserve their gradient as we go into ghost cells below `klo`. The normal velocity is set to `0` with a `Dirichlet` condition. This allows us to NOT need special stencils with MOST since we will for example recover the same `d_z u` as a one-sided diff and ensure `d_x w = 0` at the bottom surface.
3. There is no longer an explicit most option, we always write into the diffusive flux structures

Things to be done:
- [x] Verify with ABL MOST simulation
- [x] Re-run GABLS case
- [x] Re-run Bomex case
- [x] Fix ctests where needed
- [x] Update docs
- [x] Update inputs in Exec
- [x] Test on Hurricane Henri case
- [x] Clean up naming conventions
- [x] Test cmake compilation


AMLattanzi:
The ABL_MYNN25 case confirms that the new SGSDiff method captures the appropriate diffusive flux while the development branch does not. The development branch yields an incorrect diffusive flux due to the averaging of `mu_turb(k)` and `mu_turb(k-1)` when computing the stress (MOST doesn't do this averaging). 

At early times, the diffusive flux is under-predicted by 2X.

AMLattanzi:
@ewquon here are the GABLS results with SGSDiff:
image
image
image

AMLattanzi:
For simplicity, the SatAdj moisture model was utilized with the SGSDiff PR for Bomex. The mean qc profile looks very reasonable and inline with the variations we have seen. This did not impact the `500 m` cloud height that seems to be a tad off.
image

ewquon:
I have verified that this change does not appreciably change simulated ABLs when running LES. Checks:
* GABLS1 SBL: dx=dy=dz=3.125 m, Deardorff (Moeng 1984 coeffs)
* Moeng & Sullivan 1994 CBL: dx=dy≈50 m, dz≈20 m, Q=0.24 Km/s, Deardorff (also Moeng 1984 coeffs)

Updates to the validation repo are forthcoming.

PR #2222, opened 2025-03-28
LES updates, including `erf.les_type=Smagorinsky2D` and `erf.mix_isotropic`

ewquon:
UPDATE: stripped out MOST mods

* ~~Unless we're using the original MOST pathway, calculate strain and turbulent viscosity _after_ applying BCs~~
* ~~Save MOST diagnosed stresses to ghost cells~~
* Add 2-D Smagorinsky option that uses an alternate SmnSmn formulation following WRF 
* Add `erf.mix_isotropic` option (default: true); if false, then separate horizontal and vertical diffusivities are calculated (for both momentum and heat), based on separate length scales
* Add common flags to TurbChoice struct; remove duplicated code throughout

~~@AMLattanzi the MOST changes are in b9e3cdaf5822947642caa37dafd86231b9883e6c, 9191d7ff4b4dbccb809e6707ff62d15e934f3c06, and 1704a1dd23a0b35e3ac08cecada36bd8f4518e40~~

~~Should not break the github CI or nightly tests 🤞 — seems like everything uses the original MOST pathway.~~


Source/Diffusion/ERF_EddyViscosity.H
```diff
@@ -54,29 +54,43 @@ ComputeSmnSmn (int& i, int& j, int& k,
...
-        s23bar = 0.5  * ( tau23(i  , j  , k+1) + tau23(i  , j+1, k+1) );
+        s23bar = 0.5  * ( s23(i  , j  , k+1) + s23(i  , j+1, k+1) );
     }
     else
     {
-        s23bar = 0.25 * ( tau23(i  , j  , k  ) + tau23(i  , j  , k+1)
-                        + tau23(i  , j+1, k  ) + tau23(i  , j+1, k+1) );
+        s23bar = 0.25 * ( s23(i  , j  , k  ) + s23(i  , j  , k+1)
+                        + s23(i  , j+1, k  ) + s23(i  , j+1, k+1) );
+        if ((i==124) && (j==191) && (k<=1)) {
```
AMLattanzi: Forgot an AllPrint

Source/BoundaryConditions/ERF_ABLMost.cpp
```diff
@@ -456,8 +456,8 @@ ABLMost::compute_most_bcs (const int& lev,
                                              t23_arr, t32_arr);
                     } else {
                         if ((k == klo-1) && vbxx.contains(i,j,k) && exp_most) {
-                            t13_arr(i,j,klo) = stressx;
-                            if (t31_arr) t31_arr(i,j,klo) = stressx;
+                            t13_arr(i,j,k) = stressx;
```
AMLattanzi: These lines are confusing because the current MOST loops over a box that for the zlo face is from `k \in [-ng[2] -1]`. So we are storing the flux in a ghost cell of t13/t23? Doesn't this cause an issue when we compute the diffusive src since it should be stored at `k==0`?

ewquon: I copy from k=-1 later on https://github.com/erf-model/ERF/blob/9191d7ff4b4dbccb809e6707ff62d15e934f3c06/Source/TimeIntegration/ERF_MakeTauTerms.cpp#L260-L273

PR #2226, opened 2025-04-01
Fix scalar advection when anelastic to be time-centered

asalmgren:


Source/TimeIntegration/ERF_SlowRhsPost.cpp
```diff
@@ -100,6 +100,8 @@ void erf_slow_rhs_post (int level, int finest_level,
 {
     BL_PROFILE_REGION("erf_slow_rhs_post()");
 
+    amrex::Print() << "IN POST " << std::endl;
```
AMLattanzi: Legacy print needs removing.

AMLattanzi (approved):

PR #2227, opened 2025-04-01
Correctly load u and v from wrfbdy

ewquon:
From the WRF v4 tech doc:
image

The `U_B[XY][SE]` and `V_[XY][SE]` fields in wrfbdy* follow this definition -- this was verified (using [this script](https://github.com/erf-model/erftools/blob/9c8dbaf44fd71cbc25371b66376259e7a92035be/bin/verify_wrfinput_wrfbdy_consistency.py)) by back-calculating `U` and `V` and comparing with the corresponding wrfinput file. Therefore, when loading in a wrfbdy file, in addition to rescaling by the column mass it is also necessary to remove the map factor to avoid double-counting.


ewquon:
After much deliberation, it appears that the x- and y-velocities ("U" and "V") in the wrfinput have mapfactor baked in and so the original IO code, which divides the input x- and y- velocities by `m`, correctly gives `u` and `v` as meters (on earth) per second for real data cases.

PR #2237, opened 2025-04-11
[Scaling] Adding GPU weak scaling plots to docs

nataraj2:
This PR adds the GPU weak scaling comparison on Perlmutter, Frontier and Aurora.


asalmgren:
@nataraj2 -- I like adding the scaling for all three machines, but let's not remove the scaling plot that shows the benefit of GPU-aware MPI.   I think it's significant that we can take advantage of that.

nataraj2:
Have updated with the GPU aware MPI plot.

PR #2284, opened 2025-05-05
Draft for generalization of map factor terms

asalmgren:


AMLattanzi:
Derivations for comparison with the implemented code:
image
image
image

PR #2365, opened 2025-06-07
Further generalize specified massflux option

ewquon:
Allow specified mass flux within a layer of fluid (given by two heights `const_massflux_layer_lo` and `const_massflux_layer_hi`). May be used drive canonical LES (e.g., to match hub-height or rotor-averaged wind speeds) -- FYI @mchurchf.


ewquon:
Ugh, rebase nightmare

PR #2441, opened 2025-07-23
More wrfinput fixes

ewquon:
* Don't cast sea surface temperature (SST) or skin temperature (TSK) to int
* Convert TSK and SST to potential temperatures (otherwise, it's implied that the surface pressure is always 1e5)  -- these are used with the surface layer code to compare with the potential temperature at a specified zref to determine surface fluxes. @akashdhruv and @AMLattanzi, any idea how this might affect coupling with the LSM?
* Add option (`erf.real_extrap_w=false`) to set the vertical velocity in bdydata to 0 instead of foextrap from the interior. Another option (future todo) would be to read w from the wrfbdy.


akashdhruv:
These changes won't break the LSM code as it currently stands. The logic to use LSM-computed values for setting boundary conditions hasn't been implemented yet. At this stage, we are driving the LSM with atmospheric forcing data and copying its output to the relevant MultiFabs, which are later overwritten by the BCs. I’ll run a test to confirm that software compatibility remains intact.

AMLattanzi:
Agreed that the changes don't have any obvious issues with the LSM implementation. Without the LSM, the surface and sea temperatures are just now "correct" and I again don't see an issue.

PR #2442, opened 2025-07-23
RRTMGP + NOAH-MP

AMLattanzi:
## RRTMGP + NOAH-MP
This PR couples the inputs to and from RRTMGP and NOAH-MP.

NOAH-MP -> RRTMGP vars
- t_sfc (surface temp)
- sfc_emis (surface emissivity)
- sfc_alb (dir/dif_vis/nir)

RRTMGP -> NOAH-MP vars
- sw_flux_dn (short wavelength flux down)
- lw_flux_dn (long wavelength flux down)

## NOAH-MP Changes
1. Defined data members are now 2D and used for the coupling (@akashdhruv)

## RRTMGP Changes
1. Surface emissivity from LSM is passed to RRTMGP instead of `Kokkos::deep_copy(0.98)` like is done in eamxx.


cmkaul:
It's very exciting to see this progress! A quick comment that it would be helpful to maintain the distinction between Noah and Noah-MP land surface models to avoid confusion.

akashdhruv:
@cmkaul Would you prefer that we also change the name from `LandSurfaceModel/NOAH` to `LandSurfaceModel/NoahMP`

cmkaul:
Yes, I think that is a good idea. Thanks!

akashdhruv:
> Yes, I think that is a good idea. Thanks!

Sounds good!

PR #2453, opened 2025-07-28
Nudging Modifications

AMLattanzi:
## Sponge Approach
1. Change the nudging to follow the same approach as sponging. A `start` and `end` point is defined and the sponge intensity depends upon where one is in that region (quadratic decay).
2. For now, do not fill ghost cells and do not sponge `theta` with BDY file data.

## Approaches
1. Change to sponge (everything else the same)
2. Do not sponge `theta/qv` but do fill the ghost cells.
3. Do not sponge or fill ghost cells with `theta/qv` and use `FOEXTRAP` at boundary
4. No Laplacian term and sponge the full width (not width-1)


AMLattanzi:
# Tornado results
- `1000 steps` and `xy plane @ 5000 m`

## Development
Development_1000_steps

## Sponge
Sponge_1000_steps

## No Sponging `Theta/Qv`
Sponge_1000_steps_NoRelax_T_Qv

## FOEXTRAP and No Sponging `Theta/qv`
Sponge_1000_steps_No_T_Qv

asalmgren (approved):

PR #2490, opened 2025-08-16
change the stencil for lateral vels at bottom surface to not reach be…

asalmgren:
…low the sfc for the metric-weighted terms


Source/Diffusion/ERF_ComputeStrain_T.cpp
```diff
@@ -134,9 +134,15 @@ ComputeStrain_T (Box bxcc, Box tbxxy, Box tbxxz, Box tbxyz, Box domain,
             Real dz0  = ( z_nd(i,j,k+1) - z_nd(i,j,k-1) );
             Real idz0 = 1.0 / dz0;
 
-            Real GradUz = 0.5 * idz0 * ( u(i  ,j  ,k+1) + u(i  ,j-1,k+1)
+            Real GradUz = (k == 0) ?
+                                idz0 * ( u(i  ,j  ,k+1) + u(i  ,j-1,k+1)
```
AMLattanzi: There appears to be a few errors here, with one being legacy. With the current code, the `idz0` was built with the `k+1` and `k-1` indices, so it should not be used with the one-sided difference at `k==0`.  The second, and legacy, issue is that `z_nd` is being used, which is at the nodes, so there is a mismatch when using it with `u/v` which are at cell faces since no averaging was done in the vertical. My first inclination is that we need to do this:

``` bash
// NOTE: Not averaging z_nd to u/v face is an approximation
Real dz0 = (k==0) ?  0.5 * ( z_nd(i,j,k+2) + z_nd(i,j,k  ) ) :
                     0.5 * ( z_nd(i,j,k+2) + z_nd(i,j,k+1) - z_nd(i,j,k  ) - z_nd(i,j,k-1) );
Real idz0 = 1.0 / dz0;

Real GradUz = (k == 0) ? idz0 * ( u(i  ,j  ,k+1) + u(i  ,j-1,k+1)
                                - u(i  ,j  ,k  ) - u(i  ,j-1,k  ) ) :
                         idz0 * ( u(i  ,j  ,k+1) + u(i  ,j-1,k+1)
                                - u(i  ,j  ,k-1) - u(i  ,j-1,k-1) );
```

AMLattanzi (changes requested):
This to consider that could improve my comment:

1. We can either pass `z_cc` in to reduce the amount of data we touch when computing `idz0`
2. We can do formal averaging to the exact locations but would need a `idzu` and `idzv` since they are at different lateral positions

PR #2497, opened 2025-08-20
fix dz calc

asalmgren:


Source/Diffusion/ERF_ComputeStrain_T.cpp
```diff
@@ -131,18 +131,21 @@ ComputeStrain_T (Box bxcc, Box tbxxy, Box tbxxz, Box tbxyz, Box domain,
         bool need_to_test = (bc_ptr[BCVars::yvel_bc].lo(0) == ERFBCType::ext_dir_upwind) ? true : false;
 
         ParallelFor(planexy,[=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept {
-            Real dz0  = ( z_nd(i,j,k+1) - z_nd(i,j,k-1) );
+            Real zhi =            0.5d0 * (z_nd(i,j,k+1) + z_nd(i,j,k+2);
+            Real zlo = (k == 0) ? 0.5d0 * (z_nd(i,j,k  ) + z_nd(i,j,k+1) :
+                                  0.5d0 * (z_nd(i,j,k-1) + z_nd(i,j,k  );
+            Real dz0 = (k == 0) ? (zhi - zlo) : 0.5 * (zhi - zlo);
```
AMLattanzi: In the case that `(k!=0)`, I believe the 0.5 is not needed since it is already accounted for in the construction of the `zhi` and `zlo`. The only optimization is that we touch the `k+1` entry when `k==0` but it will cancel when constructing the total delta z.

AMLattanzi (approved):

AMLattanzi (approved):

PR #2503, opened 2025-08-21
Radiation Pre-Step

AMLattanzi:
Make radiation pre-step and setval the noah mp coupling variables to something reasonable.


akashdhruv (approved):
Looks good. I will run some tests on Monday.

PR #2528, opened 2025-09-08
NOAHMP + SurfaceLayer

AMLattanzi:
# Overview
Noah-MP now stores the `t_flux, q_flux, tau13, & tau23` fluxes for coupling. Those fluxes are used by the SurfaceLayer class when available.

## Notes
The `tau` stresses need to be exposed on the NoahMP side @akashdhruv, otherwise the simulation will be `FreeSlip` at the bottom wall.


akashdhruv:
Testing this now.

PR #2573, opened 2025-09-30
Fix `init_from_wrfinput` bug when using `rebalance_wrfinput`

ewquon:
This little indexing bug rears its ugly head in debug mode and mysteriously manifests as an index error in the interpolation operator during the advection source update.


Source/Initialization/ERF_InitFromWRFInput.cpp
```diff
@@ -636,9 +636,9 @@ ERF::init_from_wrfinput (int lev,
                 // First integrate from sea level to the height at klo
                 {
                     // Vertical grid spacing
-                    z_lo = 0.0;
-                    z_hi = 0.125 * (z_arr(i,j,klo  ) + z_arr(i+1,j,klo  ) + z_arr(1,j+1,klo  ) + z_arr(i+1,j+1,klo  )
-                                   +z_arr(i,j,klo+1) + z_arr(i+1,j,klo+1) + z_arr(1,j+1,klo+1) + z_arr(i+1,j+1,klo+1));
+                    z_lo = 0.25  * (z_arr(i,j,klo  ) + z_arr(i+1,j,klo  ) + z_arr(i,j+1,klo  ) + z_arr(i+1,j+1,klo  )); // surface elevation
```
AMLattanzi: The z_lo is set to `0.0` because we use `p_0` as the known reference pressure at the bottom face. If you average the z_phys you will end up with the true height, which could be very far from sea level and thus the pressure at that point is much less than p_0.

PR #2586, opened 2025-10-08
Update handling of `init_type=NCFile`

ewquon:
* Read `QV`
* Reading `R_HSE`, `T_HSE`, and `P_HSE` are optional; calc base state if not specified


ewquon:
OK, I think I've got all the wrinkles ironed out.

PR #2618, opened 2025-10-17
This version gives same answer with tiling or without but we should b…

asalmgren:
…e able to uncomment the currently commented parts


ewquon (approved):
Verified that results are identical for WPS test with wrfinput and wrfbdy. Tested 4 cases on 32 MPI ranks: with `OMP_NUM_THREADS` = 1 and 2, with and without tiling.

PR #2672, opened 2025-11-01
EB: DEBUG -- Apply FillBoundary for S_new at slow_rhs_fun_post.

skang67:
This PR fixes a bug that caused different RHS values at grid boundaries. For `terrain_type==EB`, the advection fluxes are computed on a one-cell-wider box to allow flux interpolation. As a result, we need `FillBoundary` at the end of each RK stage. (There may be a more efficient approach to avoid the extra `FillBoundary`, but I apply this to patch the grid boundary for now.) I confirmed the consistency between `np=1` and `np=4` with `fcompare`.


skang67:
- I fixed a spell in the new commit.
- The `S_new[i].FillBoundary` loop iterates from `i=0` to `i <= AMREX_SPACEDIM`, so it is applied to all variables from `cons` through `zmom`.

PR #2736, opened 2025-11-18
Add clipping from HP96

ewquon:
* Add clipping of Pr, which matches both Hong & Pan 1996 as well as Hong et al 2006. 
* Use K bounds from HP96 -- the increased lower bound appears to have a stabilizing effect


ewquon:
@hgopalan

PR #2768, opened 2025-12-01
Fix potential input_sounding gotcha

ewquon:
Currently `SurfaceLayer.t_surf` and `SurfaceLayer.q_surf` are initialized to default values if `erf.most.surf_temp` is not specified. This is inconsistent with initial conditions from a provided "input_sounding". This _shouldn't_ be an issue because by default, the code follows the specified heat flux path and the surface temperature is not updated or used by the surface layer under idealized conditions. However, other physics models (e.g., the PBL scheme) may request `m_SurfaceLayer->get_t_surf(lev)`, which would return an incorrect default value.

To be safe, if `erf.init_type = input_sounding`, then the initial surface layer t_surf and q_surf will now be updated from the sounding. A separate PR will address the PBL scheme dependencies on t_surf.

### DIFF
--- a/Source/BoundaryConditions/ERF_SurfaceLayer.H
+++ b/Source/BoundaryConditions/ERF_SurfaceLayer.H
@@ -557,8 +557,10 @@ public:
   }
 
   amrex::MultiFab* get_t_surf (const int& lev) { return t_surf[lev].get(); }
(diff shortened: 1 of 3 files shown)


asalmgren:
@ewquon -- I wonder if it would be worth creating set_t_surf and set_q_surf access functions and use those to set the value rather than accessing the MF and setVal ... it might make it more transparent down the line?

PR #2779, opened 2025-12-03
Add MRF pblh interp; fix MRF k-indexing consistency; fix has_diff check

ewquon:
Improvements:
* Follow Hong et al 2006 more closely and interpolate the PBL height after correcting for thermal excess -- effect should be more noticeable on coarser grids
* Use `t_layer` instead of `t_surf` for consistency with Hong et al 2006, Eqn 2, θ_va
* Evaluate `zval` and `Rib` at the same k index in the PBL height corrector step


hgopalan (approved):
Thanks for spotting the issue.

PR #2806, opened 2025-12-17
Updates to the MetGrid initialization and forcing pathway

wiersema1:
These updates resolve some memory bottlenecks and facilitate multilevel configurations. The MetGrid pathway now processes each time sequentially to reduce the memory footprint. Temporary variables were previously fabs, which introduced a memory bottleneck that scaled in severity with the number of ranks, but have been updated to MultiFabs to eliminate this bottleneck.

Note: this is a second PR with a clean port of the changes from the original PR, which is to be closed and not merged.


ewquon:
@wiersema1, my colleague has confirmed that this PR fixes his memory issue using the metgrid path -- so no need for his workaround, which was
```
amrex.the_arena_is_managed = true
amrex.the_arena_init_size = 10737418240  # 10 GB
```

asalmgren:
This PR has been replaced by https://github.com/erf-model/ERF/pull/2944 which contains the changes from this PR as well as additional changes to bring the code up to date.

PR #2887, opened 2026-02-14
Fix SYCL Compilation: Kokkos::View -> amrex::Table

WeiqunZhang:
`sycl::is_device_copyable<Kokkos::View>` is false. Thus `Kokkos::View` cannot be used directly in SYCL kernels.


akashdhruv:
This PR should be merged after https://github.com/E3SM-Project/rte-rrtmgp/pull/44 to bump Submodule tag for `Submodules/RRTMGP`

PR #2891, opened 2026-02-18
Fix stop_time and put lots of comments about which time is total vs e…

asalmgren:
…lapsed


Source/ERF_MakeNewLevel.cpp
```diff
@@ -156,11 +159,17 @@ void ERF::MakeNewLevelFromScratch (int lev, Real time, const BoxArray& ba_in,
         if ( (solverChoice.init_type == InitType::WRFInput) || (solverChoice.init_type == InitType::Metgrid) )
         {
             AMREX_ALWAYS_ASSERT(solverChoice.terrain_type == TerrainType::StaticFittedMesh);
+            //
+            // Note that "time" here is elapsed time, and start_time is the start_time from wrfinput/metgrid files
+            //
             init_only(lev, start_time+time);
```
AMLattanzi, line 165: These lines are inconsistent with the definition of `t_new` and `t_old` being elapsed time. The call should be changed to just `time` since this input is used to set `t_new` in `init_only`.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -580,29 +580,35 @@ SurfaceLayer::compute_SurfaceLayer_bcs (const int& lev,
...
+
+        amrex::Print() << "IN SST " << time  << " " << m_start_time << " " << m_stop_time << std::endl;
+        amrex::Print() << "ALPHA " << alpha << std::endl;
+
+        int n_time_hi = n_time_lo + 1;
+
+        //
+        // We choose to extrapolate from the final bdy slice if the run goes beyond final_bdy_time
+        //
+        if (m_start_time + elapsed_time_since_start_low >= m_stop_time) {
```
AMLattanzi, line 607: `m_stop_time` is not defined when we get to here.

Source/Utils/ERF_InteriorGhostCells.cpp
```diff
@@ -108,26 +111,35 @@ realbdy_compute_interior_ghost_rhs (const Real& bdy_time_interval,
...
+
     int n_time_p1 = n_time + 1;
-    if ((time == final_bdy_time_elapsed) && (alpha==0)) {
-        // stop time coincides with final bdy snapshot -- don't try to read in
-        // another snapshot
+
+    //
+    // We choose to extrapolate from the final bdy slice if the run goes beyond final_bdy_time
+    //
+    if (time >= final_bdy_time) {
```
AMLattanzi, line 139: I think we want to pin `n_time` as well. One could in theory get past the last bdy file time and then `n_time` is out of bounds. This change should also go in Moist set RHS.

if (time >= final_bdy_time) {
        // stop time coincides with final bdy snapshot -- don't try to read in
        // another snapshot
        n_time       = static_cast<int>( (final_bdy_time - start_bdy_time)/  dT);
        n_time_p1 = n_time;
        alpha         = 0.0;
        oma           = 1.0;
    }

Source/TimeIntegration/ERF_TimeStep.cpp
```diff
@@ -31,20 +31,21 @@ ERF::timeStep (int lev, Real time, int /*iteration*/)
...
     bool use_moist = (solverChoice.moisture_type != MoistureType::None);
-    if (solverChoice.use_real_bcs && (lev==0)) {
-        Real dT = bdy_time_interval;
+    if (solverChoice.use_real_bcs && (lev==0))
+    {
+        Real time_since_start_bdy = time + start_time - start_bdy_time;
 
-        int n_time_old = static_cast<int>( (time        ) /  dT);
-        int n_time_new = static_cast<int>( (time+dt[lev]) /  dT);
+        int n_time_old = static_cast<int>( (time_since_start_bdy        ) /  bdy_time_interval);
```
AMLattanzi, line 38: int ntimes = bdy_data_xlo.size();
Real time_since_start_bdy = time + start_time - start_bdy_time;
int n_time_old = std::min(static_cast<int>( (time_since_start_bdy        ) /  bdy_time_interval), ntimes-1);
int n_time_new = std::min(static_cast<int>( (time_since_start_bdy+dt[lev]) /  bdy_time_interval), ntimes-1);

PR #2896, opened 2026-02-19
Add Documentation and Error Message for Refinement Regions

hgopalan:
This PR provides more detailed error messages when the refinement box criterion are not properly set.


asalmgren:
@hgopalan -- thanks for doing this!   Several comments - 1) you'll need to fix white space.  You can do this in your local repo by running .github/workflows/style/check_trailing_whitespaces.sh   2) You wrote "So in that case the lo_indices should be 0 and the hi_indices should be one less than a number divisible by the refinement ratio in the vertical direction." but I think that should be only the lo/hi indices *in the vertical direction.    3) I'd say go ahead and delete the old error messages in ERF_Tagging.cpp -- yours are definitely clearer.   4) We say in the docs you have to refine all the way to the top for PBL -- should we enforce that in the tagging routine as well?  Thanks for doing this!

hgopalan:
> @hgopalan -- thanks for doing this! Several comments - 1) you'll need to fix white space. You can do this in your local repo by running .github/workflows/style/check_trailing_whitespaces.sh 2) You wrote "So in that case the lo_indices should be 0 and the hi_indices should be one less than a number divisible by the refinement ratio in the vertical direction." but I think that should be only the lo/hi indices *in the vertical direction. 3) I'd say go ahead and delete the old error messages in ERF_Tagging.cpp -- yours are definitely clearer. 4) We say in the docs you have to refine all the way to the top for PBL -- should we enforce that in the tagging routine as well? Thanks for doing this!

The PBL routine throws an error message if it is not refined. We can add it in the tagging as well if you think it is a better place.

asalmgren:
Let's add that error message to the tagging routine -- it would fit thematically :-)

asalmgren:
@hgopalan -- this is looking good -- can you add a comment reflecting the change Eliot just made (which I just merged) saying that max_level for each refinement defaults to the max_level set for the overall run -- then I think this is good to merge

PR #2912, opened 2026-02-24
Add Option for Specifying Index Space at Coarser Level Similar to WRF 

hgopalan:


asalmgren:
@hgopalan - 1) If we specify the coarse indices they don't need to be divisible by the refinement ratio (because they will be multiplied by the refinement ratio to create the fine indices) -- could you remove that check?   2) Could you replace "wrf" by "crse" in the code and the docs?   I would rather not make it seem WRF-specific.  Thanks!

asalmgren (approved):

PR #2921, opened 2026-02-26
Building but no terrain input file for immersed forcing, to use for reg test

adamwise95:


gardner48:
@adamwise95 if you want this case to run with `make test` i.e., as part of the GitHub and GitLab testing. This file will need to go in a subdirectory of `Tests/test_files` and you'll need to add a `add_test_r()` line for this case in `Tests/CTestList.cmake`.

PR #2927, opened 2026-02-28
Surface Layer W/ Vert Refinement

AMLattanzi:
# Summary
This PR modifies classes that operate on the surface only (e.g., Surface Layer and Land Surface Model) to use a modified 2D box array. Specifically, the box array is collapsed onto the valid box's `smallEnd(2)`. This change allows us to only operate **ONLY** on the surface by adding the following lines to an MFIter loop.
```
int klo = geom[lev].Domain().smallEnd(2);
if (my_bx.smallEnd(2) != klo) { continue; }
```

## Notes
The only 2D BA that does not follow this canonical form is that for the mapfactors, which are implicitly assumed to be defined at `k==0`. For this reason, the `ba2d_mf` is retained in MakeNewArrays and added to Checkpoint.

PR #2940, opened 2026-03-03
**2D Box Array Revamp**

AMLattanzi:
# Issue summary
Operations that only touch the bottom surface (`surface layer and LSM`) would segfault with vertical refinement. This was due to the fact that the 2D BA is constructed as
```
b.setRang(2,0);
```
but MFIter loops are often defined over 2D data structures. Therefore, one cannot distinguish whether a box at `(ihi, jhi, 0)` actually originated at the surface or if it was compressed to reside there. 

## Solution approach
The approach in this PR is to retain the 2D BA definition by compressing an index to `0` (this is true for 1D as well). However, the MFIter **MUST** be defined over a 3D MF and then a simple modification may be made:
```
int klo = domain.smallEnd(2);
if (bx.smallEnd(2) != klo) { continue; }
bx.makeSlab(2,klo);
```
The final set of rules then follow
1. Lower dimensional BAs compress their indices to `0`
2. MFIters over lower dimensional data must use 3D MFs with appropriate continue lines (see above)
3. 3D data should be indexed as `(i,j,k)` while 2D data should be indexed as `(i,j,0)` in all instances.

## Final note

- A correction was made to the pressure EOS call in NOAH, which again omitted moisture


AMLattanzi:
Confirmed that this PR runs with the ML case with vertical refinement outlined in [PR 2927](https://github.com/erf-model/ERF/pull/2927).

Last remaining item before merging:

- [ ] Verify all Biollante regtests pass

PR #3006, opened 2026-03-17
Implicit Diffusion Optimization & Correction

AMLattanzi:
# Summary
LU decomposition solution for implicit diffusion integrated into current development.

## Fixes
This PR will address [Issue 3041](https://github.com/erf-model/ERF/issues/3041).

## Testing
Use column test case and force a diffusion only problem with the following
- [x] _N with theta and surface layer
- [x] _S with theta and surface layer
- [x] _T with theta and surface layer

## Expansion
We should implicitly diffuse at least Q1, which impacts pressure, and TKE, which impacts turbulence models. **The implementation of additional implicit diffusion solves work will be part of a follow up PR.**


AMLattanzi:
## Efficiency
Testing with development vs PR for the `column_diffusion_test` case with a `40 X 40 X 51` grid on `4 ranks`

Development:
```
TinyProfiler total time across processes [min...avg...max]: 1.06 ... 1.063 ... 1.067
ImplicitDiffForState_N()                        60    0.02223    0.02247    0.02272   2.13%
```

Current PR
```
TinyProfiler total time across processes [min...avg...max]: 1.052 ... 1.053 ... 1.054
ImplicitDiffForState_N()                        60    0.01214    0.01222    0.01247   1.18%
```

## Accuracy
Outputs of the final `theta` state solution were confirmed to match up to 12 digits after 10 steps between development and PR 3006.

PR #3007, opened 2026-03-17
Fixes to  SDM regtest and example inputs

debog:
Minor fixes


gardner48:
Looks like the GitHub CI was reporting [failed builds as successful](https://github.com/erf-model/ERF/actions/runs/23176153353/job/67338629197?pr=3007#step:9:333) because the `grep` for warnings or errors was case-sensitive and missed `Error:` in the output.

PR #3019, opened 2026-03-17
RRTMGP Calday

AMLattanzi:
# Summary
The E3SM convention is as follows:
```
- calday 1 == Jan 1 0Z in eamxx_rrtmgp_process_interface.cpp:673
- jday   ! Julian cal day (1.xx to 365.xx) in shr_orb_mod.F90:104
```
 We were previously `0` based on `Jan 1` so we need to add 1 to calday for consistency.
 
 Additionally, RRTMGP is now called pre-step so we use `t_old` to get the total time sent to the radiation module.
 
 Finally, we revert to the averaging orbital calculation with the fixed calday. This should be smoother and more accurate for longer time lapses between radiation calls. 
 
 This PR also addresses the non-negative filtering of LSM data.


AMLattanzi:
@adamwise95 confirmed the PR gives similar results as development. If issues with a diurnal cycle reappear we should investigate the time averaging for the orbital calcs, which is reintroduced in this PR.

PR #3049, opened 2026-03-25
Split EB ParallelFor

gardner48:
Split ParallelFor into three parts. Fixes "parameter space overflowed" error in GitHub CUDA CI.


gardner48:
FYI @skang67

skang67:
Thank you for fixing this. I just learned that lambda capture has limits! It seems I passed too many parameters.

gardner48:
> It seems I passed too many parameters.

I'll note this works fine as is on matrix where we test with a newer version of CUDA that allows for a larger parameter space.

PR #3133, opened 2026-04-27
single-precision fixes

wiersema1:
Resolving problems running with ERF_PRECISION=SINGLE on GPU (HIP). These errors were silent when compiled with double precision.

**WENO eps underflow:**
The WENO structs use eps = Real(1.0e-40). In single precision 1e-40 is subnormal. Variables initialized to a uniform value have b1 = b2 = 0 which results in w = g/0² = Inf and Inf/Inf = NaN in every WENO flux. Changed eps to Real(1.0e-12) so (eps)² = 1e-24 >> FLT_MIN in float. Changes are within compiler directive for compiling in single precision. Behavior is unchanged when compiling in double precision.

**eps in compute_p_k:**
compute_p_k uses eps = Real(1.e-10). In single precision p + 1e-10 == p so F_prime = 0, delta_p_k = NaN, and since |NaN| is not > 1e-8 then compute_p_k returns NaN. Changed eps to be relative to p. max(|p| * 1e-5, 1e-5). I also found the need to relax the convergence check from 1e-8 to 1.0 Pa. Changes are within compiler directive for compiling in single precision. Behavior is unchanged when compiling in double precision.

**Velocity sentinel:**
In ERF_InitCustomPertState.cpp, lev_new[Vars::xvel/yvel] are pre-filled with sentinel 1.234e20 by ERF_MakeNewArrays. lev_new[Vars::zvel] is uninitialized on GPU. In single precision, adding a small perturbation to these values gives unintended behavior. Added compiler directive to use a valid sentinel when compiling in single precision. Behavior is unchanged when compiling in double precision.

**Real vs. double-literal comparison:**
In ERF_MakeSources.cpp, tflux != 1e-8 and Olen_in != 1e-8 compare Real against a double literal. In single precision these comparisons are always true. Changed to Real(1.0e-8).


AMLattanzi:
@wiersema1 we need to confirm some of those changes. Particularly the WENO `eps` values have a huge impact on the conservation properties of the advection operator. The values used are not arbitrary and using smaller values showed poor conservation properties for advecting scalars; these errors caused huge `q_c` accumulation in simulations like BOMEX.

I would recommend either testing with an advecting scalar or run BOMEX with WENO to confirm.

wiersema1:
@AMLattanzi, thanks Aaron! I agree we should hold off on these changes until we are confident in their impact. I'll run and compare some moist cases, including BOMEX, when I have the availability to do so.

asalmgren:
@wiersema1 -- Dave this all looks good except for "Real tolerance = Real(1.0);" in ERF_HSEUtils -- is that really what you meant?

asalmgren:
Most of the changes in this PR are now merged in https://github.com/erf-model/ERF/pull/3174

PR #3138, opened 2026-04-29
Revert Prob Name Get

AMLattanzi:
This is not valid for real sims without names.

### DIFF
--- a/Source/ERF.cpp
+++ b/Source/ERF.cpp
@@ -2349,7 +2349,7 @@ void
 ERF::ReadParameters ()
 {
     std::string prob_name = "Unknown";
-    ParmParse pp_pn("erf"); pp_pn.get("prob_name", prob_name);
+    ParmParse pp_pn("erf"); pp_pn.queryAdd("prob_name", prob_name);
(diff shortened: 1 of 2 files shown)


gardner48:
#3134 changed `query` to `get` because one input file had a typo in the prefix (`tert.prob_name` instead of `erf.prob_name`) and others files were missing the problem name both of which led to unexpected results. Throwing an error would have helped catch this sooner. It also looks like an unknown problem name could cause similar issues but catching that would require validating the input problem name. 

I think switching `.queryAdd("prob_name", prob_name);` in `ERF.cpp` to `.query("prob_name", prob_name);` could address the first two cases since `.get("prob_name",my_prob_name);` in `ERF_Prob.cpp` should fail, but that wouldn't address providing an unknown problem name (and might still an issue for real sims). Maybe real sims should also have a problem name?

AMLattanzi:
@gardner48 Thank you for the input. Trying to sort out the edge case for real sims that you noted so we can use `get`. I want to keep the regtests and users able to run real sims in the meantime.

PR #3140, opened 2026-04-29
[EB] Debug DiffusionSrcForMom & State

skang67:
This PR fixes a couple of issues in the Diffusion functions to apply MOST boundary conditions on the EB:
- The EB facet area is corrected in `DiffusionSrcForState_EB`.
- The if statement routing `no_slip` and `surface_layer` is fixed in `DiffusionSrcForMom_EB`.

The MOST BC for EB is not yet generalized for inclined surfaces and still assumes a flat plane. The neutral ABL case using EB was validated with FittedMesh.

 It also adds inputs files for neutral ABL and GABLS cases.


hgopalan:
There are two ways to do the MOST for inclined surface. In the easier approach, you compute the wind speed parallel to the surface and then compute MOST and then project it back to the Cartesian coordinates. The other option is to compute the stress tensor in local coordinate system and then transform to the cartesian Coordinate system.

skang67:
Thank you for the note, @hgopalan. I am planning to do the MOST calculations in the local surface-aligned coordinates for tangential & friction velocity components, and stress tensor, and then transform the stress tensor to the Cartesian coordinate using the rotational tensor when incorporating it to the DiffusionSrcForMom. For the local coordinates, I will use an orthonormal basis (tangential_x, tangential_y, normal), which is rigidly rotated from the original Cartesian basis. I don't know which option this plan fits to.

PR #3141, opened 2026-04-29
Get Prob Name Real

AMLattanzi:
# Summary
1. Default the prob name to `Unknown`.
2. Query the `init_type` in ReadParams. If init type does **not** involve a netcdf file, get the prob name since it is a canonical case.
3. Now we can use `get` and `query` from [PR 3134](https://github.com/erf-model/ERF/pull/3134).


gardner48 (approved):
Looks good to me. This should fix the problem we had with a bad prefix or missing  problem name in the input file.

Someone could still provide an invalid problem name, but that's a separate issue that can be addressed later. In that case, I'm not sure if there's a cleaner solution than checking the input against all the valid names.

PR #3152, opened 2026-05-06
WIP: [Dycore] Update Theta_m Substep

AMLattanzi:
# Summary
In KSD 07, `theta_m` is updated in the acoustic sub-stepping algorithm since it a direct analog for pressure when moisture is present. Currently, in ERF, only `theta_d` is updated in the acoustic algorithm and `q_v` is updated as a slow variable. This PR examines a test implementation to update `theta_m` in the acoustic solver without having to completely overhaul the entire dycore (until proven worth the effort).

## Approach
To update `theta_m` in the acoustic solver, `q_v` is advected as a fast variable and then updated, via advection, in the acoustic sub-stepping routine. This modification aims to properly advect `theta_m` (pressure) on the small acoustic steps to add robustness to the solver. This approach follows KSD 07 and is a logical extension to the updates for `rho & theta_d`. Note that `q_v` will again be updated in the slow post routine with the average momenta obtained from the acoustic solver. The choice to use or ignore that update is not clear yet. 

## Results
@adamwise95 Please post any results with the PR to the comments.


Source/TimeIntegration/ERF_Substep_T.cpp
```diff
@@ -468,21 +477,29 @@ void erf_substep_T (int step, int /*nrk*/,
...
                                        yflux_lo * (prim(i,j,k,0) + prim(i,j-1,k,0)) ) * dyi * mfsq) * myhalf;
+            temp_rhs_arr(i,j,k,2) = (( xflux_hi * (prim(i,j,k,3) + prim(i+1,j,k,3)) -
+                                       xflux_lo * (prim(i,j,k,3) + prim(i-1,j,k,3)) ) * dxi * mfsq+
+                                     ( yflux_hi * (prim(i,j,k,3) + prim(i,j+1,k,3)) -
+                                       yflux_lo * (prim(i,j,k,3) + prim(i,j-1,k,3)) ) * dyi * mfsq) * myhalf;
 
             if (l_reflux) {
                 (flx_arr[0])(i,j,k,0) = xflux_lo;
                 (flx_arr[0])(i,j,k,1) = (flx_arr[0])(i  ,j,k,0) * myhalf * (prim(i,j,k,0) + prim(i-1,j,k,0));
+                (flx_arr[0])(i,j,k,2) = (flx_arr[0])(i  ,j,k,0) * myhalf * (prim(i,j,k,3) + prim(i-1,j,k,4));
```
pressel: Should the last index here be 3 instead of 4?

Source/TimeIntegration/ERF_Substep_T.cpp
```diff
@@ -691,22 +716,27 @@ void erf_substep_T (int step, int /*nrk*/,
                   avg_zmom_arr(i,j,k+1)      += facinv * zflux_hi / (mf_mx(i,j,0) * mf_my(i,j,0));
                   if (l_reflux) {
                       (flx_arr[2])(i,j,k+1,0) =      zflux_hi / (mf_mx(i,j,0) * mf_my(i,j,0));
-                      (flx_arr[2])(i,j,k+1,1) = (flx_arr[2])(i,j,k+1,0) * myhalf * (prim(i,j,k) + prim(i,j,k+1));
+                      (flx_arr[2])(i,j,k+1,1) = (flx_arr[2])(i,j,k+1,0) * myhalf * (prim(i,j,k,0) + prim(i,j,k+1,0));
+                      (flx_arr[2])(i,j,k+1,3) = (flx_arr[2])(i,j,k+1,0) * myhalf * (prim(i,j,k,3) + prim(i,j,k+1,3));
```
pressel: Since flux[dir] is resized with 3 components, valid indices are 0:2. Should this write to component 2 rather than 3?

Source/TimeIntegration/ERF_Substep_T.cpp
AMLattanzi: Yes that is correct.

Source/TimeIntegration/ERF_Substep_T.cpp
AMLattanzi: Yes that is correct.

hgopalan:
WRF has an option use_theta_m = 1 to add  it. 
Copilot Notes: 

use_theta_m = 1 tells WRF to use moist potential temperature (theta‑m) as the prognostic thermodynamic variable instead of the default potential temperature (theta).

Advantages of use_theta_m = 1

Reduces numerical noise associated with latent heating in moist regions.
Improves thermodynamic consistency because water vapor effects are embedded in the prognostic variable.
Can enhance stability in very moist environments (e.g., deep convection, tropical simulations).
Helps better maintain energy conservation in strongly precipitating systems.
Beneficial in high‑resolution WRF runs where moisture gradients are large.


Disadvantages of use_theta_m = 1

Can slightly increase computational cost due to extra conversions between variables.
Not always well‑tested across all physics combinations; some microphysics or cumulus schemes may not behave as expected.
May introduce differences in temperature fields compared to legacy configurations, requiring experimentation to tune.
Users may need to validate results carefully, especially for operational setups that historically used theta.

adamwise95:
My test case is a nx,ny=400, dx=1km run centered over Oklahoma. The simulation is stable if I use a fixed dt of 1 [s]. With WRF, the same case can use a dt of 5 [s]. If I use a dt of 5 [s] in ERF, the simulation crashes at step 7. The figure below shows checkerboarding in the qv field (southeast area of the domain) leading to the crash.

image


With this PR, I can run up to step 1746. The figure below shows the qv field the step before the crash and we can see checkerboarding in the qv field in the western boundary relaxation zone:



image

PR #3166, opened 2026-05-11
Apply bdy tendencies and change base state when using wrfinput

asalmgren:


pressel:
Did you intend to add ERF_InitFromWRFInput.cpp.new?

Source/Initialization/ERF_InitFromWRFInput.cpp
```diff
@@ -1210,6 +1108,171 @@ init_base_state_from_wrfinput (const Box& subdomain,
...
+
+                  // Do iterations
+                  HSEutils::Newton_Raphson_hse(tol, R_d/Cp_d, dz,
+                                               grav, C, Th_hi,
+                                               qv_hi, qv_hi,
+                                               P_hi, R_hi, F);
+
+                  // Assign data
+                  r_hse_arr(i,j,k) = R_hi;
+                  p_hse_arr(i,j,k) = getPgivenRTh(R_hi*th_hse_arr(i,j,k), 0.0);
```
pressel, line 1260: I think this mixes two theta profiles. The HSE solve passes `Th_hi = 300.0` into `Newton_Raphson_hse`, so the returned `P_hi` and `R_hi` are consistent with a 300 K theta. But `th_hse_arr` is not 300 K everywhere; with the formula above it is 288.15 K at z=0, about 300 K only near 3.4 km, and about 332.6 K at 11 km. The stored pressure is then recomputed from `R_hi * th_hse_arr(i,j,k)`, so `p_hse`, `r_hse`, and `th_hse` are generally inconsistent with the HSE solve. Should the Newton solve use `th_hse_arr(i,j,k)`, or should `th_hse_arr` be made consistent with the 300 K isentropic assumption?

asalmgren: I need to fix this -- my focus was on getting something in place but I really want this to look more like what WRF does since the pathway is for when we start with wrfinput files...

PR #3174, opened 2026-05-13
more generalization for single precision

asalmgren:
This PR captures most of the changes from https://github.com/erf-model/ERF/pull/3133

### DIFF
--- a/Source/ERF_MakeNewArrays.cpp
+++ b/Source/ERF_MakeNewArrays.cpp
@@ -166,10 +166,10 @@ ERF::init_stuff (int lev, const BoxArray& ba, const DistributionMapping& dm,
     lev_old[Vars::yvel].define(convert(ba, IntVect(0,1,0)), dm, 1, ngrow_vels);
 
     // Set these to avoid operations on uninitialized data
-    lev_new[Vars::xvel].setVal(1.234e20);
-    lev_old[Vars::xvel].setVal(1.234e20);
-    lev_new[Vars::yvel].setVal(1.234e20);
-    lev_old[Vars::yvel].setVal(1.234e20);
+    lev_new[Vars::xvel].setVal(1.234e10);
+    lev_old[Vars::xvel].setVal(1.234e10);
+    lev_new[Vars::yvel].setVal(1.234e10);
+    lev_old[Vars::yvel].setVal(1.234e10);
 
     // Note that we need the ghost cells in the z-direction if we are doing any
     // kind of domain decomposition in the vertical (at level 0 or above)
--- a/Source/Utils/ERF_HSEUtils.H
+++ b/Source/Utils/ERF_HSEUtils.H
@@ -475,7 +475,14 @@ Real compute_p_k (Real &p_k, const Real p_k_minus_1, Real &theta_k, Real& rho_k,
                   const Real theta_0, const Real theta_tr, const Real z_tr, const Real T_tr)
 {
     Real delta_p_k;
+
+#ifdef AMREX_USE_FLOAT
+    Real eps = Real(1.e-6);
+    Real tol = Real(1.e-8);
+#else
     Real eps = Real(1.e-10);
+    Real tol = Real(1.e-8);
+#endif
 
     for(int iter=0; iter<20; iter++)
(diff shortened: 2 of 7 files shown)


Source/Utils/ERF_HSEUtils.H
```diff
@@ -475,7 +475,14 @@ Real compute_p_k (Real &p_k, const Real p_k_minus_1, Real &theta_k, Real& rho_k,
                   const Real theta_0, const Real theta_tr, const Real z_tr, const Real T_tr)
 {
     Real delta_p_k;
+
+#ifdef AMREX_USE_FLOAT
+    Real eps = Real(1.e-6);
+    Real tol = Real(1.e-8);
+#else
     Real eps = Real(1.e-10);
```
pressel, line 483: I don't think this fixes the single-precision issue. For p_k ~ 1e5 Pa, p_k + 1e-6 will still round back to p_k in float, so F_plus_dF can equal F and F_prime can become zero. Can we use a relative perturbation here, e.g. max(abs(p_k) * Real(1.e-5), Real(1.e-5))? I think we should also explicitly reject non-finite delta_p_k, since fabs(NaN) > tol is false.

asalmgren: Agreed -- this was the part of Dave's PR that I wasn't sure what the right fix was so this is a placeholder.  Could you go ahead and submit a PR to do something more sensible here?

Source/Utils/ERF_HSEUtils.H
```diff
@@ -490,8 +497,7 @@ Real compute_p_k (Real &p_k, const Real p_k_minus_1, Real &theta_k, Real& rho_k,
         p_k            = p_k + delta_p_k;
     }
 
-
-    if (std::fabs(delta_p_k) > 1e-8) {
+    if (std::fabs(delta_p_k) > tol) {
```
pressel, line 500: This will not catch NaNs. If F_prime becomes zero in single precision, delta_p_k can be NaN and this condition will be false. Should this check also require std::isfinite(delta_p_k)?

Source/ERF_MakeNewArrays.cpp
```diff
@@ -166,10 +166,10 @@ ERF::init_stuff (int lev, const BoxArray& ba, const DistributionMapping& dm,
     lev_old[Vars::yvel].define(convert(ba, IntVect(0,1,0)), dm, 1, ngrow_vels);
 
     // Set these to avoid operations on uninitialized data
-    lev_new[Vars::xvel].setVal(1.234e20);
-    lev_old[Vars::xvel].setVal(1.234e20);
-    lev_new[Vars::yvel].setVal(1.234e20);
-    lev_old[Vars::yvel].setVal(1.234e20);
+    lev_new[Vars::xvel].setVal(1.234e10);
```
pressel, line 169: Should this change stay inside AMREX_USE_FLOAT? The original issue is single-precision-specific, but this changes the double-precision sentinel too. Also, this only initializes xvel/yvel; zvel is defined just below but still does not appear to be filled before custom perturbations may be added.

asalmgren: This is simply a protection against operating on uninitialized data -- if this changes has any effect at all on the solution then there is something we don't understand.   The idea is just to put a big enough number here that we will quickly catch if we actually use these values.  So I think changing this for double precision is safe (or will tell us we are confused which is also useful)

Source/Utils/ERF_Interpolation_WENO.H
```diff
@@ -121,7 +121,11 @@ struct WENO3
 
 private:
     amrex::Array4<const amrex::Real> m_phi;   // Quantity to interpolate
+#ifdef AMREX_USE_FLOAT
+    const amrex::Real eps=amrex::Real(1.0e-12);
```
pressel, line 125: This fixes the float underflow/Inf issue for uniform fields, but the WENO eps value also affects conservation. Before merging this across WENO/WENO-Z, can we confirm with an advecting-scalar test and/or BOMEX with WENO in single precision? See PR #3133.

PR #3177, opened 2026-05-13
Base State WRFInput

AMLattanzi:
# Summary 
The base state needs to be set such that `\partial_{x_h} P_0 \approx 0`. This PR now follows WRF in using a laterally homogeneous and dry state for HSE.

## Regtest Implications
This PR will break the following regtests

1. WPS_Terrain
2. MetGrid

## WRF Equations
image

## Algorithm
Setting of the base state will have two possible forks (yielded nearly identical results for WPS):
1. Read `PB & ALB` directly and use those to deduce other HSE quantities (EOS give dry temperature)
3. ALB is not present, then we default to using the WRF equations with the constants parsed from the inputs file: `T00, P00, TLP, TISO, TLP_STRAT, P_STRAT`. Compute Td from equation, populate Theta, then deduce density.

## Result
Qv in the large tornado domain at 30 m AGL after 8 hrs of simulation (6 ITC --> 14 UTC):
image


Source/Initialization/ERF_InitFromMetgrid.cpp
```diff
@@ -1173,19 +1172,165 @@ init_base_state_from_metgrid (const bool use_moisture,
...
+    //***********************************************************************************
+    {
+        Box valid_bx2d = valid_bx;
+        valid_bx2d.setRange(2,0);
+
+        // Expose for copy to GPU
+        int khi = ubound(valid_bx).z;
+        int klo = lbound(valid_bx).z;
+
+        // ARW V4 Constants (5.2.2 Reference State)
```
pressel, line 1191: This is the right ARW reference-state form, but should these constants come from the WRFInput file or ERF inputs rather than being hard-coded?

WRF treats the reference-state parameters (`p00`, `t00`, `tlp/base_lapse`, `tiso`, `tlp_strat`, `p_strat`, `p_top`) as configurable/input values. If the `wrfinput`/`wrfbdy` pair was generated with non-default reference-state constants, ERF would now construct a different dry base state than the one WRF used.

At minimum, I think we should print the values used here. Ideally, we should read the WRF values when present and fall back to these defaults only when absent.

Source/Initialization/ERF_InitFromWRFInput.cpp
```diff
@@ -1134,78 +1015,116 @@ ERF::init_from_wrfinput (int lev,
...
+    const Real P_strat = Real(10000.);
+
+    for (MFIter mfi(p_hse, TileNoZ()); mfi.isValid(); ++mfi ) {
+
+        // Verify box spans vertical extent
         Box tbx  = mfi.tilebox();
+        AMREX_ALWAYS_ASSERT( (klo == tbx.smallEnd(2)) &&
+                             (khi == tbx.bigEnd(2)) );
+
+        // Lateral grown box for ghost cells
```
pressel: Can we document the intended lateral ghost-cell policy for the base state here?

In this WRFInput path, we compute columns over a laterally grown slab using `z_phys_nd` in the grown region. In the Metgrid path, the new dry base state is computed on `valid_bx2d` and then FOEXTRAP is used for lateral ghost cells.

Should WRFInput and Metgrid use the same policy: compute base state from ghost terrain, or extrapolate from the valid interior? I do not have a strong preference, but I think the choice should be explicit and consistent.

Source/Initialization/ERF_InitFromWRFInput.cpp
```diff
@@ -1134,78 +1015,116 @@ ERF::init_from_wrfinput (int lev,
...
+                Rd_lo = Rd_hi;
+                z_lo  = z_hi;
             }
 
-            r_hse_arr(i,j,k)  = Rhse_Sum;
-            p_hse_arr(i,j,k)  = Ptot;
-            pi_hse_arr(i,j,k) = getExnergivenP(p_hse_arr(i,j,k), l_rdOcp);
-            th_hse_arr(i,j,k) = getRhoThetagivenP(p_hse_arr(i,j,k), Qv) / cons_arr(ii,jj,kk,Rho_comp);
-            qv_hse_arr(i,j,k) = Qv;
+            // Fill top and bottom ghost cells
```
pressel: Does `base_state` ever require more than one vertical ghost cell in the pressure-gradient/HSE paths? This fills only `klo-1` and `khi+1`, while the previous grown-box/clamping approach populated the whole grown region. If one layer is sufficient, a short assertion or comment would make that clear.

pressel (commented):
See inline comments.

Source/Initialization/ERF_InitFromWRFInput.cpp
AMLattanzi: No, only one ghost cell is needed at the top and bottom.

Source/Initialization/ERF_InitFromWRFInput.cpp
AMLattanzi: The state is laterally homogeneous, so there is no difference in lateral ghost cell policy (both yield FOEXTRAP). The difference is in how the boxes are looped over---i.e., grown box versus a valid box with explicit copies into the ghost cell. Both achieve the same result.

pressel: Thanks, I traced this through the WRFInput terrain path and agree.

Source/Initialization/ERF_InitFromMetgrid.cpp
AMLattanzi: I do not see these values in any of the `wrfinput` or `wrbdy` files I have. We can easily print the current values. However, a discussion is in order if we wish to parse them from somewhere.

asalmgren (approved):

Source/Initialization/ERF_InitFromMetgrid.cpp
asalmgren: My opinion is that we go ahead and merge this PR then open an issue pointing out that we could read these in from wrfinput / wrfbdy files.   Given that we know this is better than what is there, we should not delay in merging, but we should remember we can generalize later.   @pressel -- would you be good with that?   @cmkaul -- could you give this PR a try on your case that has boundary issues and confirm the solution is hopefully better but at least no worse?

Source/Initialization/ERF_InitFromMetgrid.cpp
```diff
@@ -1173,19 +1172,165 @@ init_base_state_from_metgrid (const bool use_moisture,
...
+        const Real T0      = Real(300.0);
```
pressel: I think T0 = 300.0K and P_STRAT may be incorrect. See [here](https://github.com/wrf-model/WRF/blob/f15568ccc1447780e3bd664b9f0196edd784bf33/Registry/Registry.EM_COMMON#L2910) and [here](https://github.com/wrf-model/WRF/blob/f15568ccc1447780e3bd664b9f0196edd784bf33/Registry/Registry.EM_COMMON#L2913):

Please confirm.

This appears to use the wrong WRF constant. WRF’s real-data base-state t00 defaults to 290 K, while 300 K is WRF’s perturbation potential-temperature offset t0. Also, WRF’s default base_pres_strat is 0 Pa, not 10000 Pa. If the intent is to match WRF we need to use WRF’s defaults.

pressel (changes requested):

Source/Initialization/ERF_InitFromMetgrid.cpp
AMLattanzi: The noted variables are now parsed. For whatever reason, they didn't show up in ncview (caused the confusion).

Source/Initialization/ERF_InitFromMetgrid.cpp
AMLattanzi: These are parsed now.

Source/Initialization/ERF_InitFromWRFInput.cpp
```diff
@@ -101,15 +104,83 @@ read_start_time_from_wrfinput(int lev, const std::string& fname)
...
+                                      Real& TISO,
+                                      Real& TLP_STRAT,
+                                      Real& P_STRAT)
+{
+    if (ParallelDescriptor::IOProcessor()) {
+        auto ncf = ncutils::NCFile::open(fname, NC_CLOBBER | NC_NETCDF4);
+
+        std::vector<size_t> shape;
+        std::vector<size_t> start;
+        int success = ncf.has_var("T00");
```
pressel, line 131: Thanks for adding the WRF base-state metadata path. I think this still needs one correction before we can rely on it: in typical WRF `wrfinput`/`wrfout` files, `T00`, `P00`, `TLP`, `TISO`, `TLP_STRAT`, and `P_STRAT` are global attributes, not NetCDF variables. 

This code currently uses `has_var("T00")` / `var("T00")`, so it will likely fail to read the values and silently fall back to the defaults even when the file contains the metadata.

Can we use `has_attr()` / `get_attr()` here, or support both global attributes and variables

Source/Initialization/ERF_InitFromWRFInput.cpp
```diff
@@ -101,15 +104,83 @@ read_start_time_from_wrfinput(int lev, const std::string& fname)
...
+        if (success) {
+            shape = ncf.var("P_STRAT").shape();
+            start.resize(shape.size(), 0);
+            ncf.var("P_STRAT").get(&P_STRAT ,start, shape);
+        }
+
+        ncf.close();
+    }
+
+    ParallelDescriptor::Bcast(&T00,1,ParallelDescriptor::IOProcessorNumber());
```
pressel, line 186: Could we print the final values used for `T00`, `P00`, `TLP`, `TISO`, `TLP_STRAT`, and `P_STRAT`, and ideally whether each came from the file or from the fallback default? That would make it much easier to verify this path in logs and catch unexpected fallbacks.

pressel (changes requested):
See inline comments.

Source/Initialization/ERF_InitFromWRFInput.cpp
AMLattanzi: Will add a print

Source/Initialization/ERF_InitFromWRFInput.cpp
AMLattanzi: `has_attr & get_attr` was the initial approach. This failed with `attribute not found` even when the `has_attr` returned that the variable was indeed in the netcdf file.

pressel: Thanks, @AMLattanzi . You are correct, they are variables not attributes.

pressel (approved):

PR #3199, opened 2026-05-24
Kessler testing and fixes

pressel:
## Summary

This PR adds a Kessler microphysics test suite and fixes the production bugs that the tests exposed.

The new tests cover scalar helper contracts, host/device helper behavior, public-flow conservation, rain accumulation, sedimentation, and Kessler_NoRain behavior. The shared test header documents the core contracts: local source terms conserve total water, sedimentation moves rain vertically and to `rain_accum`, public-flow water budgets include surface rain, pressure units are explicit, and GPU tests keep assertions on the host.

### Tests

- Add Kessler scalar tests for saturation adjustment, warm-rain source terms, terminal velocity, precipitation flux, sedimentation helpers, copy-in/copy-out contracts, and documented thresholds.
- Add Kessler kernel tests for host/device helper equivalence.
- Add public-flow tests that run:

  `Init -> Copy_State_to_Micro -> Advance -> Copy_Micro_to_State`

  and check total water, latent-energy proxy behavior, surface rain accumulation, detJ-weighted column budgets, and NoRain behavior.
- Add compressed-cell detJ coverage with `detJ < 1`.

### Production fixes

- Use terminal velocity, not precipitating mass flux, for the sedimentation substep CFL.
- Use the upper-cell donor state for downward sedimentation faces.
- Limit sedimentation face flux by the donor cell’s detJ-weighted available rain water.
- Stop mutating shared `fz` face values while thresholding small sedimentation fluxes.
- Use one latent-over-cp factor, `L_v / sc.c_p`, for both saturation adjustment and theta update.
- Limit rain evaporation by the subsaturation left after cloud evaporation.
- Prevent cloud-to-rain conversion from consuming cloud water that saturation adjustment already evaporated.
- Convert rain accumulation from precipitation mass per area to millimeters with named constants.
- Replace the hidden `myhalf` sedimentation CFL divisor with a named Kessler CFL constant.

## Why this changes answers

This PR is expected to change bit-for-bit answers in Kessler cases. Those changes are intentional.

The old code used the stored precipitation flux `rho * Vt * qp` to choose the number of sedimentation substeps. That quantity has flux units, not velocity units. The new code still stores flux in `fz`, but it reduces maximum terminal velocity for the CFL count. This can change the number of rain substeps, rain transport, and rain accumulation. The old reduction returned `fz`; the new reduction recomputes terminal velocity. 

The old sedimentation update also wrote thresholded zeros back into shared `fz` faces. Neighboring cells share those faces, so the update could depend on traversal and backend behavior. The new code thresholds local face-flux copies only.

The old limiter capped face flux without detJ. For compressed cells with `detJ < 1`, that could remove more rain than the donor cell contained before the nonnegative clip hid the error. The new cap uses the donor cell’s detJ-weighted available rain water.

The old saturation adjustment used `L_v / Cp_d`, while theta used the solver heat capacity. The new code derives one factor from `L_v / sc.c_p` and passes it through the Kessler source path.

The old rain evaporation used the original vapor deficit. Cloud evaporation and rain evaporation could both consume the same subsaturation. The new code applies cloud evaporation first, then caps rain evaporation by the remaining linearized capacity.

These changes fix physical and mathematical defects. They should not be judged by bit-for-bit agreement with the old Kessler results.

## Notes for reviewers

The PR should be nearly performance netural:

- no new production `MultiFab` allocations,
- no new production kernel launches,
- no new production reductions beyond the existing sedimentation reduction changing what it reduces,
- no dynamic allocation in device code,
- helper functions remain `AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE`.

The empirical Kessler closure coefficients remain unchanged. Generic thermodynamic constants come from `ERF_Constants.H`.

### DIFF
--- a/Source/Microphysics/Kessler/ERF_Kessler.cpp
+++ b/Source/Microphysics/Kessler/ERF_Kessler.cpp
@@ -68,7 +68,7 @@ void Kessler::AdvanceKessler (const SolverChoice &solverChoice)
 
                 const KesslerSourceTerms source_terms = kessler_warm_rain_sources(
                     qv_array(i,j,k), qc_array(i,j,k), qp_array(i,j,k), rho_array(i,j,k),
-                    pressure, qsat_local, dtqsat_local, dtn, do_cond);
+                    pressure, qsat_local, dtqsat_local, dtn, do_cond, d_fac_cond);
 
                 qv_array(i,j,k) += -source_terms.dq_vapor_to_cloud
                                  +  source_terms.dq_cloud_to_vapor
@@ -163,11 +163,13 @@ void Kessler::AdvanceKessler (const SolverChoice &solverChoice)
                         kessler_face_state(k, k_lo, k_hi, rho_km1, rho_k, qp_km1, qp_k);
 
                     const Real terminal_velocity = kessler_terminal_velocity(face_state.rho, face_state.qp);
-                    fz_array(i,j,k) = kessler_precip_flux(face_state.rho, terminal_velocity, face_state.qp);
+                    const Real max_flux = face_state.rho * face_state.qp / coef;
+                    fz_array(i,j,k) = amrex::min(
+                        kessler_precip_flux(face_state.rho, terminal_velocity, face_state.qp), max_flux);
 
                     if(k==k_lo){
                         rain_accum_array(i,j,k) = rain_accum_array(i,j,k)
-                                                + face_state.rho * face_state.qp * terminal_velocity * dtn / Real(1000.0) * Real(1000.0);
+                                                + fz_array(i,j,k) * dtn / Real(1000.0) * Real(1000.0);
                     }
                 });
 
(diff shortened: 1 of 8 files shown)


asalmgren (approved):

Source/Microphysics/Kessler/ERF_KesslerUtils.H
```diff
@@ -99,16 +117,14 @@ KesslerFaceState kessler_face_state (const int k,
...
-    } else if (k == k_hi + 1) {
+    // Downward sedimentation uses the upper-cell donor state at interior faces.
+    // The top boundary face uses the top cell below the boundary.
+    if (k == k_hi + 1) {
         face_state.rho = rho_km1;
         face_state.qp = qp_km1;
     } else {
-        face_state.rho = myhalf * (rho_km1 + rho_k);
-        face_state.qp = myhalf * (qp_km1 + qp_k);
+        face_state.rho = rho_k;
```
AMLattanzi, line 126: Removing the averaging here means that we are using a cell centered density. However, the sedimentation fluxes, which are at the z-faces, appear to be multiplied by these values [here](https://github.com/pressel/ERF/blob/a9d4e8c169cbf3a8fa3cf2edd4df8a05ddb765d3/Source/Microphysics/Kessler/ERF_KesslerUtils.H#L57).

Source/Microphysics/Kessler/ERF_Kessler.cpp
```diff
@@ -107,23 +107,35 @@ void Kessler::AdvanceKessler (const SolverChoice &solverChoice)
                 const Real qp_km1 = (k == k_lo) ? qp_array(i,j,k) : qp_array(i,j,k-1);
                 const Real qp_k = (k == k_hi+1) ? qp_array(i,j,k-1) : qp_array(i,j,k);
                 const KesslerFaceState face_state =
-                    kessler_face_state(k, k_lo, k_hi, rho_km1, rho_k, qp_km1, qp_k);
+                    kessler_face_state(k, k_hi, rho_km1, rho_k, qp_km1, qp_k);
 
                 const Real terminal_velocity = kessler_terminal_velocity(face_state.rho, face_state.qp);
                 fz_array(i,j,k) = kessler_precip_flux(face_state.rho, terminal_velocity, face_state.qp);
```
AMLattanzi, line 113: Face states are passed in here to compute the sedimentation flux; note fz_arry has a z nodal BA [here](https://github.com/pressel/ERF/blob/a9d4e8c169cbf3a8fa3cf2edd4df8a05ddb765d3/Source/Microphysics/Kessler/ERF_Kessler.cpp#L30).

Source/Microphysics/Kessler/ERF_KesslerUtils.H
pressel: Thanks @AMLattanzi, you are correct. This will be fixed with #3206.

Source/Microphysics/Kessler/ERF_Kessler.cpp
pressel: Thanks @AMLattanzi, you are correct. This will be fixed with #3206.

PR #3204, opened 2026-05-26
Initialize TKE to erf.tke_min with default 1e-6 when using a scheme t…

asalmgren:
…hat needs it

In the current code we have default initialization of zero for TKE unless a perturbation is specified.  This PR initializes the TKE to tke_min = 1e-6 which can be over-ridden in the inputs file by "erf.tke_min =".


hgopalan:
We can also specify an analytical profile k(z) = ustar^2/sqrt(Cmu) * (1-z/H)^2 with star and planetary boundary layer height specified as some approximate value and Cmu= 0.03-0.09.

pressel (approved):

PR #3206, opened 2026-05-26
Use face density for Kessler rain sedimentation

pressel:
## Summary

This PR fixes the Kessler sedimentation face-state treatment.

The previous Kessler update correctly upwinded rain mixing ratio for downward sedimentation, but it also used donor-cell density in the z-face flux. That mixed a donor-state choice for `qp` with a face-location choice for `rho`.

This patch separates those roles:

- use face-centered density for the z-face flux and terminal velocity;
- use donor/upwind rain mixing ratio for downward sedimentation;
- use donor-cell `rho`, donor-cell `qp`, and donor-cell `detJ` only for the donor availability limiter.

## Why this matters

`fz` is a face-centered sedimentation flux. The rain mixing ratio should be upwinded for one-way downward sedimentation, but density in the flux should remain located at the face.

The donor limiter still needs donor-cell quantities because it caps outgoing rain by the water available in the donor cell.

This keeps the finite-volume water cap while making the face flux location consistent.

## Tests

This PR updates the Kessler face-state tests to check:

- bottom face: `rho_k`, `qp_k`;
- interior face: face-centered `rho`, donor `qp`;
- top boundary face: `rho_km1`, `qp_km1`.

It also keeps the existing Kessler sedimentation and detJ column-budget tests in place.

## Performance and portability

This change adds no new production kernels, reductions, or `MultiFab` allocations.

The helper path remains AMReX GPU-callable and uses `amrex::Real`.

## Expected answer changes

This may change Kessler results compared with the previous merged implementation. That change is intentional. The flux now uses face-centered density instead of donor-cell density, while preserving donor/upwind rain transport.

### DIFF
--- a/Source/Microphysics/Kessler/ERF_KesslerUtils.H
+++ b/Source/Microphysics/Kessler/ERF_KesslerUtils.H
@@ -117,13 +117,17 @@ KesslerFaceState kessler_face_state (const int k,
 {
     KesslerFaceState face_state{amrex::Real(0), amrex::Real(0)};
 
-    // Downward sedimentation uses the upper-cell donor state at interior faces.
-    // The top boundary face uses the top cell below the boundary.
+    // Downward sedimentation upwinds rain mixing ratio from the donor cell.
+    // Air density remains face-centered for the z-face flux. The donor-cell
+    // density is used separately only for the availability limiter.
     if (k == k_hi + 1) {
         face_state.rho = rho_km1;
         face_state.qp = qp_km1;
-    } else {
+    } else if (k == 0) {
         face_state.rho = rho_k;
+        face_state.qp = qp_k;
+    } else {
+        face_state.rho = amrex::Real(0.5) * (rho_km1 + rho_k);
         face_state.qp = qp_k;
     }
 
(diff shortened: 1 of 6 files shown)


Source/Microphysics/Kessler/ERF_KesslerUtils.H
```diff
@@ -117,14 +117,18 @@ KesslerFaceState kessler_face_state (const int k,
...
     if (k == k_hi + 1) {
         face_state.rho = rho_km1;
         face_state.qp = qp_km1;
-    } else {
+    } else if (k == 0) {
         face_state.rho = rho_k;
         face_state.qp = qp_k;
+    } else {
+        face_state.rho = amrex::Real(0.5) * (rho_km1 + rho_k);
+        face_state.qp = qp_k;
```
AMLattanzi, line 135: Qp at the face should also be averaged since it is passed in to compute the sedimentation flux.

AMLattanzi (changes requested):

AMLattanzi (approved):

Source/Microphysics/Kessler/ERF_KesslerUtils.H
pressel: Thanks, that is a fair question. I agree that `rho` should be treated as a face quantity here, since `fz` is a face flux.

I think `qp` is different. Rain sedimentation is one-way downward transport, so the rain amount at a face should come from the cell the rain is falling out of. If we average `qp`, a wet cell below a dry cell can create rain flux through the face above it, even though there is no rain in the donor cell to fall through that face.

So the intent is:

- use face-centered `rho` for the face flux;
- use donor/upwind `qp` for falling rain;
- use donor `rho`, `qp`, and `detJ` only to cap the amount of rain that can leave the donor cell.

If we want a higher-order treatment later, I think the right next step would be a limited upwind reconstruction of `qp`, rather than a plain average.

PR #3232, opened 2026-05-30
Fix gtest discovery and GPU math portability

pressel:
## Summary

This PR fixes follow-up portability issues from the recent GitLab CI/gtest updates.

It makes three focused changes:

- flushes stdio before using `std::_Exit` in the `--gtest_list_tests` path, so CMake/GoogleTest discovery can reliably consume the test list while still avoiding AMReX/MPI/GPU teardown hangs on affected systems;
- replaces host `std` trigonometric calls in GPU-qualified interpolation test helpers with AMReX device-aware math helpers;
- splits `ERF_PARALLEL_TEST_MPIEXEC_PREFLAGS` before appending it to the CTest MPI launch command, matching the handling of the standard MPI launcher variables.

## Motivation

The prior CI/gtest change (PR #3228) intentionally skipped `amrex::Initialize` during GoogleTest discovery and used `_Exit` to avoid runtime teardown hangs. Since `_Exit` bypasses normal stream cleanup, the test-listing path should explicitly flush stdout/stderr first.

The interpolation smooth-periodic helpers are marked `AMREX_GPU_HOST_DEVICE`, so their trigonometric calls should use AMReX math utilities rather than host-only `std` calls.

The MPI launcher handling already split standard MPI variables for CTest, but user-provided MPI preflags still needed the same treatment.

### DIFF
--- a/Tests/Unit/CMakeLists.txt
+++ b/Tests/Unit/CMakeLists.txt
@@ -87,8 +87,10 @@ if(ERF_ENABLE_MPI)
 
   if(DEFINED ERF_PARALLEL_TEST_MPIEXEC_PREFLAGS AND
      NOT "${ERF_PARALLEL_TEST_MPIEXEC_PREFLAGS}" STREQUAL "")
+    separate_arguments(erf_parallel_test_user_mpiexec_preflags
+      UNIX_COMMAND "${ERF_PARALLEL_TEST_MPIEXEC_PREFLAGS}")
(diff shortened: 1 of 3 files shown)


gardner48:
The changes look good but two tests are now failing on one of our AMD systems with this branch:

* SatAdjBoxCoverage.PeriodicGhostFillAfterFullPublicFlow
* erf_parallel_tests_np2

I'll take a closer look shortly and let you know what I find.

asalmgren (approved):

PR #3233, opened 2026-05-30
Fixes GPU portability bugs in the SatAdj unit tests.

pressel:
## Summary

This PR fixes GPU portability bugs in the SatAdj unit tests. See comments to #3232. 

It makes three focused changes:

- replaces duplicated `std::initializer_list` ghost-cell poisoning helpers with one shared `amrex::GpuArray` helper;
- updates the serial and parallel SatAdj ghost-fill tests to use that device-safe helper;
- rewrites the moist-temperature diagnostic test so host-side checks read host mirrors, not device-resident data.

## Motivation

CUDA and HIP CI reported illegal GPU memory accesses in the SatAdj ghost-fill tests. CUDA also exposed a segfault in the moist-temperature diagnostic test.

These were test portability bugs. The ghost-fill tests captured host-only initializer-list state in GPU lambdas. The diagnostic test read data on the host after a GPU kernel wrote it.

## Impact

This PR changes test code only. It does not change production model code, numerical algorithms, or physics. It should not affect BFB results or model solutions.

The tests keep the same coverage and assertions. They only change how test data moves across host and device.

### DIFF
--- a/Tests/Unit/Microphysics/SatAdj/ERF_GTestSatAdjBoxCoverage.cpp
+++ b/Tests/Unit/Microphysics/SatAdj/ERF_GTestSatAdjBoxCoverage.cpp
@@ -91,27 +91,6 @@ void run_kessler_norain_public_flow (Kessler& kessler,
     });
 }
 
-void poison_ghost_cells (amrex::MultiFab& mf,
-                         const std::initializer_list<int>& components,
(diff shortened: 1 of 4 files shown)


gardner48:
Thanks @pressel this fixes the CI failures I was seeing

PR #3249, opened 2026-06-03
CI: Add single precision tests to GitLab

gardner48:
Add tests with single/double precision mesh data and particles
Single precision tests are build-only for now


asalmgren:
@gardner48 -- could you add these to the Github CI as well??

PR #3265, opened 2026-06-05
Allow enough ghost cells to correctly fill z_phys from PH and PHB

asalmgren:


Source/Initialization/ERF_InitFromWRFInput.cpp
```diff
@@ -564,7 +564,9 @@ ERF::init_from_wrfinput (int lev, MultiFab& mf_PSFC_lev)
               if (success) {
                   // NOTE: We call FillBoundary on mf_PH below
                   auto& ba_w = lev_new[Vars::zvel].boxArray();
-                  mf_PH.define(ba_w, dm, 1, IntVect(1,1,0));
+                  int ng_x = z_phys_nd[lev]->nGrowVect()[0]+1;
```
AMLattanzi: ngz is defined at the start of this routine and we **were** using that to set the ghost cells previously. I think we should just create the following and use it for PH and PHB regardless of the level.
```
IntVect ngp = ngz; ngp[0] += 1; ngp[1] += 1; ngp[2] = 0;
```

AMLattanzi (changes requested):

Source/Initialization/ERF_InitFromWRFInput.cpp
```diff
@@ -564,7 +564,9 @@ ERF::init_from_wrfinput (int lev, MultiFab& mf_PSFC_lev)
...
+                  int ng_y = z_phys_nd[lev]->nGrowVect()[1]+1;
+                  mf_PH.define(ba_w, dm, 1, IntVect(ng_x,ng_y,0));
```
AMLattanzi: With the larger ghost cell count on `PH/PHB` and FB calls on both those variables, is the z_phys FB call now necessary?

AMLattanzi (approved):

PR #3282, opened 2026-06-09
Update OceanSurf packing for 2way Remora

jmsexton03:


Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -956,7 +960,7 @@ SurfaceLayer::get_lsm_tsurf (const int& lev)
         ParallelFor(gtbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept
         {
             int is_land = (lmask_arr) ? lmask_arr(i,j,k) : 1;
-            if (is_land) {
+            if (copy_all_points || is_land) {
```
AMLattanzi: Is the current intent to have the ocean model copy `t_surf` even in regions where the land mask is `true`? It seems like we may want to change the logic here to copy from `t_surf` over land only if it's a true land surface model and over sea only if it's an ocean model. For the Remora sims, you would just want land mask to be 0 everywhere.
```
const bool has_sea_tsurf = (m_has_ocean_lsm_tsurf &&
                                  amrex::toLower(m_lsm_data_name[m_lsm_tsurf_indx]) == "t_surf");
...
...
if ( (!has_sea_tsurf && is_land) ||
      (has_sea_tsurf && !is_land) ) {
```

AMLattanzi (changes requested):
Minor changes associated with my comment. Everything else looks good. Thank you Jean!

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
jmsexton03: That does make more sense... For this file would it directly be:
```suggestion
            const bool has_sea_tsurf = (m_has_ocean_lsm_tsurf &&
                                       amrex::toLower(m_lsm_data_name[m_lsm_tsurf_indx]) == "t_surf");
            if ( (!has_sea_tsurf && is_land) ||
                  (has_sea_tsurf && !is_land) ) {
```
and then set `erf.is_land = 0` in `.Exec_dev/RemoraCoupling/inputs*`?

AMLattanzi: Yes. But I think we want two more changes to `Source/ERF_MakeNewArrays.cpp`  and the `ReadParameters` routine in `ERF.cpp`. We can parse the value `erf.is_land = int` from the inputs file. Then we use the value to run the setVal at L486 of MakeNewArrays. If this isn't clear I can push to this PR.

jmsexton03: In [586d49f](https://github.com/erf-model/ERF/pull/3282/commits/586d49fe625748c97165852324a6effa19f100d2) I moved the `is_land` flag to `solverChoice` / `ERF_DataStruct.H` since it was in 3 places. If I didn't follow your intent feel free to push to the PR. I merged in development, so the actions may take a while.

AMLattanzi (approved):

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
AMLattanzi: Your changes already addressed my comments. All approved.

PR #3287, opened 2026-06-10
Add plane_sampling_vars support to PlaneSampler

rybchuk:
Hi there, this is my first time adding a feature to this repo, so please let me know if I need to make changes.

# Summary
Add a `plane_sampling_vars` functionality that mirrors the existing `line_sampling_vars` functionality.

# Changes
* Corrected a comment in `LineSampler`
* Added `m_varnames {"magvel","theta"}` data member to PlaneSampler (same default as LineSampler)
* Added `plane_sampling_vars` parsing block in the PlaneSampler constructor, printing selected variables at startup
* Refactored get_sample_data to replace the hardcoded two-component kernel (`theta + magvel`) with per-variable-gated blocks, supporting the full variable set
* Updated `write_sample_data` to pass `m_varnames` to `WriteMultiLevelPlotfile` instead of a hardcoded `T, Wsp`
* Updated the docs `Inputs.rst`
* Breaking change: existing post-processing scripts that reference `T` and `Wsp` will need to change references to `theta` and `magvel`

I ran a test case and confirmed that a plane sampler file was generated with the requested variables.


asalmgren:
@rybchuk -- this looks good except at line 135 I'm wondering if dfab(i,j,k,qv_comp)); should be cfab(i,j,k,qv_comp));  ??

asalmgren:
Sorry -- line 835

rybchuk:
Thanks! I think it should be `dfab`
* `getPgivenRTh` wants `qv` which is water vapor. Line 800 calculates `qv` by dividing out density from `cfab`: `dfab(i,j,k,mf_comp) = cfab(i,j,k,RhoQ1_comp) / cfab(i,j,k,Rho_comp);` . I believe `cfab` stores `rho*qv`
* `LineSampler` uses `dfab` for the same calculation, so the `PlaneSampler` pattern is consistent

asalmgren:
Excellent - you are right -- I should have looked more closely.  Two more things. 1) Suppose we don't have qc - RhoQ2_comp will be -1 -- could you protect against that?  Also -- suppose there is qv in the state but you didn't ask to sample qv -- it seems in this case it would compute the dry pressure?   It seems to me you would still want the qv to impact the pressure independent of whether you sampled it.

rybchuk:
Sure thing
* I added guards against not having `qc` and `qv` in the state
* I refactored the pressure calculations so that moist pressure calculations now either use `qv` from `dfab` if it was already computed, and if not, the moist pressure block now computes `qv`
* I made these updates for both the line sampler and plane sampler

PR #3309, opened 2026-06-16
Protect against multiple specification of same inputs option

nataraj2:
This PR adds a routine to the `ParameterSanityChecks` that will detect if any of the options have been specified mutliple times in the `inputs`, and if yes, it will abort after printing all the occurences. This prevents accidental multiple inputs specifications by the user. An example output is 
```
Duplicate input: amr.v (2 occurrences)
amrex::Abort::1::Duplicate inputs detected !!!
SIGABRT
amrex::Abort::3::Duplicate inputs detected !!!
Duplicate input: erf.substepping_type (2 occurrences)
Duplicate input: geometry.is_periodic (2 occurrences)
Duplicate input: fabarray.mfiter_tile_size (2 occurrences)
Duplicate input: max_step (3 occurrences)
```


asalmgren:
I would not like to merge this PR -- I often keep multiple lines of the same input so I can easily change them without having to re-type the lines.

nataraj2:
Change means to put "#" on one and enable another? That modification can be done on this routine. Have been bit by this issue multiple times especially with long inputs files.

asalmgren:
Ok, that's a good point -- the comment does allow me to do what I need to do and this would protect against the user error you're seeing.   We can try it

nataraj2:
Actually the current routine allows for inputs like below without reporting duplicate errors. So, no modifications needed.
```
max_step = 100000
#max_step = 100000
```

PR #3315, opened 2026-06-17
Alternate moist nudging

asalmgren:
This PR implements a version of moist nudging in which we still nudge qv to the value specified in the bdy files, but we also nudge qc and all other moist quantities to 0.   In addition we calculate a source term for (rho theta) which accounts for the heat which would be used to convert all qc and all qi to qv.


AMLattanzi (commented):
This looks good for testing the two methods. If this proves to be the better approach, I think we should move the for loop out of `ERF_Utils.cpp` and put it in `ERF_AddMoistNudgingTerms.cpp` so we don't have to abort if no ice is present. We could loop each non precipitating component  we have and grab the appropriate latent heat constant.

PR #3334, opened 2026-06-22
Noah Interp Flux From Coarse

AMLattanzi:
# Summary
For lev>0, the land surface model will interpolate the fluxes from level 0. This allows creation of a fine patch from the coarse data without a wrfinput file.

## Challenges

- [x] Noah returns values of -9999 when the landmask is not 1 and these values are are mapped to `lsm_undefined` . The interpolation will pick up these values and smear them in the fine array; this effectively biases our fluxes to large values. To alleviate this issue, the surface layer now writes the iterated flux back into the noah flux data structure when an undefined value is encountered.  
- [x] Interpolate and handle the data arrays for RRTMGP as well.

## Testing

- [x] Tornado with a static region of refinement of `3 3 1` over the `small domain` region.

## Behavior

1.  If a `nc_file` is provided, Noah will parse data from said file. For `lev>0`, if no `nc_file` is present, then interpolate the fluxes from level 0.
2. Force a call to RRTMGP if the LSM was just updated so we are time synced and the radiation module cleans up and lsm_undefined values in the level 0 data.

## Changes
The writing of iterated fluxes into the noah structure will change regression tests because the iterated flux will now be held constant until the next noah call since it is a valid value after the overwrite. In effect, the iterations over invalid land cells will be updated at the same frequency that Noah MP is updated (rather than each step).


asalmgren (approved):

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -239,157 +247,204 @@ NOAHMP::Advance_With_State (const int& lev,
...
+            for (MFIter mfi(*lsm_fab_flux[ivar]); mfi.isValid(); ++mfi) {
+                Box vbx = mfi.validbox();
+                const Array4<      Real>& fine_arr = lsm_fab_flux[ivar]->array(mfi);
+                const Array4<const Real>& crse_arr = mf_crse_patch.const_array(mfi);
+                ParallelFor(vbx, [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept
+                {
+                    int ic = i/refRatio[0];
+                    int jc = j/refRatio[1];
+                    int kc = k/refRatio[2];
+                    if (crse_arr(ic,jc,kc) == lsm_flux_undefined) { fine_arr(i,j,k) = lsm_flux_undefined; }
```
AMLattanzi: @asalmgren directly imposing the `lsm_flux_undefined` on the interpolated fine data, given that the coarse cell  contains  `lsm_flux_undefined`, works for the tornado. However, this approach biases the fluxes in cells neighboring the undefined values since we can get a valid slope in x/y if multiple undefined cells are neighboring each other. I think the only way to truly address this is to either `0` out the crse flux where an undefined value is present or have the surface layer class write its iterated flux into the noah data structure. I think the latter should yield less bias and be the most consistent.

asalmgren (approved):

AMLattanzi:
Joplin Tornado 22 UTC **PR**
image

Joplin Tornado 22 UTC **Develop**
image


Joplin Tornado 22 UTC **OLD RESULT**
image

PR #3343, opened 2026-06-24
base state pressure gradient check for flat terrain only

wiersema1:
The verification of zero horizontal pressure gradients in the base state should only be done for flat terrain since a terrain-following grid with non-flat terrain should have non-zero gradients.

Tested by running Askervein. Was aborting, but runs with this change.

### DIFF
--- a/Source/ERF.cpp
+++ b/Source/ERF.cpp
@@ -1711,9 +1711,10 @@ ERF::InitData_post ()
 #endif
 
     // Print max values of lateral gradients of base state pressure at level 0
+    // Only for flat terrain where horizontal gradients should be zero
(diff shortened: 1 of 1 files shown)


asalmgren:
Actually we specifically want to do the diagnostic for terrain-fitted coordinates!    We only want to abort if the base state is not in HSE.    So we want the print statements there to let us know the size of dp0/dx and dp0/dy.  If you are getting an abort, it is worth paying attention to because it means dp0/dz and rho0*g are not balanced.

wiersema1:
Whoops! Misunderstood what was happening here and I moved too fast.

PR #3367, opened 2026-06-29
Update to MRF 

hgopalan:
This PR provides a comprehensive update of the MRF model 

1. Countergradient term included - Need to use implicit solve 
2.  Support for moist variables 
3. Other capabilities to keep the model close to WRF 
4. A lot of regresstion tests have been added for testing each of the capabilities of the MRF model to enable or disable them. 

The code is almost similar to WRF with flexibility to turn on and off several features mentioned in the documentation. This was done to provide an option to use the best settings for different scenarios.


Source/ERF_IndexDefines.H
```diff
@@ -212,6 +212,8 @@ namespace EddyDiff {
         Scalar_v,
         Q_v,
         Turb_lengthscale,
+        HGAMT_v,  // MRF countergradient for heat (HGAMT/h, units K/m), zero outside PBL
```
AMLattanzi, line 215: Adding these two components will cause a segfault when computing the turbulent viscosities. The factor pointer will be over ran when dereferencing. We need to change the loop to be `<=EddyDiff::Q_h` right  [here](https://github.com/AMLattanzi/ERF/blob/c35a740a15bba3c7313c0575c249d513301a3788/Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp#L739)

Source/TimeIntegration/ERF_ImplicitPre.H
```diff
@@ -65,11 +65,14 @@
 
                     if (l_use_stretched_dz) {
                         if (l_do_implicit_theta) {
+                            const bool l_use_mrf_cg = solverChoice.turbChoice[level].enable_mrf_countergradient
```
AMLattanzi: Declare this bool once before the if conditions for less redundancy.

Source/TimeIntegration/ERF_ImplicitPost.H
```diff
@@ -79,24 +79,33 @@
...
-                                                 mu_turb, solverChoice,
-                                                 bc_ptr_h, l_use_SurfLayer, l_vert_implicit_fac);
-                    } else { // no terrain
-                        ImplicitDiffForStateLU_N(bx, fine_geom.Domain(), level, RhoQ1_comp,
-                                                 slow_dt, l_bc_neumann_vals_d, cell_data,
-                                                 dxInv, q1fx_z, mu_turb, solverChoice,
-                                                 bc_ptr_h, l_use_SurfLayer, l_vert_implicit_fac);
-                    }
+                     if (l_use_stretched_dz) {
+                         const bool l_use_mrf_cg = solverChoice.turbChoice[level].enable_mrf_countergradient
```
AMLattanzi: Same comment for this bool.

AMLattanzi (changes requested):
See comments on the bounds for the turbulent viscosity loop.

hgopalan:
Fixed some bound issues and also updated reg tests which had variables which did not exist.

hgopalan:
[MacOS] ctest runs fine in Debug mode but crashes the moist bubble case in Release mode.

asalmgren (approved):

PR #3404, opened 2026-07-06
Make all time-related variables always double precision

asalmgren:


akashdhruv:
@asalmgren 

Should this be hardcoded then?: https://github.com/erf-model/ERF/blob/3d2b2e8c8bb6100a1870cd17795c132ecf398f64/CMakeLists.txt#L573-L575

asalmgren:
@akashdhruv -- I don't think so -- we're just making "time", "dt" etc always be double precision -- but all of the data itself is still whatever precision is defined (FLOAT vs DOUBLE).    We want NOAHMP data to match the precision of ERF data I would think.

akashdhruv:
Ok. Just wanted to make sure.

akashdhruv:
This may have implications on scalars set by Noah-MP for time and dt. Let me check. I can create a PR if needed after your merge.

PR #3406, opened 2026-07-07
quiet compiler warning

asalmgren:


pressel:
I looked into the single hardware CI failure here. I do not think it is caused by this PR’s integer-cast cleanup.

The failing cases are the sampled-level parser death tests:

Plotfile2DSampledLevel.RejectsUnknownCoordinate
Plotfile2DSampledLevel.RejectsUnsupportedUnits
Plotfile2DSampledLevel.RejectsIsentropicUntilCrossingPolicyExists

The log shows GoogleTest warning that the death tests are using fork() while multiple threads are already running. Since these tests run after amrex::Initialize, the hardware CI appears to be hitting a test-harness issue rather than a production-code regression.

I opened #3408 with a minimal test-only fix: switch only those sampled-level parser death tests to GoogleTest’s threadsafe death-test style. Parser behavior is unchanged.

If #3408 passes CI, it should clear the failure seen here.

PR #3408, opened 2026-07-07
Tests: use threadsafe sampled-level death tests

pressel:
## Summary

- Force sampled-level parser death tests to use GoogleTest's `threadsafe` death-test style.
- Avoid fork-based death tests after AMReX initialization may have created runtime helper threads.
- Leave sampled-level parser behavior unchanged.

## Motivation

Hardware CI can hang in the sampled-level parser death tests after GoogleTest warns that `fork()` is unsafe while multiple threads are running. These tests intentionally exercise parser abort paths after `amrex::Initialize`, so using the threadsafe death-test style is the smallest test-harness-only fix.


pressel:
This is a test-harness-only fix for the hardware CI hang seen in PR #3406.

The failing tests were sampled-level parser death tests. They run after AMReX initialization, and on the hardware CI GoogleTest warned that `fork()` was being used while multiple threads were active. This PR switches only those death tests to GoogleTest's `threadsafe` death-test style.

Production parser behavior is unchanged.

PR #3412, opened 2026-07-08
Noah-MP LSM Coupling Updates

akashdhruv:


Source/ERF_Constants.H
```diff
@@ -17,9 +17,11 @@ constexpr amrex::Real eightthirds  = amrex::Real(8.0/3.0);
...
 // Safe overly large value to fill data so that it is not uninitialized
 //      but we can easily identify if the value is used in a computation
-// Would like to use the following but std::sqrt will not be allowable
-//       in constexpr until C++26
-// const amrex::Real bogus_large_value = std::sqrt(std::numeric_limits<amrex::Real>::max());
+// Must be constexpr (constant-initialized): this value is read inside SYCL
+// device kernels (e.g. ERF_MOSTStress.H, ERF_NOAHMP.cpp), and SYCL forbids
+// using a const global that is only runtime-initialized. std::sqrt is not
+// constexpr until C++26, so use a per-precision literal (float build must not
+// overflow).
```
pressel, line 24: Do we want to keep this comment?

tools/validation/aggregate_event.py
pressel, line 1: These are helpful analysis scripts (this comment applies to this file and the others), and I can see how they’d be useful for validation. I’m just not sure we want to start adding ad hoc analysis workflows to the main ERF repository. This concern applies to the tools/validation directory broadly, not just this script.

Maybe we should consider keeping these outside ERF, moving them to a separate validation/analysis repository, or first defining a clearer policy for what kinds of post-processing and validation tools should live in-tree.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1234,6 +1259,185 @@ SurfaceLayer::compute_pblh (const int& lev,
                      moisture_indices);
 }
 
+void
+SurfaceLayer::update_surf_diagnostics (const int& lev,
+                                       const MultiFab& cons,
+                                       bool do_reflectivity)
```
pressel, line 1265: Reflectivity should not be computed in a surface layer scheme.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1234,6 +1259,185 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+                    const Real dbz = compute_max_reflectivity_dbz(rho, temp, qpr, qps, qpg,
+                                                                  1, 1, 1, 1);
+                    colmax = amrex::max(colmax, dbz);
+                }
+                cref(i,j,0) = amrex::max(cref(i,j,0), colmax);
+            });
+        }
+    }
+}
+
```
pressel, line 1440: I think this reflectivity diagnostic is outside the scope of the surface/LSM coupling work in this PR. It also does not seem like it belongs in the SurfaceLayer code path, since reflectivity is a microphysics/output diagnostic rather than a surface-layer concern.

Could we remove this from this PR and handle it separately, ideally in its own focused PR with the right ownership and tests?

Source/PBL/ERF_ComputeDiffusivityYSU.cpp
```diff
@@ -33,18 +33,23 @@ ComputeDiffusivityYSU (const MultiFab& xvel,
...
-    if (invalid_zref) {
+    // The YSU surface bulk-Richardson reconstruction uses the wind speed and
+    // temperature that MOST provides at its reference height most_zref. WRF
+    // supplies these at a 10 m diagnostic height, but on a vertically-stretched
+    // or terrain-fitted grid MOST snaps zref to the lowest cell center, which
+    // may be well above 10 m (e.g. ~20 m). The calculation is self-consistent at
+    // whatever height MOST actually reports (most_zref, ws10av, t10av are all at
+    // that same height), so we use the actual zref rather than requiring exactly
+    // 10 m. We only guard against a clearly unphysical surface reference height.
+    if (most_zref <= Real(0.0) || most_zref > Real(200.0)) {
```
pressel, line 44: I think the YSU/MRF PBL-height changes should be split out of this PR. They are important, but they are not directly part of the Noah-MP surface/LSM coupling work, and they change the boundary-layer formulation in a way that deserves focused review.

Keeping this here makes the PR harder to review because it mixes land-surface coupling, surface diagnostics, 2D output, and PBL behavior. I think the cleaner path is to remove these changes from this PR and bring them back in a separate, focused PR with targeted tests for the YSU/MRF behavior, especially over terrain and for edge cases in the PBL-height calculation.

Source/PBL/ERF_PBLHeightYSU_MRF.H
```diff
@@ -0,0 +1,334 @@
...
+                while (!above_critical && (kpbl+1) <= khi) {
+                    kpbl += 1;
+                    const Real zval = ERFPBLHUtil::zcc(i,j,kpbl,zphys_arr,have_terrain,
+                                                       dz,problo_z,imin,jmin,imax,jmax);
+                    const Real ws2  = amrex::max(ERFPBLHUtil::ws2_cc(i,j,kpbl,uvel,vvel),
+                                                 Real(1.0e-8));
+                    const Real theta= cons_arr(i,j,kpbl,RhoTheta_comp)
+                                    / cons_arr(i,j,kpbl,Rho_comp);
+                    Rib_dn = Rib_up;
+                    Rib_up = (theta-base_theta)/base_theta * CONST_GRAV * zval / ws2;
```
pressel, line 178: I think this YSU/MRF PBL-height change should be split out and checked carefully. Here `zval` appears to come from `z_phys_cc` when terrain is present, which gives absolute physical height. The existing YSU implementation uses `Compute_Zrel_AtCellCenter()` over terrain, i.e., height above the local surface, before using the height in the bulk Richardson number.

Source/ERF_Constants.H
akashdhruv: No. Going to remove it.

tools/validation/aggregate_event.py
akashdhruv: Yes removing the additions to tools.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1234,6 +1259,185 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+            const Real psih_2    = sfuns.calc_psi_h(zeta(z_temp));
+            const Real psih_zref = sfuns.calc_psi_h(zeta(zref));
+            const Real G2   = std::log(z_temp / z0) - psih_2;
+            const Real Gref = std::log(zref   / z0) - psih_zref;
+            const Real frac = (std::abs(Gref) > Real(1.0e-12)) ?
+                              amrex::min(amrex::max(G2 / Gref, Real(0.0)), Real(1.0)) : Real(1.0);
+
+            t2A(i,j,k) = tsrf(i,j,k) + (Tz(i,j,k)  - tsrf(i,j,k)) * frac;
+            q2A(i,j,k) = qsrf(i,j,k) + (Qvz(i,j,k) - qsrf(i,j,k)) * frac;
+
```
pressel, line 1390: I think we need to clarify the thermodynamic variable here before exposing this as t2. This calculation appears to stay in potential-temperature space: the function comment describes a theta(z) profile, T_ma is documented as theta at zref, and the MOST average field is sourced from Theta_prim.

If the public output is intended to be true 2 m temperature, I think this needs an explicit conversion from diagnosed 2 m potential temperature to temperature using the appropriate pressure/Exner factor. Otherwise, this should probably be renamed/documented as a 2 m potential-temperature diagnostic rather than t2.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1234,6 +1259,185 @@ SurfaceLayer::compute_pblh (const int& lev,
                      moisture_indices);
 }
 
+void
+SurfaceLayer::update_surf_diagnostics (const int& lev,
+                                       const MultiFab& cons,
```
pressel, line 1264: I think the same source-selection logic used for the surface fluxes needs to be carried through to these derived diagnostics. For the fluxes, we use Noah-MP/LSM values where they are available and valid over land, and fall back to MOST where the LSM does not provide a valid value, such as over water or sentinel-filled cells.

As written, these diagnostics appear to be computed everywhere from the MOST surface-layer reconstruction. That means t2/q2 do not use Noah-MP’s native land diagnostics where available, and u10/wspd10/wspd10max also do not explicitly preserve the LSM-vs-MOST source decision. Could we either plumb the relevant Noah-MP diagnostic fields into ERF and merge them with MOST fallbacks using the same validity/source logic as the fluxes, or split these diagnostics out until that source-selection behavior is implemented and tested?

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1234,6 +1259,185 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+            //   G(z) = ln(z/z0) - psi_h(z/L)
+            const Real psih_2    = sfuns.calc_psi_h(zeta(z_temp));
+            const Real psih_zref = sfuns.calc_psi_h(zeta(zref));
+            const Real G2   = std::log(z_temp / z0) - psih_2;
+            const Real Gref = std::log(zref   / z0) - psih_zref;
+            const Real frac = (std::abs(Gref) > Real(1.0e-12)) ?
+                              amrex::min(amrex::max(G2 / Gref, Real(0.0)), Real(1.0)) : Real(1.0);
+
+            t2A(i,j,k) = tsrf(i,j,k) + (Tz(i,j,k)  - tsrf(i,j,k)) * frac;
+            q2A(i,j,k) = qsrf(i,j,k) + (Qvz(i,j,k) - qsrf(i,j,k)) * frac;
```
pressel, line 1389: This looks like a MOST-derived diagnostic everywhere. For coupled Noah-MP runs, I think we need the same land/water source selection used by the fluxes: use the Noah-MP 2 m diagnostic where the LSM provides a valid land value, and fall back to the MOST-derived value over water or other cells where the LSM value is unavailable/invalid.

Without that merge, t2/q2 are not really coupled LSM diagnostics; they are surface-layer reconstructions applied globally.

pressel (changes requested):
Thanks for all this work and getting the PR opened so quickly, very much appreciate it!

PR #3414, opened 2026-07-08
Fully Coupled ERF-Noah-MP Interface

akashdhruv:
# Fully Coupled ERF–Noah-MP Interface

This PR completes the two-way coupling between ERF and Noah-MP (submodule bumped to the coupling-v521 commit). On the input side, ERF now feeds Noah-MP the full WRF-style precipitation breakdown — `RAINBL` plus `SR`/`MP_RAINNC`/`MP_SNOW`/`MP_GRAUP` derived from the rain/snow/graupel accumulators — so the `opt_snf=4` rain/snow partition matches WRF exactly, with device-side compute and host stage-out to the newly-coupled NoahmpIO members. Surface fluxes are returned to the dynamics in kinematic form (`t_flux=HFX/(rho*Cp)`, `q_flux=LH/(rho*L)`, `tau=TAU/rho`), deliberately omitting the Exner factor to match WRF's convention. On the output side, the 2D land return-term diagnostics (`GRDFLX`, `FIRA`, `SAV`, `SAG`, `ALBEDO`, `SFCRUNOFF`, `UDRUNOFF`, `SMSTAV`, `SMSTOT`) are exposed through `NoahmpOutputComp`/`LsmData_NOAHMP` so they appear automatically in the LSM plotfile and checkpoint; these are output-only and are not fed back into the dynamics, matching how WRF carries them.

The PR also generalizes the per-layer soil diagnostics (`SMOIS`/`SH2O`/`TSLB`) to an arbitrary Noah-MP soil-layer count, replacing the previously hardcoded 4-layer enum members and fixed-size arrays with runtime-sized vectors driven by `erf.lsm_nsoil` (default 4). `nsoil` is resolved on the first `Lsm_Data_Size()` call so the soil fabs are correctly sized before checkpoint creation — closing a regression where they were silently dropped from the checkpoint — and `Init` asserts the namelist `NSOIL` matches to fail on misconfiguration.


Source/TimeIntegration/ERF_AdvanceLSM.cpp
```diff
@@ -11,8 +11,17 @@ void ERF::advance_lsm (int lev,
...
     if (solverChoice.lsm_type != LandSurfaceType::None) {
         if (solverChoice.lsm_type == LandSurfaceType::NOAHMP) {
+            // Cumulative surface precip accumulations (mm) from microphysics, used to
+            // force Noah-MP (RAINBL). qmoist ordering is {rain_accum[, snow_accum,
+            // graup_accum]} (MicVarMap); schemes without snow/graup (e.g. Kessler)
+            // expose only rain_accum -> pass nullptr for the missing ones.
+            const int nqm = (lev < int(qmoist.size())) ? int(qmoist[lev].size()) : 0;
+            const MultiFab* rain_accum  = (nqm > 0) ? qmoist[lev][0] : nullptr;
+            const MultiFab* snow_accum  = (nqm > 1) ? qmoist[lev][1] : nullptr;
+            const MultiFab* graup_accum = (nqm > 2) ? qmoist[lev][2] : nullptr;
```
pressel: I think this should use the typed SurfacePrecipAccumulationSources interface rather than raw positional qmoist pointers. ERF already has that interface because the microphysics schemes do not all expose the same precipitation-accumulator semantics or units.

Noah-MP needs interval precipitation in water-equivalent millimeters for RAINBL, MP_RAINNC, MP_SNOW, and MP_GRAUP, plus a frozen fraction. That should be derived from the typed source slots and their native_to_kg_m2 conversion factors, not from the assumption that qmoist[0], qmoist[1], and qmoist[2] are always total rain, snow, and graupel accumulations in the correct units.

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -382,12 +518,39 @@ NOAHMP::Advance_With_State (const int& lev,
...
+                    // exceed a sane ceiling. Protects Noah-MP's water-balance check from
+                    // any corrupt prev-snapshot delta (belt-and-suspenders with the
+                    // zero-init above). 500 mm/interval is far above any real 10-min rate.
+                    drain = amrex::min(drain, Real(500.0));
+                    // Frozen sub-components. rain_accum(=precrt) is the TOTAL surface precip
+                    // and snow_accum(=snowprt: ice+snow), graup_accum(=grplprt) are SUBSETS,
+                    // exactly the WRF MP_RAINNC / MP_SNOW / MP_GRAUP convention
+                    // (drivers/wrf/NoahmpWRFmainMod.F90:563-568).
+                    if (has_snow)  { dsnow  = amrex::max(zero, snow_now (i,j,kklo) - snow_prv (i,j,kklo)); }
+                    if (has_graup) { dgraup = amrex::max(zero, graup_now(i,j,kklo) - graup_prv(i,j,kklo)); }
```
pressel: I think the unit conversion needs to happen before this differencing and before computing SR. These deltas need to be in common water-equivalent units before they are passed to Noah-MP as RAINBL, MP_RAINNC, MP_SNOW, and MP_GRAUP.

That matters because the microphysics accumulators are not all equivalent raw millimeter fields. For example, some schemes expose a total precipitation accumulator, while others expose species accumulations that need density conversion to water equivalent. Without using the typed source interface and its conversion factors, drain, dsnow, dgraup, and therefore SR = (dsnow + dgraup) / drain, can be dimensionally or semantically wrong.

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -323,6 +366,51 @@ NOAHMP::Advance_With_State (const int& lev,
...
+            if (m_rain_accum_prev[lev] == nullptr) {
+                // NOTE: allocate with ZERO ghost cells and setVal(0) first so no cell is
+                // ever uninitialized (an uninitialized prev-snapshot produced garbage
+                // ~1e25 deltas -> Noah-MP water-balance abort). We only ever read the
+                // valid region (surface slab), so 0 ghost cells is sufficient and avoids
+                // any ghost/nGrow mismatch in the Copy.
+                m_rain_accum_prev[lev] = std::make_unique<MultiFab>(
+                    rain_accum_in->boxArray(), rain_accum_in->DistributionMap(), 1, 0);
+                m_rain_accum_prev[lev]->setVal(0.0);
+                MultiFab::Copy(*m_rain_accum_prev[lev], *rain_accum_in, 0, 0, 1, 0);
```
pressel: I think this still needs restart-continuity handling. Seeding the previous-accumulation snapshot to the current accumulation avoids a restart spike, but it can also drop precipitation that accumulated between the last Noah-MP land call and the restart point.

For restart-continuous Noah-MP forcing, I think we need to checkpoint and restore these previous cumulative-precipitation snapshots, or otherwise reconstruct the accumulator values from the last Noah-MP call. This should also have a chained restart test where precipitation occurs between land-model calls.

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -382,12 +518,39 @@ NOAHMP::Advance_With_State (const int& lev,
                 noah_input_arr(i,j,0,NoahmpInputComp::swdown)  = SWDOWN(i,j,0);
                 noah_input_arr(i,j,0,NoahmpInputComp::glw)     = GLW(i,j,0);
                 noah_input_arr(i,j,0,NoahmpInputComp::coszen)  = COSZEN(i,j,0);
+
+                // RAINBL = accumulated precip [mm] this land-call interval (WRF convention;
+                // Noah-MP divides by DTBL -> mm/s). rain_accum is on the surface slab kklo.
+                Real drain = zero, sr = zero, dsnow = zero, dgraup = zero;
+                if (hp) {
+                    drain = amrex::max(zero, rain_now(i,j,kklo) - rain_prv(i,j,kklo));
```
pressel: I think the unit conversion needs to happen before these deltas are computed and before SR is formed. The values passed to Noah-MP as RAINBL, MP_RAINNC, MP_SNOW, and MP_GRAUP should all be in common water-equivalent millimeters.

This matters because some schemes expose total precipitation plus frozen subsets, while others expose explicit species accumulations that need density conversion to water equivalent. Without using the typed source interface and its conversion factors, drain, dsnow, dgraup, and therefore SR = (dsnow + dgraup) / drain, can be dimensionally or semantically wrong.

I would also avoid silently capping drain here as the primary protection against bad deltas. If the previous snapshot or unit conversion is wrong, I think we should fail loudly or at least warn with enough information to diagnose the source, rather than silently changing the coupled water input.

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -382,12 +518,39 @@ NOAHMP::Advance_With_State (const int& lev,
...
+                    // Physical guard: precip accumulated over one land interval cannot
+                    // exceed a sane ceiling. Protects Noah-MP's water-balance check from
+                    // any corrupt prev-snapshot delta (belt-and-suspenders with the
+                    // zero-init above). 500 mm/interval is far above any real 10-min rate.
+                    drain = amrex::min(drain, Real(500.0));
```
pressel: Is this clipping actually needed, if so why? 500.0 seem arbitrary

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -457,15 +671,56 @@ NOAHMP::Advance_With_State (const int& lev,
...
+            // synchronize before it is destroyed at the end of this MFIter body.
+            Gpu::streamSynchronize();
+        }
+
+        // Advance the precip "previous cumulative accum" snapshot to the current values
+        // now that this land call's RAINBL/SR have been consumed, so the next call
+        // differences against this step. Device-safe whole-MultiFab copy (these live in
+        // the device arena; never touch them on the host).
+        if (have_precip) {
+            MultiFab::Copy(*m_rain_accum_prev[lev], *rain_accum_in, 0, 0, 1, 0);
```
pressel: I think this still needs restart-continuity handling. Seeding the previous-accumulation snapshot to the current accumulation avoids a restart spike, but it can also drop precipitation that accumulated between the last Noah-MP land call and the restart point.

For restart-continuous Noah-MP forcing, these previous cumulative-precipitation snapshots need to be checkpointed and restored, or otherwise reconstructed from the accumulator values at the last Noah-MP call. I think this should also have a chained restart test where precipitation occurs between land-model calls.

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.H
```diff
@@ -38,7 +39,27 @@ namespace LsmData_NOAHMP {
...
+         // fed back into ERF dynamics; matches how WRF carries these). Auto-written
+         // to the LSM plotfile + checkpoint via Plot_Lsm_Data / lsm_data checkpoint.
+         grdflx,          // ground/snow heat flux [W m-2]
+         fira,            // total net longwave radiation [W m-2] (FIRAXY)
+         sav,             // solar radiation absorbed by vegetation [W m-2] (SAVXY)
+         sag,             // solar radiation absorbed by ground [W m-2] (SAGXY)
+         albedo,          // total grid surface albedo [-] (ALBEDO)
+         sfcrunoff,       // accumulated surface runoff [m] (SFCRUNOFF)
+         udrunoff,        // accumulated sub-surface runoff [m] (UDRUNOFF)
+         smstav,          // soil moisture availability [-] (SMSTAV; not computed by core -> 0)
```
pressel: Minor comment: if `SMSTAV` and `SMSTOT` are not computed by the Noah-MP core in this driver, I would prefer not to expose them as valid zero-valued diagnostics. A zero value looks physically meaningful in output.

Could we either leave these as `lsm_undefined`, omit them until they are computed, or document very clearly in the output metadata that these fields are currently placeholders rather than physical diagnostics?

pressel (changes requested):
Thanks for cleaning this up and narrowing the scope relative to the earlier PR. This is much easier to review now, and I think the overall direction is right: keep this focused on the ERF–Noah-MP coupling, return Noah-MP fluxes through the existing SurfaceLayer/LSM pathway, and leave the unrelated diagnostics and PBL changes out.

I do think there are a few coupling issues we should resolve before merging:

1. The precipitation forcing into Noah-MP should use the typed `SurfacePrecipAccumulationSources` interface rather than raw positional `qmoist` pointers. ERF already has that interface because the microphysics schemes do not all expose the same accumulator semantics or units. Noah-MP needs interval precipitation in water-equivalent millimeters, so the coupling should use the source slots and conversion factors before building `RAINBL`, `MP_RAINNC`, `MP_SNOW`, `MP_GRAUP`, and `SR`.

2. The previous-accumulation snapshots used to compute Noah-MP interval precipitation need restart-continuity handling. Seeding them to the current accumulator avoids a restart spike, but it can also drop precipitation that accumulated between the last Noah-MP land call and the restart point. I think these snapshots need to be checkpointed/restored, or otherwise reconstructed, and covered by a chained restart test.

3. I think the silent precipitation cap should be reconsidered. If the previous snapshot or unit conversion is wrong, silently capping the interval precipitation can hide the error and break the water budget. I would prefer a diagnostic, warning, or abort path rather than silently changing the coupled forcing.

4. The kinematic flux conversions look broadly consistent with the existing SurfaceLayer contract. `HFX/(rho*Cp)`, `LH/(rho*L)`, and `TAU/rho` produce the expected kinematic quantities for the LSM flux arrays. The only thing I would soften is the Exner comment: omitting Exner matches the WRF convention, but it is still an approximation relative to ERF’s exact potential-temperature flux, and the error can be larger than 1% over lower-pressure/elevated terrain cases.

akashdhruv:
Comments are addressed.

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -323,6 +364,58 @@ NOAHMP::Advance_With_State (const int& lev,
...
+                if (slot_has[s] && prev[s] == nullptr) {
+                    // NOTE: allocate with ZERO ghost cells and setVal(0) first so no cell is
+                    // ever uninitialized (an uninitialized prev-snapshot produced garbage
+                    // ~1e25 deltas -> Noah-MP water-balance abort). We only ever read the
+                    // valid region (surface slab), so 0 ghost cells is sufficient and avoids
+                    // any ghost/nGrow mismatch in the Copy.
+                    prev[s] = std::make_unique<MultiFab>(
+                        slot_accum[s]->boxArray(), slot_accum[s]->DistributionMap(), 1, 0);
+                    prev[s]->setVal(0.0);
+                    MultiFab::Copy(*prev[s], *slot_accum[s], 0, 0, 1, 0);
```
pressel: Thanks for switching this to the typed precipitation-source interface. That addresses my source/unit concern.

I think the remaining issue is restart continuity. This previous-accumulation snapshot still appears to be runtime-only state owned by the NOAHMP wrapper. On restart, it will be lazily initialized to the current microphysics accumulator, which avoids a spike but can still drop precipitation that accumulated between the last Noah-MP land call and the checkpoint.

For restart-continuous coupling, I think these previous-consumed precipitation snapshots need to be checkpointed and restored, or otherwise reconstructed from checkpointed state. Since they answer “what cumulative precipitation did Noah-MP last consume?”, I think they belong in the Noah-MP coupling layer, not in SurfacePrecipAccumulationSource.

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -398,6 +567,21 @@ NOAHMP::Advance_With_State (const int& lev,
...
+                    clamped_cells.push_back(ClampedPrecipCell{i, j, drain_h});
+                    drain_h = lsm_max_precip_interval;
+                }
+                const Real dfroz_h = dsnow_h + dgraup_h;
+                noahmpio->RAINBL(i,j)    = drain_h;                                       // [mm]
+                noahmpio->SR(i,j)        = (drain_h > zero) ? amrex::min(Real(1.0), dfroz_h/drain_h) : zero; // [-]
+                noahmpio->MP_RAINNC(i,j) = drain_h;                                       // [mm]
+                noahmpio->MP_SNOW(i,j)   = dsnow_h;                                       // [mm]
+                noahmpio->MP_GRAUP(i,j)  = dgraup_h;                                      // [mm]
             });
```
pressel, line 617: This is better than a silent cap because it reports clamped cells. I still think we should handle the full precipitation invariant here.

Right now only drain_h is clamped. If dsnow_h + dgraup_h > drain_h, then SR is capped at 1, but MP_SNOW + MP_GRAUP can still exceed MP_RAINNC. That gives Noah-MP an inconsistent precipitation breakdown.

Could we add an invariant check after the water-equivalent conversion, before assigning the Noah-MP inputs? Something like 0 <= MP_SNOW + MP_GRAUP <= MP_RAINNC + tol. If the invariant fails, I would prefer a warning/abort or a consistent rescaling of the frozen components, rather than only clipping the total.

Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.cpp
```diff
@@ -382,12 +527,36 @@ NOAHMP::Advance_With_State (const int& lev,
...
+                Real drain = zero, dsnow = zero, dgraup = zero;
+                if (hp) {
+                    Real dd[NoahmpPrecipSlot::NumSlots];
+                    for (int s(0); s < NoahmpPrecipSlot::NumSlots; ++s) {
+                        dd[s] = slot_present[s]
+                              ? amrex::max(zero, (accum_now[s](i,j,kklo) - accum_prv[s](i,j,kklo)) * accum_fac[s])
+                              : zero;
+                    }
+                    // Frozen -> MP_SNOW / MP_GRAUP (hail folded into graupel; no hail input).
+                    dsnow  = dd[NoahmpPrecipSlot::snow];
```
pressel, line 557: Minor question: should hail be folded into MP_GRAUP, or should we add an explicit MP_HAIL path for completeness? The generic ERF precipitation-source interface has a hail slot, and the WRF Noah-MP bridge carries MP_HAIL separately when available.

If no current ERF microphysics scheme supplies hail, folding it into graupel may be fine for now. I think it would still be worth documenting that this is an intentional fallback rather than an exact WRF parity path for hail-capable schemes.

pressel (changes requested):
Thanks for the quick updates here. This is in much better shape now. I appreciate the move to the typed `SurfacePrecipAccumulationSources` interface; that addresses the main source/unit concern and makes the precipitation coupling much more robust across microphysics schemes. The comments around the Noah-MP forcing path and the kinematic flux conversions are also much clearer now.

I think there are still a couple of items worth resolving before merge. The main one is restart continuity for the previous precipitation-accumulation snapshots. Those snapshots represent the cumulative precipitation Noah-MP last consumed, so they are part of the Noah-MP coupling state. If they are reinitialized from the current accumulator on restart, we can avoid a spike but still drop precipitation that accumulated between the last land-model call and the checkpoint. I think these should be checkpointed/restored, or otherwise reconstructed from checkpointed state, and covered by a chained restart test.

The other issue is the precipitation guard. It is good that clamped cells are now reported, but I think we should also enforce consistency between the total and frozen components. If `MP_SNOW + MP_GRAUP` exceeds `MP_RAINNC`, then `SR` may be capped at 1 while the component inputs remain inconsistent. I think we should either fail loudly or rescale consistently after checking the invariant.

The hail handling is a smaller question. Since the generic ERF interface has a hail slot and WRF carries `MP_HAIL` separately, folding hail into graupel should probably be documented as an intentional fallback unless we add a separate `MP_HAIL` path.

Overall, this is moving in the right direction. I think the typed-source update fixed the biggest issue; I would just like to see the restart-state and precipitation-invariant pieces tightened before this lands.

pressel (approved):
Thanks for all of the work on this @akashdhruv and @demetresfytanides . The PR looks good to merge to me.

PR #3435, opened 2026-07-13
This Make.ERF seems to build RRTMGP without extra steps

asalmgren:


AMLattanzi (approved):
This looks correct to me. I didn't inspect the netcdf stuff in detail though. @jmsexton03 could you take a look at find netcdf stuff and make recommendations?

PR #3456, opened 2026-07-15
CPM vertical extent flexibility and vertical velocity option

adamwise95:
This PR contains two updates to inflow perturbations. The first is that the user can now specify the vertical extent of the perturbations using `erf.perturbation_klo` and `erf.perturbation_khi`. These correspond to the k indices for the lower and upper extent of the perturbation region. I have modified the CPM example file in `Exec/RegTests/TurbulentInflow/` for how someone would use these inputs. 

For the example inputs, we have a stable boundary layer with an inversion located at z = 200m. I have modified the inputs so that the perturbations only extend up to the inversion height:

image

If the user does not specify the bounds, the default behavior is that the perturbations extend the entire column as was the default previously.

image

The second update is an additional pathway where we perturb the vertical velocity instead of theta for CPM. The user can simply specify the range of the vertical velocity perturbations using `erf.perturbation_w_amp` when specifying `erf.perturbation_type = "CPM_W"`. An example input file demonstrating this pathway has also been added. Here is how the vertical velocity perturbations look at initialization:

image

The motivation for this pathway is that CPM breaks down for scales other than the atmosphere (i.e., wind tunnel scales). 

Documentation has been updated relating to both of these changes.

### DIFF
--- a/Docs/sphinx_doc/InflowTurbulenceGeneration.rst
+++ b/Docs/sphinx_doc/InflowTurbulenceGeneration.rst
@@ -273,10 +273,10 @@ Below is an example of the input tags necessary for a simulation with CPM:
 Best practices are to set :math:`{Nx}_{pb}=8`, :math:`{Ny}_{pb}=8`, and
 :math:`{Nz}_{pb}=3` regardless of the physical size of the boxes/cells and to set
 ``erf.perturbation_layers = 3``. The user can also specify the vertical extent of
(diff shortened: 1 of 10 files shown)


asalmgren:
@adamwise95 -- this looks good except for two things:  1) I see _zlo instead of _klo in the docs?  2) Could we default perturbation_khi to be the top of the domain instead of 0?

PR #3478, opened 2026-07-21
[IF]: wall modeling capability for buildings pathway

adamwise95:
This PR adds the wall model to immersed forcing when using buildings. The wall model functions similarly as the existing one for terrain but logic is needed to apply it for horizontal walls. Immersed forcing for buildings is recommended only for fine-scale flows (5.0 m or less) but will still work up to grid spacings of 50.0 m. This logic is added in the initial if statement so that immersed forcing cam ne used in a multi scale setup where one might have mesoscale levels and then nest down to microscale levels. The buildings and `terrain_IB_mask` will be defined at all levels but the forcing is only applied at the finer levels.

Documentation has been added that describes the equations used for the wall model with different formulations for interior cells compared to partially immersed cells (i.e., walls). Example files leveraging the wall model have also been added.

To demonstrate the wall model, here are some short, 300 s time-averaged profiles midway (both vertically and horizontally) through a cube (these results are from the input files that have been added). 

image

Here, we are demonstrating two extreme ends of the spectrum for the wall model. The wall roughness length of 2.49 m is close to half the grid spacing, which numerically results in a similar behavior to a no-slip boundary condition. The wall roughness length of 0.01 m results in a boundary condition closer to a slip wall. The previous figure is for the fully compressible solver and the following figure is for the anelastic solver with similar behavior in both.

Screenshot 2026-07-21 at 4 18 05 PM

For the building wall thermal boundary conditions, the following figure is an example where we set the buildings walls to be a temperature of 305 K.

image

A couple of other small things. I've added the 'box' geometry that previously only worked with EBs to also work with immersed forcing. So the user can just simply specify a box as opposed to having to read in a file.

The other thing is that I have explicitly set the velocities to 0 and do not update the density and temperature for fully immersed cells at the end of `ERF_SlowRhsPost.cpp`. I'm open to alternative suggestions if there is a better way to do this. But the reason I did this is that I found it harder to force the flow closer to 0 when using the anelastic solver compared to the fully compressible solver. This is because the tilmestep at which the forcing term is applied is generally much coarser for anelastic and specifying a similarly strong drag coefficient for anelastic compared to fully compressible can result in numerical instability. Note that explicitly setting the velocities to 0 does not work for the fully compressible solver and results in a crash.

This is one approach that I like and the results are very good for the cube case. I also converted the forcing term to an acceleration like tendency and added some numerical damping to reduce crashes related to immersed forcing when using the anelastic solver. @rybchuk has also been exploring applying the forcing term in an implicit formulation.


adamwise95:
I'd also like to add @hgopalan as a reviewer for this PR but I can't figure out how to do it...

hgopalan:
One thought. We have too many if loops. Can we convert it such that we calculate N, S, E , W, Bottom, Top for all cells and then multiply by the multi-fabs marking it? https://github.com/hgopalan/amr-wind/blob/EB/src/equation_systems/icns/source_terms/PartialDragForcing.cpp.

PR #3480, opened 2026-07-22
PBL Terrain Correction

AMLattanzi:
# Summary
The vertical derivatives for `QKE` sources were not correct for vertically stretched grids. Furthermore, the calls inside the PBL models were also not formally correct. Structs that compute the `dz_inv` for each grid type in the interior and for a Dirichlet wall are added and utilized throughout.

# Tests
@asalmgren please try this on your ocean-atmosphere case and post any results here.


hklion:
I'm getting an arithmetic error in the first ERF advance step when running the coupled ocean-atmosphere case.

(Edited to add, I haven't had a chance to try w/debug to get a backtrace)

AMLattanzi:
> I'm getting an arithmetic error in the first ERF advance step when running the coupled ocean-atmosphere case.
> 
> (Edited to add, I haven't had a chance to try w/debug to get a backtrace)

Thanks Hannah. A copy-paste error was corrected. I also refactored everything for cleanliness and general book keeping.

hklion:
I ran this at [b2001b4](https://github.com/erf-model/ERF/pull/3480/commits/b2001b476177964184f5ed73f5a466220635b3cc), and it seems to be doing the right thing. We're getting a boundary layer.

PR #3500, opened 2026-07-31
Use Native SHOC cloud fraction with RRTMGP

wang1202:
Summary
-------
Use Native SHOC's diagnosed liquid-cloud fraction in the RRTMGP radiation calculation when Native SHOC and RRTMGP are enabled together.

This preserves the legacy binary cloud-fraction behavior for non-Native-SHOC configurations and provides a runtime switch for controlled comparison runs.

Implementation
--------------
- Added erf.rad_use_shoc_cldfrac, defaulting to true.
- Added an optional liquid-cloud-fraction field to the radiation interface.
- Reordered Native SHOC + RRTMGP radiation so SHOC runs first and radiation receives the current-step diagnostic and updated atmospheric state.
- Added separate liquid, ice, and total cloud fractions:
  - liquid fraction uses Native SHOC when available and enabled;
  - ice fraction remains binary;
  - total fraction is max(liquid fraction, ice fraction).
- Liquid water path now uses the liquid fraction, while ice water path uses the binary ice fraction.
- Added validation for the optional cloud-fraction MultiFab layout.
- Added focused unit tests for fraction selection, bounds, and ice-only clouds.
- Updated the input documentation and Native SHOC README.

Scope and limitations
---------------------
- Cloud ice retains a binary cloud fraction because Native SHOC currently diagnoses liquid-cloud macrophysics only.
- Snow, graupel, effective-radius, cloud-overlap, and EAMxx SHOC coupling were not changed.


AMLattanzi (approved):

wang1202:
The impact of this modification isn't very significant, but it affects a lot of files. I've closed this PR for now and will resubmit when I see more of a need.

PR #3510, opened 2026-08-03
Fix EB momentum advection divergence signs

WeiqunZhang:
Negate the complete flux-divergence sum in the regular EB branches for all three momentum components. This corrects the signs of the y- and z-direction advection contributions.


skang67:
This fix is correct. Thank you for checking.

PR #3511, opened 2026-08-03
Fix high-side Open BC momentum indexing

WeiqunZhang:
Offset ivm2 instead of ivm1 at high open boundaries. This restores the correct momentum-gradient sign for tangential advection tendencies.

### DIFF
--- a/Source/Advection/ERF_AdvectionSrcForOpenBC.cpp
+++ b/Source/Advection/ERF_AdvectionSrcForOpenBC.cpp
@@ -256,7 +256,7 @@ AdvectionSrcForOpenBC_Tangent (const int& i,
     int sgn = 1; if (do_lo) { sgn = -1; }
 
     IntVect ivm1(i,j,k); if ( do_lo) { ivm1[dir] -= sgn; } // Mom indexed into domain for do_lo
-    IntVect ivm2(i,j,k); if (!do_lo) { ivm1[dir] += sgn; } // Mom indexed out  domain for do_hi
+    IntVect ivm2(i,j,k); if (!do_lo) { ivm2[dir] += sgn; } // Mom indexed out  domain for do_hi
 
     IntVect ivs1(i,j,k); if ( do_lo) { ivs1[dir] -= sgn; } // Scalar indexed into domain for do_hi
     IntVect ivs2(i,j,k); if (!do_lo) { ivs2[dir] -= sgn; } // Scalar indexed into domain for do_lo


AMLattanzi (changes requested):
Using this PR on the `SquallLine_2D` case with `Open` BCs in the x-direction and printing the `imv1/2` cells yields the following:

```
ivm1 ivm2 lo: (1,1,1) (0,1,1)
ivm1 ivm2 hi: (191,1,1) (192,1,1)
``` 

This shows that the PR flips the sign of the gradient computation for the high side since it is computed as:

```
Real mom_grad  = ( mom_norm_arr(ivm1) - mom_norm_arr(ivm2) ) * dxInv;
```

This line has been flagged before, so a comment was added at the top of the routine (below). The take away is that the tangent advection operator loops over a BA that does not include the cell at the boundary. Since `ivm1` is the high side of the gradient, it should live at the boundary cell when looping xhi.

```
// NOTE: Indices (i,j,k) correspond to data that is index 1/2 dx off open bdy.
//       Therefore, momentum indexing (ivm1/2) have 1 extra cell on the high
//       side that should be accessed while scalar indexing (ivs1/2) does not.
```

Finally, the squall line simulation fails at step 104 with the PR while it runs well past that if the current change is reverted.

WeiqunZhang:
Comments from Claude

If the goal was readability                                                                                                                                              
                                                                                                                                                                           
Both branches reduce to the same thing, so the confusing conditionals can just be deleted with zero behavior change:                                                     
                                                                                                                                                                           
  IntVect ivm1(i,j,k); ivm1[dir] += 1; // Mom on the high-index face of cell (i,j,k)                                                                                       
  IntVect ivm2(i,j,k);                 // Mom on the low-index  face of cell (i,j,k)                                                                                       
                                                                                                                                                                           
That makes mom_at_cc obviously the cell-centered momentum and mom_grad obviously its exact cell-centered derivative, and removes the assign-to-ivm1-on-the-ivm2-line trap for good.                                                                                                                                                                
                                                                                                                                                                           
Recommendation: close, or replace with the no-op simplification above.                                                                                                   
                                                                                                                                                                           
One pre-existing item, unrelated to this PR but adjacent: the φ multiplying mom_grad uses prim_tang_arr(ivs1), which is cell i at the high boundary but cell i+1 at the low boundary — so the two sides don't multiply by the same cell's value relative to (i,j,k). Also the trailing comments on ivs1/ivs2 have their do_lo/do_hi labels swapped relative to their conditions. Worth a separate look.

PR #3512, opened 2026-08-03
Fix vertical ghost handling in boundary conditions

WeiqunZhang:
Clamp lateral theta-profile lookups to valid vertical indices to prevent out-of-bounds reads. Fill all base-state components in the first lower ghost layer so qv0 is extrapolated consistently.

### DIFF
--- a/Source/BoundaryConditions/ERF_BoundaryConditionsBaseState.cpp
+++ b/Source/BoundaryConditions/ERF_BoundaryConditionsBaseState.cpp
@@ -269,20 +269,12 @@ void ERFPhysBCFunct_base::impose_vertical_basestate_bcs (const Array4<Real>& des
     const auto& dom_lo = lbound(domain);
     const auto& dom_hi = ubound(domain);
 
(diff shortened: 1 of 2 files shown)


Source/BoundaryConditions/ERF_BoundaryConditionsBaseState.cpp
```diff
@@ -271,12 +259,9 @@ void ERFPhysBCFunct_base::impose_vertical_basestate_bcs (const Array4<Real>& des
 
     Box bx_zlo1(bx); bx_zlo1.setBig(2,dom_lo.z-1); if (bx_zlo1.ok()) bx_zlo1.setSmall(2,dom_lo.z-1);
     ParallelFor(
-        bx_zlo1, [=] AMREX_GPU_DEVICE (int i, int j, int k)
+        bx_zlo1, ncomp, [=] AMREX_GPU_DEVICE (int i, int j, int k, int n)
```
AMLattanzi: The physical BCs for the base state appear to always be called for all the components. I do not see a reason to split this low side kernel, which loops over 1 vertical ghost layer, from the kernel below that operates on the second vertical layer. The high side does not do this. I would recommend deleting this kernel and modifying the one below for the low side such that it loops over both vertical ghost cells and copies from domlo (just like the high side)

AMLattanzi (approved):

PR #3514, opened 2026-08-03
Fix MOST averaging initialization across terrain and AMR levels

WeiqunZhang:
Honor user-provided MOST reference heights for embedded-boundary terrain, using the first cell center only when no reference height is specified.

Preserve subgrid velocity corrections for previously initialized AMR levels, and populate terrain interpolation positions through the configured averaging radius to prevent radius-two stencils from reading uninitialized ghost data.


Source/BoundaryConditions/ERF_MOSTAverage.cpp
```diff
@@ -580,7 +580,7 @@ MOSTAverage::set_z_positions_EB (const int& lev)
     ParmParse pp(m_pp_prefix);
     auto read_z = pp.query("most.zref",zref_tmp);
 
-    if (!read_z) {
```
skang67, line 583: I agree this fix!

PR #3527, opened 2026-08-03
Cloud Chamber initialization, wall fluxes, and budgets - Stage 1

pressel:
## Summary

This PR adds Stage 1 Cloud Chamber support for a closed, single-level,
Cartesian anelastic domain.

It provides:

- physical-temperature initialization for dry and SatAdj configurations;
- resolved thermal and moisture transfer on all six chamber walls;
- interval-local thermal and water conservation budgets;
- SatAdj thermodynamic diagnostics for anelastic runs;
- focused unit and regression coverage;
- complete user-facing Sphinx documentation.

The existing `legacy_theta_qv` initialization path remains supported and now
has a positive startup regression.

This is a proof-of-concept code path for initialization, wall-flux,
thermodynamic, and conservation testing. It is not yet a quantitatively
validated Pi-Chamber LES.

## Configuration and initialization

The new problem is selected with:

```text
erf.prob_name = "Cloud Chamber"
```

Two physical Stage 1 modes are supported.

### Dry thermal chamber

- omit `erf.moisture_model`;
- omit `prob.initial_relative_humidity`;
- initialize a prescribed physical-temperature profile;
- evolve without `qv` or `qc` state fields.

### Moist equilibrium chamber

- set `erf.moisture_model = SatAdj`;
- provide `prob.initial_relative_humidity` as a fraction from 0 to 1;
- initialize vapor from temperature, local HSE pressure, and RH;
- initialize `qc = 0` before SatAdj establishes vapor-cloud-water
  equilibrium.

The initializer uses a deterministic sinusoidal perturbation, so the initial
condition does not depend on MPI decomposition, tiling, or random-number
state.

The parser rejects ambiguous combinations of physical and legacy profile
keys. It also provides explicit diagnostics for missing SatAdj RH, RH supplied
in dry mode, unsupported moisture models, invalid wall configurations, and
unsupported geometry.

## Stage 1 supported scope

Physical-mode runs require:

- one Cartesian AMR level with no refinement;
- `ConstantDz`;
- nonperiodic geometry in all three directions;
- fixed-density anelastic dynamics;
- gravity;
- explicit vertical diffusion;
- `ConstantAlpha`;
- six stationary `NoSlipWall` faces;
- a prescribed temperature on every wall;
- no terrain, embedded boundaries, or immersed buildings.

The single level may still be decomposed into multiple AMReX boxes for
parallel execution.

## Resolved wall transfer

Each physical wall selects:

```text
<face>.temperature = <temperature in K>
<face>.moisture = dry|wet
<face>.wall_transfer_model = resolved_molecular
```

The resolved wall treatment:

- converts prescribed wall temperature to potential temperature using the
  adjacent-cell HSE pressure;
- computes wall-normal transfer across the near-wall half-cell distance;
- uses the configured `alpha_T` and `alpha_C` coefficients;
- sets dry-wall vapor flux exactly to zero;
- sets cloud-water wall flux exactly to zero;
- permits wet-wall vapor evaporation or condensation;
- obtains the wet-wall saturation state from wall temperature and the
  adjacent-cell HSE pressure;
- applies the same corrected face flux to both the state RHS and budget
  diagnostics.

A wall marked `dry` is impermeable to water but can still exchange heat.

The configured transfer coefficients are prescribed model inputs. They are
not calibrated engineering wall-law coefficients or automatically selected
molecular properties.

## SatAdj and diagnostics

The microphysics interface now accepts optional anelastic base-state context.

SatAdj uses that context to diagnose temperature and pressure consistently
from the anelastic conserved state. Other microphysics implementations retain
their existing behavior through the default interface path.

Plotfile selection and diagnostics support the chamber workflow, including
temperature, saturation, and relative-humidity checks for SatAdj cases.

## Conserved-scalar budgets

Setting:

```text
erf.cloud_chamber_budget_interval = <positive step interval>
```

writes interval-local rows to:

```text
cloud_chamber_budget.dat
```

The budget reports:

- `rhoTheta`;
- water vapor;
- cloud water;
- total nonprecipitating water;
- all six physical-face contributions;
- the net boundary contribution;
- the state change;
- the internal microphysics source;
- the residual;
- the derived numerical tolerance;
- `PASS`, `FAIL`, or `UNSUPPORTED_SOURCE`.

The face values are coordinate-oriented. The net boundary contribution uses
low-face minus high-face flux for each coordinate direction.

Strict integrated thermal closure is enforced in the dry chamber. Moist
conservation is assessed using total nonprecipitating water, for which the
SatAdj vapor/cloud-water exchange cancels.

`UNSUPPORTED_SOURCE` is limited to cloudy `rhoTheta` rows whose current budget
does not fully represent latent heating. It must not be interpreted as a
successful closure result.

A parity regression verifies that enabling budget output does not alter the
simulated state.

## Validation

The added validation covers:

- physical-temperature and RH initialization;
- deterministic initial perturbations;
- all-six-face thermal and moisture wall behavior;
- multi-box physical-face ownership;
- dry thermal-budget closure;
- all-dry total-water conservation;
- wet-wall water exchange and closure;
- budget-enabled versus budget-disabled solution parity;
- OpenMP parity;
- strict invalid-configuration diagnostics;
- retained `legacy_theta_qv` startup compatibility;
- SatAdj anelastic temperature and pressure diagnostics;
- warning-clean compilation of the added source.

Relevant registered tests include:

```text
AnelasticWallDiffusion_X
AnelasticWallDiffusion_Y
AnelasticWallDiffusion_Z
CloudChamber_Dry
CloudChamber_Dry_ThermalBudget
CloudChamber_Legacy_Config
CloudChamber_SatAdj
CloudChamber_SatAdj_Parity
CloudChamber_SatAdj_AllDry
CloudChamber_SatAdj_WetBudget
CloudChamber_SatAdj_OpenMP
CloudChamber_Invalid_DryPhysicalRH
CloudChamber_Invalid_SatAdjPhysicalMissingRH
CloudChamber_Invalid_PhysicalLegacyThetaPerturbation
CloudChamber_Invalid_LegacyPhysicalTemperature
```

The thermal regression requires at least three finite `rhoTheta` intervals,
all with `PASS` status and residuals below their printed tolerances. Its
negative control also proves that a missing budget file fails the checker.

The Linux compiler-warning gate reports:

```text
Warnings: 0
```

`MoistBubble`, `SHOC_Stable_Cloud`, and `ScalarAdvectionUniformU` pass on the
Linux CI configuration used for this branch.

## Documentation

The new Sphinx page documents:

- how to choose dry or SatAdj mode;
- required solver and geometry settings;
- complete runnable regression inputs;
- initialization equations and notation;
- wall heat and moisture behavior;
- flux orientation;
- budget interpretation;
- Stage 1 invariants;
- a run-and-check workflow;
- scientific limitations and troubleshooting.


AMLattanzi:
@pressel this PR broke the `TurbulentInflow_anelastic` nightly [regtests](https://ccse.lbl.gov/pub/RegressionTesting1/ERF/). If this is unexpected, please look into the failure. If it was expected, please comment why and I'll remake the benchmark.

pressel:
@AMLattanzi investigating now

pressel:
This regression-test change appears to be expected and results from correcting the anelastic temperature diagnostic.

For anelastic simulations, temperature is now computed as:

```cpp
const Real theta = rhoTheta / rho;
const Real temp  = getTgivenPandTh(p0, theta, rdOcp);
```

This evaluates:

```math
T = \theta \left(\frac{p_0}{P_{00}}\right)^{R_d/c_p}.
```

ERF’s anelastic formulation evolves `rho0 * theta` while using the fixed hydrostatic reference density `rho0`. Therefore,

```math
\theta = \frac{\rho_0\theta}{\rho_0},
```

and temperature should be reconstructed using the hydrostatic base-state pressure `p0`.

Previously, the diagnostic used:

```cpp
getTgivenRandRTh(rho, rhoTheta)
```

That function reconstructs pressure from `rhoTheta` using the compressible equation of state. This is appropriate for the compressible model, but it is inconsistent with the anelastic model, where `rho0` is fixed and changes in `rho0 * theta` should not imply a new compressible pressure.

This change affects the regression test because ERF regression tests compare the variables written to the plotfile using AMReX `fcompare`. The plotfile contains both prognostic fields and derived diagnostics. For `TurbulentInflow_anelastic`, `temp` is one of the output variables, so the diagnosed temperature is stored in the plotfile and compared directly against the existing benchmark.

As a result, the regression test can fail even when the underlying anelastic solution is unchanged. If `density`, `theta`, and the velocity components agree with the benchmark, but `temp` differs, then the failure is caused by the corrected output diagnostic rather than a change in the prognostic model state.

In that case, the expected action is to regenerate the benchmark. The tolerance should not be loosened, since this is an intentional change in the diagnostic definition rather than numerical noise.

The expected comparison signature is:

> `temp` changes, while the underlying anelastic state and velocity fields remain unchanged.

PR #3534, opened 2026-08-04
MYNN2.5 Clean Up

AMLattanzi:
# Summary
1. MYNN2.5 operates on a growntilebox when it does not need to. The ghost cells are populated after the eddyviscosity is filled for valid cells with LES and PBL models. Additionally, it uses a tilebox when the tiling is explicitly false.
2. Change tile flag to `TileNoZ()` since we need the whole column for an integral and operate on the valid region only. The grown box leads to extra checks, factors for reduction ops on the GPU, and unnecessary operations that can be removed.
3. Modify the level 2 functions to limit directly with `eps` for `GM` instead of `GM + eps`. This change impacts problems that start with a constant velocity and thus have no shear production.
4. Modify the level 2 Rf function to use a conjugate formulation to minimize roundoff error
5. Modify the level 2 remaining functions to avoid a divide by 0 error
6. Explicitly use the `async_arena` since FABs are defined inside an MFIter. This prevents the destructor from being called before the GPU is done.


AMLattanzi:
This PR **may** modify regtests that use MYNN2.5

This PR **did not** modify any ctests:
```
100% tests passed, 0 tests failed out of 70
```

PR #3542, opened 2026-08-04
Fix top metric in terrain implicit state diffusion

WeiqunZhang:
Use the lower-face metric for the top-row sub-diagonal coefficient so the two rows sharing that face use a consistent diffusive flux.


AMLattanzi (approved):

AMLattanzi:
# Confirmation
1. The deletion of the metric lo at the bottom face is correct since it is not used. 
2. The a coeff in the tridiagonal system should use the lo metric since it is at the lower face (this is clearly a copy-paste error).

# Workflow Current
This PR will modify the terrain regtests
```
MetGrid
WPS_Test_Terrain
```

This PR did not break any ctests
```
100% tests passed, 0 tests failed out of 71
```

# Workflow New
Added a test for terrain `ABL_MOST_IMP_DIFF_WOA`

PR #3486, opened 2026-07-23
PBLH Corrections

hgopalan:
MRF and YSU depend on planetary boundary layer height to compute the turbulent viscosity. The variable becomes prognostic instead of diagnostic. So we have to modify the calculation and filling up the surface layer. Few other changes: 

1. Fixed the low wind speed limits for MRF and YSU. A geostrophic wind speed of 0.1 m/s will be still stable.
2. Added PBLH smoothing between neighbouring cells and improved PBLH diagnosis for unstable ABL 
3. Added QNSE stability condition for very stable ABL for MRF/YSUNew. This prevents the collapse of the very stable ABL under low geostrophic winds and high cooling rate.

PR #3543, opened 2026-08-04
Fix terrain YSU countergradient scaling

WeiqunZhang:
Apply the missing vertical grid factor to terrain-fitted momentum countergradient terms and document the dormant YSUNew dispatch gap.


hgopalan:
There is a discrepancy between the actual terms in the YSU paper and how the code is implemented using WRF. The term is written as d/dz ( - nut gammac) (https://www2.mmm.ucar.edu/wrf/users/physics/phys_refs/PBL/YSU.pdf). We calculate gammac ~ surface flux / pblh. We tested it for very low geostrophic wind speeds and @pressel confirmed that the results are as expected. 

The Fact term already contains a dz_inv and the extra dz_inv would dilute the term with an extra dz and it should not be included.

asalmgren:
@hgopalan -- Could you add a comment in the code to that effect so that next time we run a check like this it doesn't identify this as an issue again?     Thx!

hgopalan:
https://github.com/erf-model/ERF/pull/3486 should fix the comments section.

Source/Diffusion/ERF_ImplicitDiff_T.cpp
```diff
@@ -419,7 +419,7 @@ ImplicitDiffForMomLU_T (const Box& bx,
               if (use_ysu_mom_countergradient && stagdir < 2) {
                   const int hgam_comp = (stagdir == 0) ? EddyDiff::HGAMU_v : EddyDiff::HGAMV_v;
                   const Real gam_hi = myhalf * (mu_turb(i,j,klo,hgam_comp) + mu_turb(i,j,klo+1,hgam_comp));
-                  RHS_a(i,j,klo) += Fact * gfac * rhoAlpha_hi * gam_hi / met_h_zeta_hi;
+                  RHS_a(i,j,klo) += Fact * gfac * dz_inv * rhoAlpha_hi * gam_hi / met_h_zeta_hi;
```
AMLattanzi, line 422: @hgopalan Shouldn't the implementations for YSU and MRF counter gradient follow the same paradigm? It appears that this differs from L137 in [PR3544](https://github.com/erf-model/ERF/pull/3544/changes). Can you comment on any assumptions about the bottom flux being `zero` so we can rectify these two PRs?

hgopalan: The countergradient term is d/dz(K gammac). The gammac term is written as w'theta' / pblh. So we have only one dz_inv we should use. This one has extra and it will dilute the counter gradient further. Line 137 is correct. I was not sure what the met_h_zeta... term was and assumed it was a curvature correction.

AMLattanzi: I agree. The metric term now looks incorrect as well.

AMLattanzi (changes requested):
Again considering the counter gradient term:
Image

There is no second derivative, so adding a dz_inv is not correct. Finally, the `met_h_zeta` also looks incorrect for the same reason it should be removed from [PR 3544](https://github.com/erf-model/ERF/pull/3544) --- i.e., the `detJ` from the divergence has been moved to the RHS.

AMLattanzi:
This PR is incorrect and being closed. A new PR addressing #3773 should be completed alongside merging of #3544 .

PR #3544, opened 2026-08-04
Fix MRF countergradient scaling over terrain

WeiqunZhang:
Treat rho-alpha gamma as a physical vertical flux in the implicit terrain diffusion RHS. Avoid applying an extra vertical metric factor to the countergradient flux at boundary and interior faces.


hgopalan:
I have not run enough cases on terrain to understand the effects of dropping the term. We can add a comment in there on why it was removed, in case we need to revert it back.

AMLattanzi:
@hgopalan The formulation for the diffusion source follows `F_hi - F_lo / dz`, where `F ` is the diffusive flux. The code corrects the `dz` terms with a metric term that computes `met_zeta = dz / dzeta`, where `dzeta` is what the code calls `dz` (confusing I know). Thefore, we have `dz = met_zeta * dzeta`. The implict solve moves the `met_zeta` on the outer derivative to the RHS. So this PR appears to be treating your term like a flux and **NOT** dividing by `met_zeta`. 

Given the following turbulent flux with a diffusive and counter gradient term:
image
the counter gradient term should in fact **NOT** have the `met_h_zeta` since the `detJ` on the divergence has been moved to the RHS --- i.e., the `met_h_zeta` for the diffusive flux comes from the correction of the vertical gradient of `K \partial_z C`.

AMLattanzi (approved):

AMLattanzi (approved):

PR #3550, opened 2026-08-05
Fill transverse EB scalar flux stencils

WeiqunZhang:
Grow the second-order scalar flux boxes in the directions transverse to each face so EB centroid interpolation reads initialized neighboring fluxes at box and tile boundaries.


skang67:
I checked this modification with WitchOfAgnesi, and it does not break. I confirm the changes.

PR #3553, opened 2026-08-05
EB: Fix directional area-fraction ghost fill

WeiqunZhang:
Use the factory direction when extrapolating the extra high-side area-fraction slice so the v- and w-factory ghosts are filled correctly.


skang67:
This seems a bug and I agree on the fix.

PR #3564, opened 2026-08-06
Support reference pressure in anelastic microphysics

pressel:
## Summary

This PR adds consistent reference-pressure thermodynamics for supported Eulerian microphysics schemes when ERF is run with anelastic dynamics.

On an anelastic AMR level, ERF uses

```math
\theta = \frac{\rho\theta}{\rho},
\qquad
p_{\mathrm{micro}} = p_{\mathrm{base}},
```

and diagnoses temperature as

```math
T
=
\theta
\left(
\frac{p_{\mathrm{base}}}{p_{\mathrm{ref}}}
\right)^{R_d/c_p}.
```

Here, \(p_{\mathrm{base}}\) is ERF's hydrostatic base-state pressure and \(p_{\mathrm{ref}}\) is the fixed reference pressure used in the definition of potential temperature.

The anelastic pressure-temperature conversion uses ERF's configured `erf.c_p`.

Compressible AMR levels retain their existing local equation-of-state behavior.

## Supported Eulerian microphysics

The reference-pressure treatment applies to:

* `Kessler`
* `Kessler_NoRain`
* `SAM`
* `SAM_NoIce`
* `SAM_NoPrecip_NoIce`
* `Morrison`
* `Morrison_NoIce`
* `WSM6`
* `WDM6`

`SatAdj` already uses the hydrostatic base-state pressure in anelastic configurations.

`MoistNoCondensation` does not have a pressure-dependent condensation source requiring this treatment.

## Pressure conventions

ERF handles the scheme-specific pressure units internally:

* Kessler and SAM use hPa/mbar internally.
* Morrison, WSM6, and WDM6 use Pa.

Users do not provide an additional microphysics pressure field or perform these pressure-unit conversions themselves.

## Compressible non-regression

The existing compressible thermodynamic path is intentionally preserved.

In particular, the historical WSM6 and WDM6 temperature-to-potential-temperature conversion behavior is retained in compressible configurations, including when `erf.c_p` differs from the default value.

For anelastic levels, those conversions instead use the configured (R_d/c_p) together with the held hydrostatic base-state pressure.

## SAM

The existing compressible SAM coefficient construction is unchanged:

1. horizontally average density, potential temperature, and water-vapor mixing ratio;
2. diagnose pressure and temperature from those averaged state variables.

The anelastic path instead uses thermodynamic fields diagnosed from the hydrostatic base-state pressure.

## Mixed AMR hierarchies

`erf.anelastic` may be specified per AMR level.

[...]

PR #3565, opened 2026-08-07
CUDA build fixes and nvcc warning cleanup

debog:
Fixes four CUDA build failures and warning sources found on LLNL LC Matrix (nvcc 13.1.1, `ERF_ENABLE_TESTS=ON`), plus the report-gate pattern that hid them. Each failure masked the next, so they surfaced serially after #3548.

- `Tests/CMakeLists.txt`: compile the six test-helper executables via `setup_target_for_cuda_compilation()`. They link `AMReX::amrex` and include the generated `AMReX_Config.H`, so `AMREX_USE_CUDA` is defined and `AMReX_GpuReduce.H` requires `<cub/cub.cuh>`; built as plain CXX they receive no CUDA include path and fail. Latent since #3492 added `Tests/ShocPlotfileCheck.cpp`, and invisible upstream because `ERF_ENABLE_TESTS` defaults to `OFF` and `cuda-ci.yml` does not set it.
- `Tests/Unit/Microphysics/SatAdj/ERF_GTestSatAdjTemperatureDiagnostics.cpp`: move the `ParallelFor` into a free function. nvcc rejects an extended `__device__` lambda whose enclosing function has private class access, which googletest's `TestBody` always has. Introduced by #3527; currently failing `cuda-ci.yml` on `development`.
- `Source/Microphysics/{SAM,Kessler,Morrison,WSM6,SuperDropletsMoist}`: add `using NullMoist::Update_Micro_Vars;`. #3527 added a second `Update_Micro_Vars(MultiFab&, const MultiFab*)` overload to `NullMoist`; classes overriding only the single-argument form hide it, emitting nvcc `#611-D` roughly 700 times per build. Dispatch is unchanged. #3564 covers all of these except `SuperDropletsMoist` by overriding both overloads.
- `Source/TimeIntegration/ERF_SlowRhsPre.cpp`: remove the dead `z_t` (`#550-D`), unused since #3342.
- `.github/workflows/cuda-ci.yml`: the report gate greps `warning:|error:`, which never matches nvcc's `warning #NNN-D:`. Run `31113179047` at `1ecf606ea` logged 818 `#611-D` diagnostics and reported `Warnings: 1`.


pressel:
@debog, thanks for the fixes.

PR #3727, opened 2026-08-09
Fix Morrison constant droplet number input

wang1202:
This PR ensures erf.morrison_ndcnst is used by removing the later hard-coded reset to 250 cm⁻³. The default remains 250 cm⁻³ when the parameter is not specified, and the resolved value is preserved for both the C++ and Fortran Morrison paths.


pressel:
@wang1202 Thanks for getting this done! I have had fixing this on my todo list for a while.

wang1202:
@pressel The option was already there. I just found that changing `erf.morrison_ndcnst` didn't work, and I thought that the hard-coded `m_ndcnst = Real(250.0)` was forgotten to be removed.

asalmgren:
Some of these changes were addressed by other issues and the remaining changes were addressed in https://github.com/erf-model/ERF/pull/3850

PR #3766, opened 2026-08-11
Stretched colum adv

AMLattanzi:
Add test for stretched column advection to exercise metric correction


AMLattanzi:
With the `met_h_zeta` fix
image

Without the `met_h_zeta` fix:
image

Note that the test passes with and without the change implemented.

AMLattanzi:
This test does not work because the substepping algorithm doesn't have a finite update to the lateral momenta.

PR #3770, opened 2026-08-11
[Noah-MP] Initialize default values for variables before initial write.

akashdhruv:


cmkaul:
This PR solved the crash on first write problem I was seeing; tested on multiple cases.

cmkaul (approved):

PR #3779, opened 2026-08-13
Unsync RRTMP & NOAH Single Level

AMLattanzi:
# Summary
Allow LSM and RRTMGP to not time sync when running single level. The time syncing is utilized to fill the LSM data structures in regions where the land model does not operate, this prevents interpolation of uninitialized or invalid data when employing a fine AMR patch. For single level, no interpolation occurs so we do not need to time sync.


akashdhruv (approved):
Makes sense to me.

PR #3804, opened 2026-08-14
[IF] Separate immersed forcing into another file, prepare for face-centered volume fractions

adamwise95:
This PR largely moves the immersed forcing code from ERF_MakeMomSources.cpp and ERF_MakeSource.cpp into its own file ERF_ImmersedForcing.cpp.

Other major change is that IF is now ready for face-centered volume fractions (when `USE_FC_FACTORY=1` becomes the default in ERF_EB.H). There is also plotting support for these nodal volume fractions. For now, we default to the previous behavior where we average cell-centers to the faces. There is also a bug fix here where partially immersed cells (walls) were not forced correctly when using the wall model but only for orthogonal walls. For example, east-west walls were not forced in the x-momentum routine when they should have been. This bug happened when I previously reworked immersed forcing to reduce the number of if statements.

Other small things include zero-ing the eddy diffusivities for fully immersed cells (in ERF_AdvanceDycore.cpp) and using the `eb2.small_volfrac` input parameter to clip small `terrain_blank` values. This parameter automatically clips smaller volume fractions for EBs, but `terrain_blank` is actually an inverted volume fraction compared to EBs, so it needs to be clipped on the opposite end.


adamwise95:
This will also affect both immersed forcing reg tests as I updated it to now include the `eb2.small_volfrac` input parameter.

PR #3871, opened 2026-08-18
[Hurricane] Adding NOAHMP plotting script

nataraj2:
This PR adds a python script to read the netcdf files and plot the surface stresses from the `lnd*` directories written by the NOAHMP land surface model.


asalmgren:
@nataraj2 -- my understanding is that lnd files may be going away -- @akashdhruv can you comment?

akashdhruv:
I am planning to keep it for diagnostics and make it available when compiling Noah-MP in DEBUG mode, so the script can be helpful. However, for production runs 2D plot file should be the default IO pathway.

PR #3880, opened 2026-08-20
Add time-varying surface forcing to MOST

ckendrick:
This adds options to `SurfaceLayer` to interpolate surface temperature and fluxes over time from a user-provided file. For the surface flux pathway, this is equivalent to using `surface_layer.flux_type = "custom"` but using time-varying data from a file instead of constant values.

The new options added are:
- `erf.most.use_sfc_fluxes` boolean flag to enable interpolating surface flux values from file.
- `erf.most.use_sfc_sst` boolean flag to enable interpolating surface temperature (SST) from file.
- `erf.most.sfc_file` path to text file, required if either option above is enabled.

The user-provided text file to use for the `erf.most.sfc_file` option has the following format:
```
day sst(K) H(W/m2) LE(W/m2) TAU(m2/s2)
 205.500   293.65      4.43     22.73      0.00
```
where each new line represents a distinct time value to interpolate between. The day column is converted to seconds and the elapsed simulation time in ERF is used.

For the SST pathway (`erf.most.use_sfc_sst = true`), the input-file format can be simplified to only `day sst(K)` column data, e,g:
```
day sst(K)
202.227083    290.460
```

Some configurations are not currently supported with these options (LSM, EB).


asalmgren:
# Code review: PR #3880 — "Add time-varying surface forcing to MOST"

- URL: https://github.com/erf-model/ERF/pull/3880
- Author: Coleman Kendrick (`ckendrick`), branch `most_sfc_flux_file`
- Size: 22 files changed, +768 / -0
- Scope reviewed: `Source/BoundaryConditions/ERF_SurfaceLayer.{H,cpp}`,
  `Docs/sphinx_doc/SurfaceLayer.rst`, `Tests/*`

## High

### 1. Off-by-one bounds check in `read_cols` → out-of-bounds `std::vector` write
`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:1698`

The guard is `if (j > ncols)`, but `col_data` has exactly `ncols` entries, so
`j == ncols` slips through and `col_data[j].push_back(tmp)` on line 1704 writes
past the end.

Two concrete triggers:

- any data line with more columns than the first data line;
- a blank or comment line immediately after the header — the column-count pass
  then sets `ncols = 0` with `col_data` empty, and the very first real data row
  does `col_data[0].push_back` on an empty vector.

Both are heap corruption / UB rather than a clean error. Should be `j >= ncols`.

### 2. `read_cols` never rejects short lines, producing ragged columns
`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:1696`

A row with fewer tokens than `ncols` — a truncated last line, a missing TAU
value — pushes into only the leading columns. `sfc[0]` then holds more entries
than `sfc[3]`/`sfc[4]`, and `interpolate_sfc_column` indexes
`sfc[col][sfc_time_ind+1]` off the end of the shorter column once the run reaches
that interval: silent garbage or a crash. Needs a `j != ncols` check after each
line.

### 3. Prescribed fluxes are missing the 1/ρ conversion, and the ρ conventions disagree
`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:282`

`t_star = H/Cp_d` has units kg·K/(m²·s), i.e. ρ·w'θ'. With
`flux_type = custom` and `most.rhosurf` unset — exactly the new `ABL_MOST_SFC`
test — `specified_rho_surf == false`, so `custom_flux::compute_t_flux` returns
`rho * tstar`. The imposed `rho*<θ'w'>` is then a further factor of ρ (~1.2) too
large, with the error tracking the local density. The kinematic flux should be
`H/(rho*Cp_d)` — 200 W/m² → 0.166 K·m/s, not 0.199. Line 283 has the same
problem for `q_star = LE/L_v`.

Conversely, if the density-weighted convention was intended, `specified_rho_surf`
must be set true — but then line 281's `u_star = sfc_ustar` is wrong, because
that convention needs `sqrt(rho)*ustar` (cf. lines 213-215). Either way the three
assignments are mutually inconsistent.

### 4. No guard that `flux_type == CUSTOM` when `use_sfc_fluxes` is on
`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:268`

`flux_type` defaults to `moeng`, and `moeng_flux::compute_t_flux` interprets
`t_star` as a temperature *scale* (flux ≈ `-rho*ustar*tstar`), not as a flux. A
user who sets only `erf.most.use_sfc_fluxes = true` gets a heat flux of
`-rho*ustar*H/Cp_d` — wrong magnitude by `rho*ustar`, and wrong sign — with no
warning, after the MOST iteration results were silently discarded by the `setVal`
calls. Add an abort, or force `flux_type = CUSTOM`.


## Medium

### 5. `setVal` applies the ocean fluxes over land as well
`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:281`

Unlike `fill_tsurf_with_sfc_sst`, which respects `m_lmask_lev`, the three
`setVal` calls overwrite every cell, wiping the land values that
`compute_fluxes` / LSM just produced. In a mixed land/sea domain (real/WRF init
with a landmask) the file's ocean H, LE and u\* are imposed over land too. Either
mask the assignment or abort when the landmask contains land.

### 6. The RANS boundary-TKE update runs before the prescribed fluxes are installed
`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:223`

The `m_update_k_rans` block reads `u_star_arr` / `t_star_arr` at lines 223-263,
but `m_use_sfc_fluxes` overwrites them at 281-283. With
`erf.most.use_sfc_fluxes` + `rans_type = kEqn` + `dirichlet_k`, the Dirichlet
`RhoKE` is computed from the pre-override values — which, in the intended
`flux_type = custom` configuration, are the constant `most.ustar`/`most.tstar`
(0.0 in the new test), so surface TKE is pinned near zero instead of being
derived from the prescribed u\*. The `m_use_sfc_fluxes` block must come first.

### 7. Declaration/definition signature mismatch breaks `ERF_PRECISION=SINGLE`
`Source/BoundaryConditions/ERF_SurfaceLayer.H:788`

Declared as `fill_tsurf_with_sfc_sst(const int&, const double&)` but defined at
`ERF_SurfaceLayer.cpp:1283` as `(const int&, const Real&)`. These match only when
`amrex::Real` is `double`; with `ERF_PRECISION=SINGLE` (a supported option,
`CMakeLists.txt:138`) the out-of-class definition matches no declaration and the
build fails. Use `amrex::Real` in the declaration — the sibling
`update_sfc_time_index` / `interpolate_sfc_column` declarations already do.

### 8. Interpolation clamps above but not below, and `sfc_time_ind` only moves forward
`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:321`

`update_sfc_time_index` never decrements, and `interpolate_sfc_column` handles
`elapsed_time > t1` but linearly *extrapolates* when `elapsed_time < t0`. With
AMR subcycling the fine level runs ahead of the coarse level, so the next
coarse-level `update_fluxes` call — at an earlier time — can find
`sfc_time_ind` already advanced and extrapolate backwards outside the bracketing
interval. Add a lower clamp (`elapsed_time < t0 → x0`) and let the index walk
back.

### 9. Column 4 units: header says `TAU(m2/s2)`, code uses it as `u_star` (m/s)
`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:273`

The in-code comments say `USTAR(m/s)`, while the PR body and the shipped
`Tests/test_files/ABL_MOST_SFC/sfc.txt` header both say `TAU(m2/s2)`, i.e. u\*².
If a real dataset supplies TAU = 0.09 m²/s² (u\* = 0.3 m/s), the code sets
`u_star = 0.09` and the surface stress is ~11× too small. The test file's TAU
column is all `0.00`, so the gold files cannot detect this. Pick one convention
and make header, comment, docs and code agree (take the `sqrt` if it is TAU).

### 10. Documented parameter names are missing the `erf.` prefix
`Docs/sphinx_doc/SurfaceLayer.rst:23`

Parsing uses `ParmParse pp("erf")` (`ERF_SurfaceLayer.H:417`), so the real names
are `erf.most.use_sfc_fluxes`, `erf.most.use_sfc_sst`, `erf.most.sfc_file` —
matching the rest of the file (`erf.most.z0`, …). A user copying
`most.use_sfc_fluxes = true` from the new paragraph has it silently ignored by
`pp.query` and runs with no surface forcing and no error.

The new options are also absent from the input-parameter table at lines ~179-192,
and the file format — column order, units, day→seconds conversion, the
`flux_type = custom` requirement — is documented only in the PR description, not
in Sphinx.


### 11. The file's absolute day-of-year is discarded; the table is re-anchored to the run start
`Source/BoundaryConditions/ERF_SurfaceLayer.H:443`

`start_day = sfc[0][0]` shifts times relative to the file's *first row*, so the
day column carries no absolute meaning and `ERF::start_time` is ignored. For a
real case initialized at day 206.0 with a file that begins at day 205.5 — exactly
the shipped `sfc.txt` — the 205.5 forcing is applied at t=0: a 12-hour phase
error in the diurnal cycle, silently. Anchoring to `start_time`, as the
boundary/low-frequency data paths do via `elapsed_time_since_start_low`, would
fix it; at minimum the constraint needs documenting.

## Low

### 12. Neither new regression test exercises the time interpolation
`Tests/test_files/ABL_MOST_SFC/ABL_MOST_SFC.i:4`

`max_step = 10` with `erf.fixed_dt = 1.0` covers 10 s, while the first interval
in `sfc.txt` is 0.042 day = 3628.8 s (66750 s for the SST test). Both runs only
ever see the first table row, so `update_sfc_time_index`, the interval advance,
the end-of-table clamp, and findings 3, 8 and 9 above are all untested by the
gold files. A larger `fixed_dt`, or a file with second-scale spacing, would give
real coverage.

### 13. `use_sfc_fluxes()` / `use_sfc_sst()` accessors are dead code
`Source/BoundaryConditions/ERF_SurfaceLayer.H:1080`

No caller anywhere in `Source/`. Either wire them up — e.g. for the `flux_type`
guard in finding 4 — or drop them.

ckendrick:
@asalmgren

> #### 1. Off-by-one bounds check in `read_cols` → out-of-bounds `std::vector` write
> #### 2. `read_cols` never rejects short lines, producing ragged columns

Fixed in commit [6be7cf2](https://github.com/erf-model/ERF/pull/3880/changes/6be7cf24f1a5fa55e65ac9021e76f6821f2e9bc5).

> #### 3. Prescribed fluxes are missing the 1/ρ conversion, and the ρ conventions disagree
> #### 4. No guard that `flux_type == CUSTOM` when `use_sfc_fluxes` is on
> #### 6. The RANS boundary-TKE update runs before the prescribed fluxes are installed
> #### 7. Declaration/definition signature mismatch breaks `ERF_PRECISION=SINGLE`
> #### 9. Column 4 units: header says `TAU(m2/s2)`, code uses it as `u_star` (m/s)

Fixed in commit [427bfb0](https://github.com/erf-model/ERF/pull/3880/changes/427bfb0f233b4d33375ead2695588b35d5b405d7). The surface fluxes are meant to use the custom flux pathway, so the flux type is now always set to custom. The tau to ustar conversion is fixed.

> #### 5. `setVal` applies the ocean fluxes over land as well
For now, this is intentional (mirroring the flux_type=custom path). We use this to test fluxes over land as well as ocean, but it probably doesn't make too much sense to use it in a setup for both at the moment. We could extend this or add an option to selectively apply for land or ocean. The documentation is updated to mention both land and ocean fluxes are being set.

> #### 8. Interpolation clamps above but not below, and `sfc_time_ind` only moves forward
> #### 10. Documented parameter names are missing the `erf.` prefix
Fixed in commit [4ad21d7](https://github.com/erf-model/ERF/pull/3880/changes/4ad21d77005057f3d7b827beaa33740768caade6).

> #### 11. The file's absolute day-of-year is discarded; the table is re-anchored to the run start
For now this is an issue to address in the future, but this current limitation is added to the documentation. ERF doesn't currently have a standard way to get the calendar day outside of the radiation. The input column in calendar days is also left to be compatible with existing files.

> #### 12. Neither new regression test exercises the time interpolation
The tests are still interpolating fluxes using the file data but are intentionally keep short to keep test time down and to test the fluxes/temperatures coming in to ERF on a larger scale matching a real setup. I can follow up with a smaller unit test to specifically test the time interpolation part isolated on its own.

> #### 13. `use_sfc_fluxes()` / `use_sfc_sst()` accessors are dead code
We do have some other code that uses these functions, but I can still remove these for now if needed.

asalmgren:
@ckendrick -- could you go ahead and address comment 13 (i.e. remove those functions until they're used in code in the repo) then we'll merge ... thx!!

ckendrick:
@asalmgren No problem, they are removed now

PR #3881, opened 2026-08-20
Substep Vertical Pressure Decomposition

AMLattanzi:
### Extensive testing should be completed before merging this PR.
---------------------------------------------------------------------------

# Summary
This PR modifies the coefficients for the tridiagonal solve in the substepping algorithm. The motivation for this change is that it yields a consistent vertical pressure gradient decomposition and it makes the substep solution independent of the base state.

## KSD Formulation
The pressure gradient relation and its relation to potential temperature
image

The linearized approximation
image

**The concern with this formulation is that it formally replaces perturbations from the base state, `\phi^{'}`, with perturbations from the RK stage, `\phi^{''}`. By doing so, the pressure perturbation in the substep is not with respect to the RK stage and the ensuing coefficients depend upon the base state.**

## Proposed Change
Formally decompose the pressure gradient into a slow and fast component. The slow has not changed and is with respect to the base state, consistent with the RK integrator on its own. The fast now considers `\partial_z P^{''}` as a starting point.
image


AMLattanzi:
The arguments provided here have been proposed for MPAS:
```
More directly, the MPAS-A technical note (height-based terrain-following ζ, explicitly
following Klemp et al. 2007) makes precisely this change and gives the same
justification as your PR description: "the coefficient of the Θ''_m/Θ^t_m term in
(3.26) differs slightly from that in equation (14) in Klemp et al. (2007), and represents
a somewhat more consistent derivation for the double-prime perturbation variables.
We also note that in their current form the coefficients of all of the double-prime
terms in (3.25)–(3.28) are exactly the same as in the corresponding acoustic
equations obtained directly from (3.6)–(3.9) using the full variables for p, ρ̃_d, and
Θ̃_m without expressing them as perturbations from a specified reference sounding
profile."
```

asalmgren:
# Code review: PR #3881 — "Substep Vertical Pressure Decomposition"

- URL: https://github.com/erf-model/ERF/pull/3881
- Author: Aaron M. Lattanzi (`AMLattanzi`), branch `SubStepPGF`
- Size: 1 file changed, +28 / -32 (`Source/TimeIntegration/ERF_MakeFastCoeffs.cpp`)
- Also covered: the uncommitted working-tree changes to that file and to
  `Exec/RegTests/WPS_Test/inputs_wps`

## The core math checks out

No bug in the new formulation. Re-deriving it: with `p = R_d π ρθ_m` and
`π ∝ (ρθ_m)^(R_d/c_v)`, the fast pressure-gradient force is

```
-∂_z(γ R_d π Θ'') = -γ R_d π_c ∂_z Θ''  -  γ R_d (R_d/c_v) π (∂_z Θ_m / Θ_m) Θ''_m|face
```

The code's second term,
`-Gamma*R_d*R_d*Thm_grad*myhalf*pi_stage/(c_v*stage_cons(RhoTheta_comp))`,
matches that exactly, including the moisture bookkeeping: the moist `Thm` in
`Thm_grad` divided by the *dry* `ρθ` in the denominator is correct, because it
multiplies the dry `drho_theta` and `(ρθ_m)''/Θ_m == (ρθ)''/(ρθ)`. Units reduce
to m/s²/K, matching the first term.

Other checks:

- In the hydrostatic limit the new term reduces to `+halfg·(R_d/c_v)·ρ/Θ` — the
  same value and the same sign as the removed base-state term (`Thm_grad < 0`),
  so the tridiagonal's diagonal dominance is preserved. The new piece is
  O(g/θ) ≈ 0.03 against O(γR_d·dzi·π) ≈ 4 on the diagonal.
- The `myhalf` on each of `coeff_P` / `coeff_Q` correctly reproduces the face
  average.
- The terrain and `ConstantDz` branches are consistent — only
  `inv_detJ_on_kface` differs, correctly.
- `RvoRd` is a valid `constexpr` (`Source/ERF_Constants.H:56`).
- `detJ_cc` is filled for both `StretchedDz` and `VariableDz`, so the added
  `inv_detJ_on_kface` inside `Thm_grad` is safe.

## Findings

### 1. Medium — the documentation still describes the deleted formulation
`Docs/sphinx_doc/TimeAdvance.rst:115`

The documented vertical-momentum substep equation still carries the old
base-state term `- g ρ̄(R_d/c_v) (π^t/π) Θ''/Θ^t`, which this PR removes. The
implemented term is now `- γ R_d²/c_v · π^t/Θ^t · ∂_z Θ_m^t · Θ''`, with no `ρ`
or `π` dependence — and that base-state independence is the entire point of the
PR, so the docs now describe a formulation the code no longer implements.

Separately, and pre-existing: the signs of the last two terms in that display
(`- g ρ̄… Θ''` and `+ g ρ''`) are opposite to what the code has ever computed —
the code uses `+halfg·R_d·ρ/(c_v·…)` in `coeff_P` and `-halfg·drho` for
buoyancy, which is what the derivation gives. Worth fixing while the block is
being rewritten.

### 2. Low — Doxygen block documents parameters that no longer exist
`Source/TimeIntegration/ERF_MakeFastCoeffs.cpp:20`

The comment block still has `@param[in] r0` and `@param[in] pi0`, but both
parameters are now commented out and unused. ERF builds docs in CI
(`.github/workflows/docs.yml`), and Doxygen warns on documented-but-nonexistent
parameters.

### 3. Low — `r0` / `pi0` are now dead API
`Source/TimeIntegration/ERF_MakeFastCoeffs.cpp:36`

They remain in the declaration
(`Source/TimeIntegration/ERF_TI_fast_headers.H:138-139`) and the caller still
passes them (`Source/TimeIntegration/ERF_TI_substep_fun.H:133`). Nothing
misbehaves, but dropping the arguments would keep a future reader from assuming
the coefficients still depend on the base state.

Also cosmetic: the reformatted signature at lines 37 and 39, and the wrapped
`detJ` initializer at lines 81-82, lose the column alignment used throughout the
file.

### 4. Low — explicit and implicit halves of the vertical fast PGF disagree by O(q) in moist runs
`Source/TimeIntegration/ERF_MakeFastCoeffs.cpp:135`

Pre-existing, but this PR rewrites exactly these coefficients, so it is the
natural place to fix it.

`coeffP_a` / `coeffQ_a` are stored *before* the moisture normalization at lines
141-142, while `coeffA_a` / `coeffB_a` / `coeffC_a` are built from the
`/(one+q)`-divided values. In a moist run the explicit RHS in
`erf_substep_{T,MT,NS}` (e.g. `Source/TimeIntegration/ERF_Substep_T.cpp:564`,
`:570`, `:584`) therefore uses an un-normalized vertical fast PGF while the
implicit tridiagonal LHS uses the normalized one — an inconsistency of order `q`
(~2% in the tropical boundary layer) between the two halves of the same term.
The horizontal fast PGF does divide by `(one+q)`
(`ERF_Substep_T.cpp:312`, `:350`), which makes the vertical direction the odd one
out.

PR #3884, opened 2026-08-20
WRF hydrometeor lateral-boundary advection and nudging

wang1202:
# WRF hydrometeor lateral-boundary advection and nudging

## User-facing configuration

Enable WRF real lateral boundaries and the WRF hydrometeor fields in the
`erf` input section:

```ini
erf.init_type            = "WRFInput"
erf.use_real_bcs         = true
erf.nc_init_file_0       = "wrfinput_d01"
erf.nc_bdy_file          = "wrfbdy_d01"
erf.use_wrf_bdy_qc_qi    = true
erf.bdy_moist_nudge_type = 3
```

`erf.use_wrf_bdy_qc_qi = true` enables WRF hydrometeor boundary data. It
requires WRFInput real boundaries and a `wrfbdy` file containing the relevant
fields. Set `erf.bdy_moist_nudge_type = 3` to directly nudge the active moist
components toward the time-interpolated WRF boundary values. The usual
`erf.real_width` and `erf.bdy_nudge_factor` settings control the boundary-zone
width and nudging strength.

## Hydrometeor behavior

ERF determines the behavior from the active `erf.moisture_model` state layout.
If the moisture scheme has separate `qc`, `qi`, `qr`, `qs`, and `qg`
components, all of those categories are advected and filled at the lateral
boundaries separately. With moisture nudging type 3, they are also nudged
separately using the corresponding WRF fields:

| ERF component | WRF field |
|---|---|
| `qc` | `QCLOUD` |
| `qi` | `QICE` |
| `qr` | `QRAIN` |
| `qs` | `QSNOW` |
| `qg` | `QGRAUP` |

This separate-component path applies to schemes that provide all five
hydrometeor components, such as Morrison, WSM6, SAM, WDM6, and SuperDroplets.

For schemes that do not provide all five components, ERF uses the aggregate
fallback:

- `QCLOUD + QRAIN` supplies `qc`.
- `QICE + QSNOW + QGRAUP` supplies `qi` when an ice component is active.

The precipitating categories that are not active in the selected moisture
scheme are not added as ERF state components.

## Notes

- The WRF graupel variable is named `QGRAUP`.
- `erf.bdy_moist_nudge_type = 3` requires
  `erf.use_wrf_bdy_qc_qi = true`.
- Regenerate any `erfbdy` cache after switching between aggregate and
  separate-component moisture layouts.

## Validation

- CUDA production targets built successfully.
- The focused WRF boundary moisture unit-test source compiled successfully.


asalmgren:
@wang1202 -- I'm wondering if the ice etc are available in wrfbdy and are used in the moisture model, should we force the nudging to always use nudging type 3?     This wouldn't break any existing cases and would make the logic more straightforward...

wang1202:
> @wang1202 -- I'm wondering if the ice etc are available in wrfbdy and are used in the moisture model, should we force the nudging to always use nudging type 3? This wouldn't break any existing cases and would make the logic more straightforward...

I'm not sure whether all sources have those hydrometeors (ice, snow, rain...), but it's in the wrfbdy generated from HRRR. Maybe after more testings to see that the runs are stable we can make this the default case?

In my testings, I did see that the results are improved (although it seems not to be the reason of the missing patch of cloud above Nantucket.)

erf_ctr_bdyhydro_obs_frame_00036_000

PR #3895, opened 2026-08-21
[IF] Only computer planar averages on slow step, logic for performance, and CPU vs GPU fix

adamwise95:
Planar averages are now only computed on the slow step and passed in.

Added logic statements to improve performance for immersed forcing with buildings. In the future, I will make additional masks at initialization to further boost performance.

Added fill boundary calls after zeroing out the eddy diffusivities which should solve the CPU vs GPU discrepancies for the reg test.


asalmgren:
# Code review: PR #3895 — "[IF] Only computer planar averages on slow step, logic for performance, and CPU vs GPU fix"

- URL: https://github.com/erf-model/ERF/pull/3895
- Author: `adamwise95`, branch `IF_planaravg_fix`
- Size: 9 files changed, +200 / -118
- Reviewed at head `e29e89929`

## High

### 1. Null `Table1D` dereference: immersed forcing + `custom_w_subsidence` + moisture
`Source/SourceTerms/ERF_MakeSources.cpp:117`

```cpp
bool compute_averages = (is_slow_step && dptr_wbar_sub && !use_immersed_forcing);
```

The `!use_immersed_forcing` term means that when immersed forcing is on, the
whole `compute_averages` block is skipped — including the moisture branch that
fills `qv_plane_tab` / `qc_plane_tab`. The tables passed in from the caller only
cover rho and RhoTheta (`r_plane_avg` / `t_plane_avg`). The RhoQ1/RhoQ2
subsidence kernel at lines 394-417 then evaluates `dptr_qv_plane(k+1)` /
`dptr_qc_plane(k-1)` on default-constructed `Table1D`s (`p == nullptr`) →
`nullptr[k]`.

Failure scenario: any run with `erf.terrain_type = ImmersedForcing` (or
`BuildingsType::ImmersedForcing`) plus `custom_w_subsidence = true` and a
moisture model — segfault / illegal memory access on the first slow step.
Previously `compute_averages` was `(is_slow_step && dptr_wbar_sub) || ...`, so
the qv/qc tables were always built when subsidence was active.

Related behavior change worth confirming is intentional: even in the dry case,
theta subsidence now uses start-of-timestep averages rather than
current-RK-stage averages.

## Medium

### 2. The `MeshType::VariableDz` guard is now unreachable for immersed-forcing runs
`Source/SourceTerms/ERF_MakeSources.cpp:122`

`AMREX_ALWAYS_ASSERT(solverChoice.mesh_type != MeshType::VariableDz)` sits inside
`if (compute_averages)`. Before this PR immersed forcing set
`compute_averages = true`, so an immersed-forcing run with variable dz aborted
with a clear message. Now `compute_averages` is always false for immersed
forcing, and the replacement computation in
`Source/TimeIntegration/ERF_AdvanceDycore.cpp:134-249` carries no equivalent
assert.

Failure scenario: immersed forcing + `VariableDz` runs to completion using plane
averages taken at fixed k-index rather than fixed height, producing silently
wrong `theta_mean` / `rho_mean` in the immersed forcing source terms instead of
aborting.

### 3. `smn_arr(i,j,k) *= mask` reads `SmnSmn` before it has ever been written
`Source/TimeIntegration/ERF_AdvanceDycore.cpp:629`

`SmnSmn_lev[lev]` is allocated in `Source/ERF_MakeNewArrays.cpp:698` with no
`setVal` (unlike `eddyDiffs_lev`, which is `setVal(zero)` on line 696), and it is
first written by `make_tau_terms`
(`Source/Diffusion/ERF_MakeTauTerms.cpp:291/448/583`, `nrk==0`) — which runs
*later* in `advance_dycore`, inside the integrator. The old code only *assigned*
into `smn_arr`; the new multiply *reads* it.

Failure scenario: immersed forcing + Deardorff LES or kEqn RANS (the exact
conditions under which `SmnSmn_lev` is non-null), first timestep, an AMReX build
with `amrex.init_snan=1` and `amrex.fpe_trap_invalid=1` (the default pairing in
assertion/debug builds): the array is filled with signaling NaNs, `snan * 0.0`
raises `FE_INVALID`, and the run aborts before doing any work.

## Low

### 4. The `SmnSmn` half of this fix cannot have any effect
`Source/TimeIntegration/ERF_AdvanceDycore.cpp:635`

`SmnSmn_lev` is created with 0 ghost cells
(`Source/ERF_MakeNewArrays.cpp:698`), so `SmnSmn->FillBoundary(...)` is an
unconditional no-op. More importantly the masking of `SmnSmn` is dead:
`make_tau_terms` overwrites every valid cell of `SmnSmn` at `nrk==0` later in the
same `advance_dycore` call, and nothing reads `SmnSmn` in between. If the intent
is to keep `SmnSmn` zero inside immersed bodies, the masking has to move
into/after `make_tau_terms`.

### 5. Switching from assignment to multiplication no longer sanitizes non-finite eddy diffusivities
`Source/TimeIntegration/ERF_AdvanceDycore.cpp:626`

`eddy_arr(i,j,k,n) *= mask` leaves `NaN` / `Inf` intact (`NaN * 0 == NaN`),
whereas the previous `eddy_arr(i,j,k,n) = Real(0.0)` scrubbed them.

Failure scenario: a PBL/LES/MOST closure that produces a non-finite `K` inside a
fully blanked cell (e.g. a `log(z/z0)` or divide-by-zero on garbage state inside
a building) used to be neutralized here and now propagates into the diffusion
terms via the subsequent `FillBoundary`. Given the stated goal is
reproducibility, an explicit `if (t_blank == one) { eddy_arr = 0 }` plus the new
`FillBoundary` gets the reproducibility fix without weakening the scrub.

### 6. The `&& r_plane_avg && t_plane_avg` guard has no fallback path
`Source/SourceTerms/ERF_MakeSources.cpp:110`

If a caller ever passes empty tables while `use_immersed_forcing` is true, the
guard skips the assignment but `compute_averages` is still forced false by
`!use_immersed_forcing`, so `dptr_r_plane` / `dptr_t_plane` stay null and are
then dereferenced by `ImmersedForcingTerrain_Scalar` /
`ImmersedForcingBuildings_Scalar` (lines 473, 492). Both current call sites pass
valid tables, so this is latent rather than live — but the guard reads as
protective when it isn't. Either drop the `r_plane_avg && t_plane_avg` test in
favour of an `AMREX_ALWAYS_ASSERT`, or make the guard actually fall back to
computing the averages locally.

### 7. Immersed-forcing runs now pay for the nudging-only plane averages
`Source/TimeIntegration/ERF_AdvanceDycore.cpp:134`

Widening the condition to `if (use_nudging || use_immersed_forcing)` pulls in the
`u_ave` / `v_ave` `PlaneAverage` computations (lines 197-234) — two extra global
reductions plus two device allocations per timestep per level — even when
`nudging_from_input_sounding` is off, and their results (`dptr_u_plane`,
`dptr_v_plane`) are only consumed inside the `use_nudging` block at lines
446-456. This runs contrary to the PR's stated performance goal. The rho/RhoTheta
average is all immersed forcing needs; the u/v (and qv) work should stay behind
`use_nudging`.

### 8. `use_immersed_forcing` shadows the enclosing-scope variable of the same name
`Source/TimeIntegration/ERF_TI_slow_rhs_pre.H:111`, `Source/TimeIntegration/ERF_TI_substep_fun.H:71`

Both headers are `#include`d into the body of `ERF::advance_dycore`
(`ERF_AdvanceDycore.cpp:708-709`), which as of this PR declares
`bool use_immersed_forcing` at line 126. The values are identical so there is no
behavioral bug, but `-Wshadow` is enabled in the HIP CI job
(`.github/workflows/hip.yml:39`) and this will add warnings. Reuse the enclosing
variable instead of re-deriving it.

## Checked and found sound

- The `if (mask == one)` short-circuits added throughout
  `Source/SourceTerms/ERF_ImmersedForcing.cpp` are behavior-preserving: every
  mask at lines 412-418 / 575-581 / 742-748 is assigned exactly `one` or `zero`,
  and the guarded blocks only write locals that are unused when the mask is zero
  (the `l_implicit_drag` `lambda` at 476/639/818 re-derives from the masks). They
  also remove some previously-unconditional out-of-range `k-1` reads.
- Table sizing/indexing is consistent: `init_stuff` sizes
  `[domain.smallEnd(2)-1, domain.bigEnd(2)+1]` and `advance_dycore` writes
  `k-offset` for `k ∈ [0, ncell)` with `offset = ng_c[2] = 1`, matching the
  existing `make_sources` convention; the tables are resized under exactly the
  same condition that enables immersed forcing, and restart goes through
  `MakeNewLevelFromScratch → init_stuff`.
- `SmnSmn` and `eddyDiffs` share the same `BoxArray` / `DistributionMapping`, so
  cross-indexing with the `eddyDiffs` `MFIter` is valid; `Table1D` / `Array4`
  both provide `explicit operator bool` so the truthiness tests compile and mean
  "non-null".
- Both `make_sources` call sites were updated to the new signature.

adamwise95:
Modified the PR to address the comments above. For 2, planar averaging with variable dz may not be acceptable for subsidence, but doing planar averages at k levels (which is what we are doing in ERF_AdvanceDycore.cpp) is sufficient for immersed forcing. 

For 3, zero-ing SmnSmn was not a primary goal of this PR, zero-ing the eddy-diffusivities is. I just removed zero-ing SmnSmn zero-ing from this PR to address that comment.

Source/SourceTerms/ERF_MakeSources.cpp
```diff
@@ -100,15 +102,19 @@ void make_sources (int level,
...
-    // Compute averages on slow step for subsidence, or whenever immersed forcing runs
-    // NOTE: We recompute planar averages every substep when immersed forcing uses substeps.
-    //       This is simpler than passing persistent storage through the call chain, and
-    //       the cost is relatively small compared to the full solve.
-    bool compute_averages = ( (is_slow_step && dptr_wbar_sub) ||
-                              (use_immersed_forcing && ((is_slow_step && !use_ImmersedForcing_fast) ||
-                                                        (!is_slow_step && use_ImmersedForcing_fast))) );
+
+    // Use passed-in planar averages for immersed forcing (computed once in ERF_AdvanceDycore.cpp)
+    if (use_immersed_forcing && r_plane_avg && t_plane_avg) {
```
asalmgren, line 112: **Medium — the subsidence recompute silently discards the passed-in averages.**

These lines set `dptr_r_plane`/`dptr_t_plane` from the persistent `r_plane_avg`/`t_plane_avg`, but `if (compute_averages)` below then unconditionally reassigns both at lines 164-165 to the stage-local `r_plane_tab.table()`/`t_plane_tab.table()`.

Consequence: with `erf.custom_w_subsidence = true`, an immersed-forcing run gets a per-RK-stage average recomputed from `S_data`, while the same run with subsidence off gets the frozen start-of-timestep average from `state_old`. So the immersed-cell relaxation target depends on whether subsidence is enabled, and the PR's "compute once per slow step" goal is defeated for exactly that configuration.

Either use the passed-in tables for the immersed-forcing term regardless of `compute_averages`, or keep two separate table pairs.

Source/TimeIntegration/ERF_AdvanceDycore.cpp
```diff
@@ -595,34 +614,23 @@ void ERF::advance_dycore (int level,
...
-
-                    ParallelFor(bx, [=] AMREX_GPU_DEVICE (int i, int j, int k) {
-                        if (t_blank_arr(i,j,k) == Real(1.0)) {
-                            smn_arr(i,j,k) = Real(0.0);
-                        }
-                    });
-                }
-            }
+            // Synchronize ghost cells after modifying interior cells
+            eddyDiffs->FillBoundary(fine_geom.periodicity());
```
asalmgren: **Medium — `FillBoundary` cannot reach the ghost cells that matter.**

The zeroing loop above covers `mfi.tilebox()` (valid cells only), and `FillBoundary` only repairs ghosts backed by another box's valid region or a periodic image. Ghost cells outside a non-periodic physical boundary, and fine-level ghosts at a coarse/fine interface, keep whatever `ComputeTurbulentViscosity` put there — non-zero inside a building.

Failure scenario: a building touching the x-lo inflow boundary leaves `eddyDiffs(-1,j,k)` non-zero while `eddyDiffs(0,j,k)` is zeroed, so the diffusive flux on that face still mixes into the immersed cell. Same for a building straddling a level boundary under AMR.

`terrain_blanking[lev]` is filled *through* its ghost region — `MultiFab::Subtract` against the EB volume fraction uses `z_phys_nd[lev]->nGrowVect()` (`ERF_MakeNewArrays.cpp:964`) — so iterating `mfi.growntilebox(eddyDiffs->nGrowVect())` would zero all of them directly and make this `FillBoundary` unnecessary. Drop `TileNoZ()` on that loop so grown tiles don't overlap.

To be clear: the CPU-vs-GPU reproducibility goal *is* met as written, since domain-exterior ghosts are decomposition-independent. The gap above is a physics gap, not a regression of the stated fix.

Source/SourceTerms/ERF_MakeSources.cpp
```diff
@@ -100,15 +102,19 @@ void make_sources (int level,
...
+        // Use pre-computed values from persistent storage (already in Table1D format)
+        dptr_r_plane = r_plane_avg;
+        dptr_t_plane = t_plane_avg;
+    }
+
+    // Compute planar averages if needed for subsidence
+    bool compute_averages = (is_slow_step && dptr_wbar_sub);
```
asalmgren, line 119: **Low — the `MeshType::VariableDz` guard is now unreachable for immersed forcing.**

`AMREX_ALWAYS_ASSERT(solverChoice.mesh_type != MeshType::VariableDz)` a few lines below sits inside `if (compute_averages)`, which no longer includes the immersed-forcing clauses. The replacement computation in `ERF_AdvanceDycore.cpp:132-249` carries no equivalent assert, and there is no other IF/VariableDz check in the code base.

This combination is reachable: `mesh_type` becomes `VariableDz` when `terrain_type` is `StaticFittedMesh` or `MovingFittedMesh` (`ERF_DataStruct.H:503-506`), which composes freely with `buildings_type = ImmersedForcing`. Fitted terrain plus immersed-forcing buildings used to abort with a clear message and now runs to completion using plane averages taken at fixed k-index rather than fixed height.

Source/TimeIntegration/ERF_AdvanceDycore.cpp
```diff
@@ -119,14 +119,17 @@ void ERF::advance_dycore (int level,
...
+    bool use_immersed_forcing = (solverChoice.terrain_type == TerrainType::ImmersedForcing ||
+                                  solverChoice.buildings_type == BuildingsType::ImmersedForcing);
+
     Table1D<Real> dptr_r_plane, dptr_t_plane, dptr_qv_plane;
     TableData<Real, 1> r_plane_tab, t_plane_tab,  qv_plane_tab;
 
     Table1D<Real> dptr_u_plane, dptr_v_plane;
     TableData<Real, 1> u_plane_tab, v_plane_tab;
-    if (use_nudging)
+    if (use_nudging || use_immersed_forcing)
```
asalmgren, line 132: **Low (efficiency) — this widened condition pulls in an unused `qv` plane average.**

The `qv_plane_tab` block below is now entered for immersed-forcing runs, but `dptr_qv_plane` is read only by the nudging kernel at line 412, which is gated on `use_nudging`. So an immersed-forcing run with any moisture model pays for a `line_average` + host-to-device copy + kernel every timestep per level for nothing.

It belongs inside the `if (use_nudging)` block alongside the u/v averages you already moved there.

Source/TimeIntegration/ERF_AdvanceDycore.cpp
```diff
@@ -185,48 +188,64 @@ void ERF::advance_dycore (int level,
...
+        }
+
+        // Store planar averages in persistent ERF member variables for immersed forcing
+        if (use_immersed_forcing) {
+            Table1D<Real> r_avg_persistent = r_plane_avg[level].table();
+            Table1D<Real> t_avg_persistent = t_plane_avg[level].table();
+
+            // Copy from local computation to persistent storage
+            // Both use the same Table1D indexing with offset, so direct copy
+            ParallelFor(ncell, [=] AMREX_GPU_DEVICE (int k) noexcept
```
asalmgren, line 254: **Low (efficiency) — redundant allocation plus kernel launch.**

When `use_nudging` is false, `r_plane_tab`/`t_plane_tab` exist only to be copied here. The earlier loop (line 168) could write `r_avg_persistent(k-offset)` directly, with `dptr_r_plane = r_plane_avg[level].table()`, saving one arena allocation pair and one kernel per level per timestep.

Source/ERF.H
```diff
@@ -1006,6 +1006,10 @@ private:
     amrex::Vector<amrex::Gpu::DeviceVector<amrex::Real>> zvel_bc_data;
     amrex::Vector<amrex::Gpu::DeviceVector<amrex::Real>> th_bc_data;
 
+    // Planar averages for immersed forcing (computed on slow step, used in slow and fast steps)
+    amrex::Vector<amrex::TableData<amrex::Real, 1>> r_plane_avg;   // density planar average per level
```
asalmgren, line 1011: **Low — `TableData` used without including its header.**

`amrex::TableData<amrex::Real, 1>` is new to `ERF.H` in this PR (it appears nowhere in `ERF.H` on `development`), but no `#include <AMReX_TableData.H>` was added here — it compiles only through the transitive include in `ERF_EulerianMicrophysics.H`. You correctly added the include to `ERF_SrcHeaders.H`; worth adding to `ERF.H` too so this doesn't break if the microphysics include chain is reorganized.

Source/SourceTerms/ERF_ImmersedForcing.cpp
```diff
@@ -1163,7 +1195,7 @@ void ImmersedForcingBuildings_Scalar (const Box& bx,
         }
 
         // Force fully immersed cells to planar average rho and theta
-        if (t_blank == 1.0) {
+        if (t_blank == 1.0 && r_avg && t_avg) {
```
asalmgren: **Low — this guard can only test the pointer, not whether the table was filled.**

`Table1D::operator bool()` is true as soon as `resize` has run, so `r_avg && t_avg` cannot detect an unfilled table — and `t_avg(k) / rho_avg` two lines below would divide by whatever the arena held. The same guard is at line 956.

I could not construct a path that reaches here before `advance_dycore` fills the tables (including on restart), so this is defense-in-depth rather than a live bug: either zero-fill both tables right after the `resize` calls in `init_stuff`, or replace the guard with an `AMREX_ALWAYS_ASSERT` so a real misconfiguration aborts instead of silently dropping the forcing.

asalmgren (commented):
### Review at head `cfcb4813f`

Thanks for the turnaround — all seven items from the earlier pass on `e29e89929` are addressed:

- **Null `Table1D` deref** (IF + `custom_w_subsidence` + moisture): fixed by dropping the `!use_immersed_forcing` term, so `qv_plane_tab`/`qc_plane_tab` are built whenever subsidence is active.
- **`SmnSmn` signaling-NaN read**: fixed. I confirmed removing the masking loop is a true no-op — `make_tau_terms` rewrites every valid cell at `nrk==0` (`ERF_MakeTauTerms.cpp:223`) later in the same `advance_dycore` call, and `need_SmnSmn` there (`Deardorff || kEqn`) is exactly the `use_keqn` used for allocation, so nothing reads the array in between. The new `setVal(zero)` is good hardening.
- **NaN scrub**: back to assignment under `if (t_blank_arr == one)`, so non-finite values are cleared again rather than multiplied by zero.
- **`-Wshadow`**: renamed to `l_use_IF` in both TI headers.
- **u/v plane averages**: correctly confined to `use_nudging`.

Two things I'd like to see changed before merge, plus five smaller items — all inline.

**Verified sound, so it doesn't get re-litigated:**

- The ~30 new `if (mask == one)` guards in `ERF_ImmersedForcing.cpp` are exactly behavior-preserving. Every mask is assigned via `? one : zero` (lines 412-418, 575-581, 742-748), and `wall_mask`/`east_west_mask` (resp. `north_south_mask`, `roof_mask`) are mutually exclusive via `l_use_most`, so the `||` guards on their sum are equivalent. All block-local temporaries (`u1_cellaway`, `rho_*_inside`, `theta_surf`, `u_target`, `bc_forcing_*`) are reassigned before use inside each block, so skipping a block leaks nothing into the next, and the `l_implicit_drag` `lambda` re-derives from the masks. The guards additionally suppress real `NaN * 0` paths, e.g. the roof branch's `cell_data(i,j,k-1,...)` read at `k == 0`.
- Table sizing/indexing is consistent: `init_stuff` grows `geom[lev].Domain()` by 1 in z, `advance_dycore` uses `fine_geom.Domain()` with `ng_c[2] = 1`, `PlaneAverage::ncell_line()` returns `nz+2`, and the copy loop writes exactly `[-1, nz]`. The resize sits inside the same `terrain_type/buildings_type == ImmersedForcing` predicate used in `advance_dycore`, `make_sources`, and both TI headers.
- Lifetime/ordering: `advance_dycore` runs unconditionally per level per timestep before any `make_sources`, and restart reaches `init_stuff` via `ReadCheckpointFile -> MakeNewLevelFromScratch` (`ERF_Checkpoint.cpp:733`), so the persistent tables are always sized and filled before they are read.
- `ncomp = RhoQ2_comp` in `advance_dycore` (vs `RhoQ2_comp+1` in `make_sources`) is correct: `RhoQ2_comp == RhoQ1_comp+1`, so the alias covers `RhoQ1_comp`, the only moisture component averaged here.
- Both `make_sources` call sites were updated to the new signature.

Source/SourceTerms/ERF_MakeSources.cpp
adamwise95: Added a comment, but this is the intended behavior. There should not be a guard preventing immersed forcing from running with variable dz.

asalmgren (commented):
### Code review

Reviewed the planar-average refactor, the immersed-forcing mask guards, and the eddy-diffusivity zeroing change. Three issues, posted inline.

1. `AMREX_ALWAYS_ASSERT(r_avg && t_avg)` inside the `ParallelFor` in `ImmersedForcingTerrain_Scalar` (L957) and `ImmersedForcingBuildings_Scalar` (L1200) replaced a working null guard, but `amrex::Assert` compiles to `ignore_unused` on device when `NDEBUG` is set and `AMREX_USE_ASSERTION` is not -- i.e. in optimized GPU builds. The null dereference on the following line is no longer prevented.
2. `make_sources` (L111) still tolerates null tables via `if (use_immersed_forcing && r_plane_avg && t_plane_avg)`, which is what makes (1) reachable. A host-side assert here would be effective.
3. The persistent table is sized `domain.grow(2,1)` in `init_stuff` while the filler in `advance_dycore` sizes itself from `solverChoice.ave_plane`; with `erf.Ave_Plane != 2` this is an out-of-bounds device write into persistent storage.

Two non-blocking notes:

- The averages are now taken from `state_old` at the top of `advance_dycore`, i.e. before the `use_lsf` and `use_nudging` blocks mutate `state_old[IntVars::cons]` in place. With nudging and immersed forcing both enabled, the immersed-cell target rho/theta excludes the current step's nudging increment. This looks consistent with the stated intent of accepting a one-step lag, just flagging it.
- The PR description mentions "added fill boundary calls after zeroing out the eddy diffusivities", but the diff adds no `FillBoundary` -- it grows the loop box to `growntilebox(eddyDiffs->nGrowVect())` instead. That is arguably the better fix, since `terrain_blanking` is `FillBoundary`'d once at init and is static thereafter, and I verified it is in bounds (`terrain_blanking` has `ComputeGhostCells+2 >= 4` ghost cells, `Subtract`ed over `ng_sub = min(ComputeGhostCells+2, 5) >= 4`, on the same BoxArray/DistributionMap as `eddyDiffs`, which has 2). The description and the diff just disagree.

Separately, I checked the removal of the `SmnSmn` zeroing block: it was dead code, since `SmnSmn` is recomputed unconditionally at `nrk==0` in `erf_make_tau_terms`, which runs after that point. The new `setVal(zero)` at allocation is the part that has effect.

Generated with [Claude Code](https://claude.ai/code)

Source/SourceTerms/ERF_ImmersedForcing.cpp
```diff
@@ -921,7 +953,8 @@ void ImmersedForcingTerrain_Scalar (const Box& bx,
         }
 
         // Force fully immersed cells to planar average rho and theta
-        if (t_blank == one && r_avg && t_avg) {
+        if (t_blank == one) {
+            AMREX_ALWAYS_ASSERT(r_avg && t_avg);  // Tables must be filled before calling this function
```
asalmgren: **`AMREX_ALWAYS_ASSERT` is a no-op inside device code in release GPU builds, so this replaces a real guard with nothing.**

The previous condition here was `if (t_blank == one && r_avg && t_avg)`. The null check is now gone from the kernel, and the assert that replaced it does not run on the GPU. `AMREX_ALWAYS_ASSERT(EX)` expands to `amrex::Assert(...)`, and `amrex::Assert` is:

```cpp
#if defined(NDEBUG) && !defined(AMREX_USE_ASSERTION)
        AMREX_IF_ON_DEVICE((amrex::ignore_unused(EX,file,line);))
#else
        AMREX_IF_ON_DEVICE(( AMREX_DEVICE_PRINTF(...); AMREX_DEVICE_ASSERT(0); ))
#endif
```

(`Submodules/AMReX/Src/Base/AMReX_BLassert.H`, `Submodules/AMReX/Src/Base/AMReX.H` L364)

So in exactly the build this PR targets -- an optimized GPU build, where `NDEBUG` is set and `AMREX_USE_ASSERTION` is not -- the check compiles away and the very next line, `const Real rho_avg = r_avg(k);`, dereferences a null device pointer. The old code just skipped the forcing term.

This is reachable through the code as written, because `make_sources` still explicitly tolerates null tables (see the comment on `Source/SourceTerms/ERF_MakeSources.cpp` L111).

Suggested fix: keep the cheap `&& r_avg && t_avg` in the kernel condition, and move the `AMREX_ALWAYS_ASSERT` to the host side in `make_sources`, where it actually executes. Secondary point: even on host builds, this evaluates the assert once per cell inside the `ParallelFor`.

Source/SourceTerms/ERF_ImmersedForcing.cpp
```diff
@@ -1164,6 +1197,7 @@ void ImmersedForcingBuildings_Scalar (const Box& bx,
 
         // Force fully immersed cells to planar average rho and theta
         if (t_blank == 1.0) {
+            AMREX_ALWAYS_ASSERT(r_avg && t_avg);  // Tables must be filled before calling this function
```
asalmgren: Same issue as the identical assert in `ImmersedForcingTerrain_Scalar` at L957.

`AMREX_ALWAYS_ASSERT(EX)` expands to `amrex::Assert(...)`, which is compiled out in device code whenever `NDEBUG` is defined and `AMREX_USE_ASSERTION` is not -- i.e. every optimized GPU build:

```cpp
#if defined(NDEBUG) && !defined(AMREX_USE_ASSERTION)
        AMREX_IF_ON_DEVICE((amrex::ignore_unused(EX,file,line);))
#else
        AMREX_IF_ON_DEVICE(( AMREX_DEVICE_PRINTF(...); AMREX_DEVICE_ASSERT(0); ))
#endif
```

So a null `r_avg` is no longer caught, and `const Real rho_avg = r_avg(k);` on the next line dereferences a null device pointer. Please keep `&& r_avg && t_avg` in the `if` condition and put the hard assert on the host side in `make_sources`.

Source/SourceTerms/ERF_MakeSources.cpp
```diff
@@ -99,20 +101,26 @@ void make_sources (int level,
...
-    // Compute averages on slow step for subsidence, or whenever immersed forcing runs
-    // NOTE: We recompute planar averages every substep when immersed forcing uses substeps.
-    //       This is simpler than passing persistent storage through the call chain, and
-    //       the cost is relatively small compared to the full solve.
-    bool compute_averages = ( (is_slow_step && dptr_wbar_sub) ||
-                              (use_immersed_forcing && ((is_slow_step && !use_ImmersedForcing_fast) ||
-                                                        (!is_slow_step && use_ImmersedForcing_fast))) );
+
+    // Use passed-in planar averages for immersed forcing (computed once in ERF_AdvanceDycore.cpp)
+    if (use_immersed_forcing && r_plane_avg && t_plane_avg) {
```
asalmgren, line 112: **This condition silently degrades to null tables, and the downstream kernels no longer check for them.**

```cpp
if (use_immersed_forcing && r_plane_avg && t_plane_avg) {
    dptr_r_plane_if = r_plane_avg;
    dptr_t_plane_if = t_plane_avg;
}
```

If a caller passes default-constructed (null) tables while immersed forcing is enabled, `dptr_r_plane_if` / `dptr_t_plane_if` stay null and are handed straight to `ImmersedForcingTerrain_Scalar` / `ImmersedForcingBuildings_Scalar`, which in this PR dropped their `r_avg && t_avg` guards in favour of an `AMREX_ALWAYS_ASSERT` that does not execute on device in release builds (see the comments on `Source/SourceTerms/ERF_ImmersedForcing.cpp` L957 and L1200).

The two changes together turn "silently skip the forcing" into "null-pointer dereference in a GPU kernel".

Today both call sites (`ERF_TI_slow_rhs_pre.H` L112-114 and `ERF_TI_substep_fun.H` L72-74) compute `l_use_IF` with the same predicate used to allocate the tables in `init_stuff`, so it is not reachable right now -- but this is precisely the invariant the assert was meant to protect, and this is the place where a host-side assert would actually fire:

```cpp
AMREX_ALWAYS_ASSERT(!use_immersed_forcing || (r_plane_avg && t_plane_avg));
```

Source/ERF_MakeNewArrays.cpp
```diff
@@ -111,6 +111,13 @@ ERF::init_stuff (int lev, const BoxArray& ba, const DistributionMapping& dm,
         terrain_blanking_yface[lev]->setVal(one);
         terrain_blanking_zface[lev]->setVal(one);
 #endif
+
+        // Initialize planar average storage for immersed forcing
+        // Sized to match PlaneAverage output with ghost cells, using same indexing as subsidence
+        Box domain = geom[lev].Domain();
+        Box tdomain = domain; tdomain.grow(2, 1);  // Grow by 1 ghost cell in z-direction
```
asalmgren, line 118: **The table size hardcodes the z-direction, but the code that fills it sizes itself from `solverChoice.ave_plane`.**

Here the persistent table is given extent `[-1, nz]` (`nz+2` entries). The filler in `advance_dycore` derives its loop bound from the `PlaneAverage` object instead:

https://github.com/erf-model/ERF/blob/18f0c306e1b3e5eab6bd3ec2c0bde1424be1354e/Source/TimeIntegration/ERF_AdvanceDycore.cpp#L138-L142

https://github.com/erf-model/ERF/blob/18f0c306e1b3e5eab6bd3ec2c0bde1424be1354e/Source/TimeIntegration/ERF_AdvanceDycore.cpp#L172-L177

`PlaneAverage::ncell_line()` is measured along `m_axis == solverChoice.ave_plane`, not along z (`Source/Utils/ERF_PlaneAverage.H`, ctor: `domain.grow(axis_in, m_ghost_to_inc[axis_in]); m_ncell_line = dom_hi[m_axis] - dom_lo[m_axis] + 1 + ...`).

`ave_plane` is a user input (`erf.Ave_Plane`, `Source/DataStructs/ERF_DataStruct.H` L758, default 2). With `erf.Ave_Plane = 0` on an immersed-forcing run, `ncell` becomes `nx + 2*ngrow_state` while `offset` stays 1, so the `ParallelFor(ncell, ...)` writes indices up to `nx + 2*ngrow - 2` into an allocation holding only `nz+2` doubles -- an out-of-bounds write into long-lived arena memory.

The same index mismatch existed before for the nudging path, but only into a scratch `TableData` that was freed immediately; this PR makes it corrupt persistent per-level storage. Either size the table from `ncell` at fill time, or assert `solverChoice.ave_plane == 2` where the table is allocated.

PR #3897, opened 2026-08-22
Forest canopy heat diagnostics

wang1202:
## Summary

This PR adds gridded NetCDF forest-canopy support, fixed-leaf-temperature canopy heat and moisture exchange, and additional turbulence diagnostics for plotfiles and line sampling.

## Key changes

- Add NetCDF inputs for gridded canopy LAI, height, drag coefficient, and terrain, with constant-`Cd` support and a BellForest regression case.
- Add a scoped canopy exchange path that prescribes leaf potential temperature through `forest_leaf_theta_fixed`, then applies sensible heat and, when moisture is active, latent/transpiration source terms. This does not add radiation, photosynthesis, prognostic stomata, or a leaf energy-balance solve.
- Add interval mean, resolved covariance, fluctuation, and TKE plot variables with time- or plotfile-based reset controls.
- Add `profiles_xy` input handling and line sampling for `sgs_tke`, `sgs_tau13`, `sgs_tau23`, and `sgs_hfx3`.
- Document the forest/terrain inputs, fixed-leaf-temperature exchange, interval plot diagnostics, and SGS line outputs.

## Testing

- Built the CUDA/MPI/NetCDF/RRTMGP `erf_exec` and `erf_unit_tests` targets.
- Passed all 13 `Plotfile3DSelection.*` unit tests on a GPU node.
- Completed a one-step BellForest GPU smoke test exercising the forest heat inputs, mean/fluctuation plot variables, profile outputs, and SGS line-sampling variables.


asalmgren:
# Code review: PR #3897 — "Forest canopy heat diagnostics"

- URL: https://github.com/erf-model/ERF/pull/3897
- Author: Aaron Wang (`wang1202`), branch `forest-canopy-heat-diagnostics`
- Size: 30 files changed, +2572 / -122

## High

### 1. The transpiration source is silently discarded
`Source/SourceTerms/ERF_CanopyBiophysics.H:113`

`AddCanopyBiophysicsHeatSources` is called from `slow_rhs_fun_pre`
(`Source/TimeIntegration/ERF_TI_slow_rhs_pre.H:213`) and adds to
`S_rhs[IntVars::cons]` at `RhoQ1_comp`. But in every RK stage the MRI integrator
calls `slow_rhs_post` *after* `slow_rhs_pre` (`Source/TimeIntegration/ERF_MRI.H:359`),
and `AdvectionSrcForScalars` **assigns** rather than accumulates:

```
advectionSrc(i,j,k,cons_index) = -invdetJ*mfsq*(...)
```

(`Source/Advection/ERF_AdvectionSrcForState.cpp:285`, looping `ivar` over
`[RhoKE_comp, RhoQ1_comp]`). The canopy latent/vapor tendency is therefore
overwritten before it is ever used.

Failure scenario: a moist run with `forest_biophysics_heat = true` gets the
documented sensible-heat term (`RhoTheta_comp` survives, it is component 1) but
zero transpiration — the feature described in
`Docs/sphinx_doc/theory/Forest.rst` has no effect.

Fix: route the moisture term through `cc_src` / `slow_rhs_post`, not
`slow_rhs_pre`.

### 2. NetCDF terrain zeroes the upper-x / upper-y boundary nodes
`Source/ERF_ProbCommon.H:494`

The stencil guard is `i_t >= 0 && i_t < nx-1 && j_t >= 0 && j_t < ny-1`, with
`z_arr = 0.0` in the `else`. For a file whose coordinates span the domain
exactly, the node at `i == Domain().bigEnd(0)+1` gives `fx == nx-1` →
`i_t == nx-1` → guard fails → height 0. Same for `j`.

The text-file path deliberately avoids this by clamping the bracket index
(`iindex_terrain = min(max(...,0), nx-2)`, `Source/ERF_ProbCommon.H:~547`), and
the new gridded *forest* reader in this same PR also clamps
(`ii_c = min(ii, grid_nx-2)`).

Failure scenario: the shipped `Exec/RegTests/BellForest` case has ~4.4 m terrain
at the mid-edge of the eastern/northern boundary, which becomes a 4.4 m cliff in
one node row/column; a case with a non-trivial edge (e.g. a real DEM) gets a
large spurious step at two domain faces.

## Medium

### 3. The moment accumulator is reset by the wrong plot stream
`Source/IO/ERF_Plotfile.cpp:2205`

The reset runs at the end of every `Write3DPlotFile` call whenever
`mean_vars_reset_mode == "plotfile"`, regardless of whether the stream being
written contains any `*_mean` / `*_fluct` / `tke` variable.

Failure scenario: with `erf.plot_vars_1` carrying the mean fields at
`plot_int_1 = 1000` and a second stream `erf.plot_vars_2 = density` at
`plot_int_2 = 10`, the means written at step 1000 average only the last 10
steps.

Fix: condition the reset on the stream actually containing at least one
interval-mean variable.

### 4. Interval means are not checkpointed, unlike the feature they mirror
`Source/ERF_MakeNewArrays.cpp:267`

`vel_t_avg` and `t_avg_cnt` are written and restored
(`Source/IO/ERF_Checkpoint.cpp:134,252,746,905`), but `interval_means`,
`t_mean_cnt`, and `mean_vars_time_reset_done` are not. On restart `init_stuff`
zeroes them, so (a) the accumulation is silently lost and (b) in `time` reset
mode `mean_vars_time_reset_done` returns to 0, so the "one-shot" reset fires
again at the first step after restart.

Failure scenario: a run restarted at t = 5000 s with
`mean_vars_reset_time = 1000` averages from 5000 s, not 1000 s, with nothing in
the output indicating it.

### 5. The documented `HGT_M` terrain variable cannot actually be read
`Source/ERF_ProbCommon.H:413`

`read_terrain_netcdf` aborts unless `dims.size() == 2`, but `HGT_M` in WPS
`geo_em` files is `HGT_M(Time, south_north, west_east)` (3D).
`Docs/sphinx_doc/Inputs.rst` advertises `HGT_M` as an accepted name. The forest
reader added in this PR handles the leading time dimension
(`Source/SourceTerms/ERF_ForestDrag.cpp:~460`); this one does not.

### 6. Out-of-bounds device read for a single-column / single-row canopy file
`Source/SourceTerms/ERF_ForestDrag.cpp:241`

`int ii_c = amrex::min(ii, grid_nx - 2)` yields `-1` when `grid_nx == 1`, and
`lai_data[jj_c*grid_nx + ii_c]` then reads before the buffer. `read_netcdf_file`
explicitly tolerates `nx == 1` / `ny == 1` (`dx = (nx > 1) ? ... : 1.0`), so
such a file reaches this kernel. Same for `jj_c` with `grid_ny == 1`.

### 7. Grid-consistency check ignores the origin
`Source/SourceTerms/ERF_ForestDrag.cpp:71`

The LAI/height assertion compares `nx`, `ny`, `dx`, `dy` but never `xmin` /
`ymin` (`xmin_check` / `ymin_check` are read and dropped); the Cd-file assertion
(line ~92) compares only `nx` / `ny`.

Failure scenario: two files on the same-shaped grid but with shifted origins — or
a Cd file with different spacing — are accepted and the canopy is silently
mis-registered, since the kernel interpolates all three fields with
`m_grid_xmin` / `m_grid_dx` from the LAI file alone.

### 8. `sgs_tke` has no availability guard
`Source/IO/ERF_SampleData.H:473`

It reads `state(i,j,k,RhoKE_comp)` unconditionally. `RhoKE_comp` is always
allocated (index 2 of NDRY=3) but is only meaningful for Deardorff / k-eqn /
MYNN; for Smagorinsky or `les_type = None` the sampled column reports whatever
sits in that unused slot. The other three new SGS variables do have a nullptr
fallback, and the docs promise "if their diffusion storage is unavailable, ERF
writes zero" — `sgs_tke` should get an equivalent guard on the turbulence model.


## Low

### 9. Raw `double` literals in device code break `PRECISION=FLOAT`
`Source/SourceTerms/ERF_ForestDrag.cpp:~200` (gridded kernel)

e.g. `amrex::max((z_cc(i,j,k) - z_sfc), 0.0)`, `0.5 * dx[0]`,
`std::pow(ratio, 6.0)`. With `amrex::Real == float`,
`amrex::max(float, double)` fails template deduction. The code being replaced
used the typed constants `zero`, `one`, `myhalf`, `fourth` for exactly this
reason. `Source/SourceTerms/ERF_CanopyBiophysics.H:110`
(`amrex::max(qv_saturation - qv, 0.0)`) has the same issue.

### 10. Thermodynamic constants are re-declared instead of reused
`Source/SourceTerms/ERF_CanopyBiophysics.H:53`

`latent_heat = 2.501e6` vs `L_v = 2.5e6`, and `epsilon = 0.622` vs `RdoRv` in
`Source/ERF_Constants.H` (`cp_dry = 1004.5` does match `Cp_d`). The canopy
source will drift from the solver's thermodynamics if the shared constants are
ever changed.

### 11. Dead code with a latent correctness trap
`Source/SourceTerms/ERF_ForestDrag.cpp:422,542`

`m_cumLAI` and `interpolate_gridded_data` are never read by anything (only
exposed via `get_cumLAI()`), yet `define_drag_field` now allocates two extra
MultiFabs with a ghost cell per level for every forest run. The `cumLAI`
integration also uses the uniform `dx[2]` as the layer thickness, which is wrong
for terrain-following or stretched grids — worth fixing or dropping before a
consumer is written against it.

### 12. Hardcoded Cray `ncgen` path
`Exec/RegTests/BellForest/gen_nc_files.py:21`

`NCGEN = "/opt/cray/pe/netcdf/4.9.0.3/bin/ncgen"` contradicts the module
docstring ("requires ncgen on PATH") and makes the generator fail for everyone
off that machine. Use `"ncgen"`, or an env override.

### 13. Canopy heat bypasses the anelastic RK2 stage weighting
`Source/TimeIntegration/ERF_TI_slow_rhs_pre.H:213`

`erf_slow_rhs_pre` halves `cell_rhs(RhoTheta_comp)` in the anelastic `nrk == 1`
stage (`Source/TimeIntegration/ERF_SlowRhsPre.cpp:750`); the canopy source is
added afterwards and so receives full weight while every other `RhoTheta`
tendency is halved. Routing it through `cc_src` would avoid the inconsistency.

### 14. `erf.forest_biophysics*` is silently inert without a forest configuration
`Source/ERF.cpp:2614`

The `do_forest_drag` consistency check aborts on a mismatched switch, but nothing
validates `forest_biophysics_heat = true` against the absence of forest files. A
user who sets `forest_biophysics_heat` / `forest_leaf_theta_fixed` and forgets
`forest_lai_file` gets `do_forest_drag = false`, the guard in
`ERF_TI_slow_rhs_pre.H` fails, and the run proceeds with no canopy heat and no
message.

asalmgren:
# Code review: PR #3897 — "Forest canopy heat diagnostics"

- **Author:** Aaron Wang (`wang1202`)
- **Branch:** `forest-canopy-heat-diagnostics` → `development`
- **Head commit reviewed:** `ec409c9eb422246d31d007534acafd20ce57b46e`
- **Scope:** 38 files (source, docs, unit tests, new `Exec/RegTests/BellForest` case)

All findings below were verified directly against the PR branch.

---

## 1. `Source/ERF_ProbCommon.H:398` — high — `mass_point_terrain` is never broadcast, so non-I/O ranks get wrong WPS terrain

`bool mass_point_terrain = false;` is declared outside the `if (ParallelDescriptor::IOProcessor())`
block and set to `true` only inside it (line 612), but only `nx`, `ny`, `m_xterrain`, `m_yterrain`
and `m_zterrain` are broadcast (lines 665–676). The flag is then read inside the device lambda
(line 710) on **all** ranks.

**Failure scenario.** Run with `erf.terrain_file_name_nc = geo_em.d01.nc` (the WPS `HGT_M`
mass-point path, no Cartesian `x`/`y` variables) on ≥2 MPI ranks. On rank 0 the clamp
`x = min(max(x, d_xt[0]), d_xt[nx-1])` is applied and the outer node ring gets the nearest mass
value; on every other rank the clamp is skipped, `uniform_interpolation_stencil` returns
`inside == false` for the outer half-cell ring, and `z_arr` is set to `0.0`. Terrain therefore
differs between ranks depending on box ownership. The new unit test
`TerrainNetCDF.ReadsGenuineWpsGeoEmMassField` would itself fail under `mpirun -n 2`
(`lower_corner` would be `0.0`, not `1.0`).

**Fix.** `ParallelDescriptor::Bcast` the flag (e.g. as an `int`) alongside `nx`/`ny`.

---

## 2. `Source/ERF_ProbCommon.H:679` — medium — device vectors freed while the interpolation kernel may still be running

`d_xterrain`/`d_yterrain`/`d_zterrain` are `Gpu::DeviceVector`s declared at line 679; the
`ParallelFor` at line 700 that dereferences `d_xt`/`d_yt`/`d_zt` is the last statement in
`read_terrain_netcdf`, so the vectors are destroyed immediately afterwards with no
`Gpu::streamSynchronize()`.

**Failure scenario.** `ArenaAllocatorBase::deallocate`
(`Submodules/AMReX/Src/Base/AMReX_GpuAllocators.H:58`) just calls `arena()->free(ptr)` — it is not
stream-ordered. On a GPU build the async kernel can still be reading memory that has been returned
to `The_Arena()` and handed out to the next allocation, producing intermittently corrupted terrain.

Note the PR gets this right in the parallel code path it added in `ERF_ForestDrag.cpp`
("Synchronize before DeviceVectors go out of scope" / `Gpu::synchronize()`); the terrain reader is
missing the same call.

---

## 3. `Source/ERF_MakeNewArrays.cpp:270` — medium — a regrid silently wipes the interval-mean window on *every* level in `time` reset mode

`init_stuff` sets `mean_vars_time_reset_done[lev] = 0` whenever a level is (re)built, and
`RemakeLevel` calls `init_stuff` (`Source/ERF_MakeNewLevel.cpp:621`). The reset predicate in the new
block in `ERF::Advance` is `reset_needed = any level has flag == 0`, and when true it calls
`ResetIntervalMeans()`, which zeroes **all** levels.

**Failure scenario.** `erf.compute_mean_vars = true`, `erf.mean_vars_reset_mode = time`,
`erf.mean_vars_reset_time = 3600`, `amr.max_level = 1`, `amr.regrid_int = 10`. The intended one-shot
reset happens at t = 3600 s; at the next regrid, level 1's flag returns to 0 and the following
`Advance` re-runs `ResetIntervalMeans()`, discarding level 0's accumulated window even though
level 0's grids never changed. Every subsequent regrid repeats this, so `*_mean`/`*_fluct`/`tke`
only ever cover the interval since the last regrid.

---

## 4. `Source/SourceTerms/ERF_ForestDrag.cpp:291` — medium — `erf.forest_laimax` is unvalidated and produces Inf/NaN drag for `laimax >= 1`

In the gridded path, `treeZm = laimax * height_interp`, and the normalization integral takes the
`ztree < treeZm` branch with `ratio = (h - treeZm)/(h - ztree)`.

**Failure scenario.** With `erf.forest_laimax = 1.0` and `erf.forest_tree_type = 2`,
`treeZm == h`, so `ratio == 0` for every quadrature point, `expFun == 0`, and
`af = lai_interp / (expFun * dz)` is `+Inf` — the drag field becomes Inf and the solution NaNs on
the first step. With `erf.forest_laimax = 1.2`, `ratio < 0` and the `else` branch evaluates
`std::sqrt(ratio)` → NaN.

Both constructors already `AMREX_ALWAYS_ASSERT` on `tree_type`, but `laimax` is accepted unchecked
(`ERF_ForestDrag.cpp:49`, `:102`). An assert of `0 <= laimax < 1` would catch this at input-parse
time instead of as silent NaNs.

---

## 5. `Source/Utils/ERF_GridUtils.H:31` — medium — the uniformity tolerance (64·eps, relative) rejects legitimate coordinate arrays stored as `fl
oat`

`validate_uniform_axis` requires `|coord[i] - (origin + i*spacing)| <= 64*eps_double*scale`, i.e.
~1.5e-12 absolute at a magnitude of 105. The docs the PR adds explicitly advertise `lon`/`lat`
coordinates for forest files.

**Failure scenario.** A canopy file whose `lon` variable is `float` (very common for
remote-sensing/geo products) running from -105.0 with spacing 0.01 — the float32 round-off is
~6e-6, four orders of magnitude above the tolerance — so ERF aborts with
"has nonuniform lon coordinates, which are not supported" for a perfectly uniform grid.

A tolerance tied to the *spacing* (e.g. `1e-4 * spacing`) or to `float` epsilon would accept these
while still catching genuinely non-uniform axes.

*Minor:* the `index > 0 &&` conjunct in the same loop is always true and can be dropped.

---


## 6. `Source/SourceTerms/ERF_ForestDrag.cpp:291` — low — the LAD normalization integral is recomputed per cell inside the device kernel

For `tree_type == 2` the 100-iteration `expFun` loop (with `std::exp`/`std::sqrt` per iteration) now
runs *inside* the `ParallelFor` for every canopy cell, whereas the discrete-patch path hoists it out
of the kernel (it is per patch). Because `height_interp` varies horizontally it cannot be fully
hoisted, but it is independent of `k`, so `define_drag_field` for a large canopy does ~100
transcendental evaluations per cell per vertical level instead of per column. This is a startup-time
cost only, but a 2–3 order-of-magnitude regression relative to the patch path for large gridded
canopies.

---

## 7. `Docs/sphinx_doc/Inputs.rst:812` — low — a grid-table row separator was deleted, merging two parameter rows

The hunk at the profile-output table removes the `+---+---+---+---+` line between
`**erf.profile_int**` and `**erf.destag_profiles**`. docutils treats the two rows as one multi-line
cell, so the rendered table shows `erf.profile_int` and `erf.destag_profiles` (and their definitions
and defaults) collapsed into a single row. Re-adding the separator restores the table; this edit
looks accidental, since the rest of the hunk only rewords the prose below it.

---

## 8. `Source/SourceTerms/Make.package` — low — the new header is not registered for GNU-make builds

`Source/Utils/Make.package` correctly gains `CEXE_headers += ERF_GridUtils.H`, but
`Source/SourceTerms/Make.package` was not updated for the new `ERF_CanopyBiophysics.H`. Compilation
still succeeds (the directory is on the include path), but the header is absent from make's
dependency list, so editing it will not trigger a rebuild of the objects that include it in a
GNU-make build tree — a stale-object footgun for anyone iterating on the canopy source terms.

---

## 9. `Exec/RegTests/BellForest/inputs` — low — the "BellForest regression case" is not registered as a test

The PR description lists "a BellForest regression case", and the directory adds `inputs`, three
`.cdl`/`.nc` pairs and a generator, but `Tests/CTestList.cmake` and `Tests/test_files/` are
untouched, so nothing under `ctest -L regression` exercises it and no gold file exists. Either
register it (and add the gold file) or drop the claim; as it stands the gridded-NetCDF forest path
and the new `terrain_file_name_nc` path have no regression coverage — only the serial unit tests,
which (see finding 1) do not cover the MPI case where the WPS path is broken.

Also worth noting: the committed `.nc` binaries are redundant with the `.cdl` + `gen_nc_files.py`
pair.

---


## Checked and found correct

These were examined and are **not** problems, recorded so they don't get re-reviewed:

- The new plot-variable block sits in the right position relative to `derived_names` ordering.
- `mf_comp` bookkeeping and the fluctuation/TKE component indices.
- All `Write3DPlotFile` call sites were updated for the new `bool` return.
- `cc_src` is fully zeroed by `make_sources(..., is_slow_step=true)` each RK stage, so the canopy
  source neither accumulates nor double-counts; the substep's `make_sources(..., false)` only clears
  `Rho`/`RhoTheta`, so the staged `RhoQ1` source survives to `erf_slow_rhs_post`.
- `Tau[lev][tau13]` / `SFS_hfx3_lev[lev]` nullptr and size guards.
- The `sgs_*` edge/face→cell averaging stencils.
- `dt[lev]` *is* set by `ComputeDt()` before the `InitData_post` accumulation.
- The double `#include` of `ERF_InitCustomPertVels_WitchOfAgnesi.H` is safe (no include guard).
- `init_stuff` is reached from `RemakeLevel`, so `interval_means` is not left on a stale `BoxArray`.

asalmgren (commented):
Thanks for the thorough documentation and the test fixtures here. A few things need attention before this can go in — two of them will fail CI as-is. Findings are grouped by severity.

## Blocking

### 1. `Docs/sphinx_doc/Inputs.rst` (~line 812) — malformed grid table breaks the Sphinx build

A row separator lost its leading `+`, in the "List of Parameters" table between the `erf.profile_int` and `erf.destag_profiles` rows:

```
-+-------------------------------+------------------+----------------+----------------+
+-------------------------------+------------------+----------------+----------------+
```

The added line starts with `-` instead of `+`. Sphinx errors on a malformed grid table. This looks unintended — the only deliberate change in that hunk is dropping "By default, " from the paragraph below.

### 2. `Tests/Unit/SourceTerms/ERF_GTestCanopyBiophysics.cpp` — aborts the whole gtest binary in a build without NetCDF

The file is added to the *unconditional* `target_sources` list in `Tests/Unit/CMakeLists.txt`, while the sibling `IO/ERF_GTestTerrainNetCDF.cpp` is correctly wrapped in `if(ERF_ENABLE_NETCDF)`. `ERF_ENABLE_NETCDF` defaults to `OFF` (`CMakeLists.txt:134`), and `.github/workflows/gcc.yml`, `macos.yml`, and `ci.yml` all configure `ERF_ENABLE_TESTS=ON` without it, then run `ctest -L unit`.

In `ForestDrag.TypeTwoProfileProducesFinitePositiveFrontalArea`, the gridded constructor validates `tree_type`/`laimax` first (0.6 passes), then falls into the `#else` branch and calls `Abort("ERF must be compiled with NetCDF support to use gridded forest data")` — this is not an `EXPECT_DEATH`, so it kills the process rather than failing one test.

Same file, secondary issue: `fixture_dir = "../test_files/BellForest/"` depends both on the gtest working directory and on those files having been copied into the build tree, which only happens because `setup_test()` runs for the `BellForest` regression — itself inside `if(ERF_ENABLE_NETCDF)` in `CTestList.cmake`. Both tests want the same `#ifdef ERF_USE_NETCDF` / CMake guard.

### 3. First `EXPECT_DEATH` in the suite, in an MPI- and GPU-linked binary

`RejectsInvalidLaimaxBeforeReadingGriddedFiles` introduces the only death test in `Tests/Unit`. gtest forks for these, and `amrex::Abort` in the child calls `MPI_Abort(MPI_COMM_WORLD, ...)`, which can take the parent down; fork-after-CUDA-init is also unsupported on the HIP/SYCL matrices.

This PR already has the better pattern: `erf_grid_utils::validate_uniform_axis` returns an error string instead of aborting. Exposing `validate_laimax` the same way and asserting on the returned string avoids the fork entirely. Failing that, set `gtest_death_test_style = "threadsafe"` and guard the test out for MPI/GPU builds.

## Should fix

### 4. Interval-mean restart is a no-op in the default configuration

Two issues compound here.

`ERF::InitData_post` calls `Accumulate_Interval_Means` unconditionally when `compute_mean_vars` is set (`Source/ERF.cpp:1389`) — including on restart, *after* `ReadCheckpointFile` has restored `t_mean_cnt[lev]` and `interval_means[lev]`. That adds a spurious `dt`-weighted sample of the restored state, so `t_mean_cnt` no longer equals the elapsed averaging time (`51*dt` after 50 advanced steps).

Then, because `plot_file_on_restart` defaults to `true` (`Source/ERF.cpp:93`), `InitData_post` writes a 3-D plotfile and `ResetIntervalMeansAfter3DPlotfileBatch` zeroes the accumulator in `plotfile` reset mode. So the default restart sequence is: restore window, add a spurious sample, write it, discard the window. With `erf.plot_int_1 = 100` and a checkpoint at step 50, the step-100 plotfile averages 50 steps where a continuous run averages 100.

Net effect: the new format-versioned `IntervalMeansHeader` / `IntervalMeans` checkpoint machinery has no observable effect in the default configuration. Worth deciding deliberately given how much code it is. Skipping the `InitData_post` accumulation when `!restart_chkfile.empty()` fixes the first half.

### 5. `uniform_interpolation_stencil` bound check has no float32 tolerance, unlike its own validator

`Source/Utils/ERF_GridUtils.H` admits a coordinate only within `64 * eps(Real) * scale` of the file's grid, and `read_terrain_netcdf` assigns `0.0` to anything outside. But NetCDF coordinate variables are routinely stored as `float`, so a file that genuinely spans the domain reads back quantized: for `prob_hi = 1234.5678`, `d_xt[nx-1]` lands ~6e-5 short of the last ERF node against a ~2e-11 tolerance. The last row and column of `z_phys_nd` then silently become 0 — a terrain cliff with no diagnostic.

The inconsistency is internal to the new header: `comparison_tolerance()` in the same file carries a deliberate float32-ulp allowance, with a comment explaining exactly this quantization, but the interpolation bound check doesn't use it. Note that the `BellForest` fixture (2000 m domain, 50 m spacing) is exactly representable in float32, so the regression test can't catch this.

### 6. Silent zeros when a gridded field doesn't cover the domain

More generally: `read_terrain_netcdf` sets `z_arr = 0` outside the file footprint and `define_drag_field` leaves drag at 0, so a misaligned file produces a plausible-looking run with no terrain and no canopy. A one-time count of target points that fell outside the source grid would make this diagnosable. Contrast `read_custom_terrain`, which clamps and never zeroes.

Related: `ForestDrag::read_netcdf_file` accepts `lon`/`lat` as a substitute for `x`/`y` behind only a `Print()` warning. A degrees-based file makes every stencil return `inside=false`, so canopy drag is silently zero everywhere. `read_terrain_netcdf` in `ERF_ProbCommon.H` deliberately rejects lon/lat — worth making these two consistent.

### 7. `m_frontal_area` is allocated for every forest run

`MultiFab(ba,dm,1,1)` on every level, for every forest configuration, though `forest_biophysics_heat` is its only consumer. Since `m_forest_drag == cd * m_frontal_area`, this doubles canopy field memory for all existing `erf.forest_file` users. Suggest gating the allocation on `solverChoice.forest_biophysics_heat`.

### 8. `tke` as a plot variable name

This is *resolved* TKE, while ERF's SGS TKE lives in the state as `rhoKE` — both would be legal in the same `erf.plot_vars_1` list. Since plot variable names are a user-facing contract, `tke_resolved` (or similar) is much cheaper to choose now than to rename later.

### 9. Dead / no-op additions

- `plot3d_batch_resets_interval_means(int, bool, const std::string&)` in `ERF_PlotfileSelection.H` has no callers — all four production sites and the new unit test use the 4-arg form. It's also an overload trap: `(n, true, "plotfile")` silently means "any batch resets", the semantics the PR is explicitly avoiding.
- `Tests/CTestList.cmake`: the new `TEST_FILES_DIR` one-value arg to `add_test_r` is passed `"BellForest"`, identical to `TEST_NAME`, and `setup_test()` already falls back to `${TEST_NAME}` when it's empty. The whole addition is a no-op.
- `Tests/Unit/IO/ERF_GTestIntervalMeansRestart.cpp` doesn't test ERF code — it `VisMF::Write`s a MultiFab and reads it back under a reversed `DistributionMapping`, which is a test of AMReX and would pass with this PR reverted. The actual new logic is the `IntervalMeansHeader` parser in `ReadCheckpointFile` (version string, level/component counts, BoxArray match, pmap record, legacy fallback), none of which is exercised. Also `EXPECT_EQ(restored.norm0(0, 1, false), 0)` checks only component 0 of the 2 it subtracted.
- `Source/SourceTerms/ERF_ForestDrag.cpp`: the constant-`Cd` path materializes a whole constant `m_gridded_cd` field, uploads it, and bilinearly interpolates a constant per canopy cell. `m_use_const_cd` is already stored but never read — using it in the kernel drops a device array and three loads per cell.

## Nits

- `Exec/ERF_Prob.cpp` duplicates the whole `#include "Prob/ERF_InitCustomPertVels_WitchOfAgnesi.H"` body for `bellforest` instead of extending the preceding condition with `|| my_prob_name_ci == "bellforest"`. Two copies now have to stay in sync.
- `validate_laimax` and the `forest_file` parse error both report `erf.forest_laimax`, but in discrete-patch mode the offending value came from column 8 of the forest file. The message should name the file and row.
- `AddCanopyBiophysicsHeatSources` checks only `forest_biophysics_heat`, while its call site checks `do_forest_drag && forest_biophysics && forest_biophysics_heat`. Harmless, but the internal guard should match.
- The canopy heat source ignores `solverChoice.forest_substep`, which the momentum drag honors (`ERF_MakeMomSources.cpp:93`). Probably intentional — worth confirming.
- Indentation of the new `sgs_tke` `else` block in `ERF_SampleData.H`: the braces are correct, but the body and closers are misaligned enough to read as a bug.
- The PR description claims "`profiles_xy` input handling", but `profiles_xy` appears nowhere in the diff — only the leftover `Inputs.rst` paragraph edit noted in item 1.

## Verified correct

Recording these so they don't get re-litigated:

- The `compute_lad_normalization` refactor is algebraically identical to the previous inline quadrature: `ratio = (1 - z_m/h)/(1 - z/h)` makes the profile integral height-independent, so `af = LAI/(h * norm)` reproduces the old `af = LAI/(expFun * dz)` exactly. Good catch to hoist ~100 `exp` calls per canopy cell out of the kernel.
- `cc_src` staging is placed correctly. It's fully zeroed by `make_sources(..., is_slow_step=true)` before `AddCanopyBiophysicsHeatSources`; ρθ is consumed by `erf_slow_rhs_pre`, ρq_v by `erf_slow_rhs_post`, and the substep's `make_sources(is_slow_step=false)` re-zeros only `Rho`/`RhoTheta`, so the vapor source survives to the post stage and is applied exactly once per RK stage. Units check out on both source terms and the Exner factor cancels correctly.
- `mf_comp` fill order for the 17 new interval-mean/fluct/`tke` variables matches their position in the `ERF.H` name list (inserted after `umag_t_avg`, before `nut`, filled in exactly that slot), and `capabilities.interval_mean_storage` gates name selection consistently with the fill guard.
- `read_terrain_netcdf` is called collectively, so the IO-rank-only read plus `Bcast` can't deadlock, and the `Gpu::synchronize()` before the device vectors leave scope is correctly placed.
- `Tau[lev]` is always `resize(9)` with possibly-null entries, and every new SGS sampler path null-checks; the `tau13`/`tau23` (nodal x/z, y/z) and `hfx3` (nodal z) stencils stay inside valid data for `mfi.tilebox()`.
- `RemakeLevel` calls `init_stuff`, so `interval_means` is rebuilt on the new BoxArray on regrid, matching the `vel_t_avg` handling.
- `ERF.H` declares the new members twice under mutually exclusive `ERF_USE_MULTIBLOCK` branches — not a duplicate definition.
- Documentation coverage is complete: every new `erf.*` parameter, all 17 new 3-D plot variables (with the restriction subsection in `Plotfile3DReference.rst`), and the four new line-sampling variables are documented, and the equations in `theory/Forest.rst` match the implementation term for term.

## Process note

This is four largely independent features in one PR — gridded NetCDF canopy input, NetCDF terrain input, fixed-leaf-temperature canopy exchange, and the interval-mean/turbulence diagnostics. They touch disjoint code and have disjoint failure modes. Splitting them would make each much easier to review and to bisect later.

wang1202:
@asalmgren please check the update.

# PR3897 Round 3

Implemented and pushed the round-three fixes to `forest-canopy-heat-diagnostics`.

- Commit: `845d7001`
- `MyBuild/cmake.sh`: successful at 100%
- NetCDF-off unit tests: 9 passed via Slurm
- NetCDF-on ForestDrag source: compiled successfully
- Full NetCDF-on unit tests were limited by the server's serial HDF5 headers, which lack the required parallel HDF5 APIs.

The interval-mean restart logic now preserves restored diagnostics and avoids double accumulation or reset during restart initialization.

asalmgren:
# Code review: PR #3897 — "Forest canopy heat diagnostics" (second pass)

Automated correctness review, single careful pass over the full diff
(30 files, ~5.7k lines). Six findings, none blocking. Line numbers refer to
the head of `forest-canopy-heat-diagnostics` at review time.

## Verified correct

Listing these explicitly because they are the parts most likely to break
silently, and they all check out:

- The 17 new interval-mean names sit in `derived_names` at exactly the
  position where `Write3DPlotFile` fills them (between the `*_t_avg` block and
  `nut`), so `mf_comp` stays aligned.
- `derived_subvol_names` is only 5 entries, so the subvolume path is
  unaffected by that insertion.
- `cc_source` is allocated with `nvars`, so the new `RhoQ1_comp` canopy source
  is in range; it is consumed by `erf_slow_rhs_post`, and
  `make_sources(is_slow_step=false)` zeroes only `Rho`/`RhoTheta`, so there is
  no double-application.
- The checkpoint `IntervalMeansHeader` parser matches `BoxArray::writeOn`'s
  exact output format, and the checkpoint is always written *after* the
  plotfile reset within a step.
- The one-shot time-reset flag survives regrid and restart correctly.
- `Tau[lev][...].get()` / `SFS_hfx3_lev` nullptrs are guarded for uncreated
  levels.
- The SGS edge/face-to-cell averaging stencils and staggering are right.
- `compute_lad_normalization` is algebraically identical to the discrete-patch
  integral it replaced, so discrete-patch results are unchanged.

## Medium

### 1. The `std::min` cancels the float32 allowance it exists for
`Source/Utils/ERF_GridUtils.H:91`

The comment above this line says the ulp-based tolerance exists so that a
valid float32 axis is not rejected "because each stored coordinate was rounded
independently". But taking `min` with `0.05 * |spacing|` re-introduces the
rejection in exactly that case: whenever the spacing is small relative to the
coordinate magnitude, `0.05*spacing` is the smaller term and the float32
allowance never applies.

Concrete failure: a 1 m canopy raster in UTM (`x ~ 5e5`) with `x`/`y` stored as
`float`. `comparison_tolerance` is about 0.477, `0.05*spacing` is 0.05, so the
effective tolerance is 0.05 — but two independently rounded float32
coordinates near 5e5 (ulp about 0.03) can make `|local_spacing - spacing|`
reach ~2 ulp, i.e. 0.06–0.13. `validate_uniform_axis` then aborts a perfectly
uniform grid with "has nonuniform x coordinates, which are not supported", and
there is no input the user can change to get past it.

Fix: use `std::max` of the two rather than `std::min` — the intent is "at
least a few float ulps, and also at least a tiny fraction of the spacing".

Related, same function (line 82): `spacing` is derived solely from
`coordinates[1] - coordinates[0]`, and `uniform_interpolation_stencil` then
reconstructs `upper = origin + (point_count-1)*spacing`. For a long float32
axis the accumulated drift from the actual stored `coordinates[n-1]` can reach
`(n-1) * ulp` (about 60 m for n=1000 at x ~ 5e5), far larger than the stencil's
own bound tolerance — so points near the far edge can be mapped to the wrong
cell instead of being caught. Deriving `spacing` from
`(coordinates.back() - coordinates.front()) / (n-1)` removes the drift.


### 2. `MAP_PROJ = 6` is accepted, but its `DX`/`DY` are degrees, not metres
`Source/ERF_ProbCommon.H:514`

This check admits projection 6 (latitude/longitude) alongside 1/2/3. For 1/2/3
the WPS `DX`/`DY` global attributes are in metres and the rest of the branch is
fine. But in a `geo_em` file written with `map_proj = 'lat-lon'`, `DX` and `DY`
are in **degrees**.

The consequence is downstream in the same block: `close_enough(dx[0], wps_dx)`
(around line 601) compares ERF's metres-per-cell against a degree value, and
`m_xterrain[i] = geom.ProbLo(0) + (i+0.5)*wps_dx` (around line 621) builds the
source axis in degrees. So a user with a lat-lon `geo_em` file either hits the
misleading abort "WPS terrain grid does not match the ERF Geometry" (metres vs.
degrees, with nothing pointing at the unit mismatch), or — if they set
`geometry.prob_hi` in degrees to satisfy the check — silently gets
metre-valued terrain heights on a degree-spaced horizontal grid, corrupting the
terrain-fitted metrics with no warning.

Fix: drop 6 from the accepted set with a specific abort message, or convert
`DX`/`DY` to metres using `CEN_LAT`. The new `Inputs.rst` text also advertises
projection 6 as supported and needs the same correction.

## Low

### 3. Tiled `MFIter` with `validbox()` inflates the coverage counts
`Source/SourceTerms/ERF_ForestDrag.cpp:59`

`TilingIfNotGPU()` is `true` on CPU, so this `MFIter` yields one entry **per
tile** (default tile size `(1024000,8,8)`), while `mfi.validbox()` on the next
line always returns the whole FAB box regardless of tiling. `reduce_op.eval`
therefore re-evaluates the entire valid box once per tile. The box is also 3-D
even though the coverage test is purely horizontal (the lambda ignores `k`), so
every `(i,j)` is counted `nz` times.

Concrete: for a BellForest-style 40x40x40 grid the warning prints
"outside 1600000 of 1600000" instead of "1600 of 1600". The percentage survives
because both counters inflate identically, but the absolute counts are
meaningless and the check does roughly 25 x 40 times more device work than
needed.

Fix: use `mfi.tilebox()`, and restrict the box to a single `k` (a 2-D slab, as
the terrain version in `ERF_ProbCommon.H` does with `setSmall(2,...)` /
`setBig(2,...)`).


### 4. `ReduceLongSum` over values every rank computed identically
`Source/ERF_ProbCommon.H:729`

`read_terrain_netcdf` is reached from `prob->init_terrain_surface(...)`, which
every rank calls with its own private full-domain `FArrayBox`
(`FArrayBox terrain_fab(makeSlab(bx,2,0),1)`) — not a distributed MultiFab, so
there is nothing to reduce over.

Concrete: the same case on 64 ranks reports 64x the real counts
("outside 6400 of 102400" for what is actually "outside 100 of 1600"). The
percentage is unaffected; the absolute numbers would mislead anyone sizing
their terrain file.

Fix: drop the two reductions and print the local values. The collective itself
is safe (all ranks take the same branch) — this is purely about the numbers.

### 5. New `need_frontal_area` / `level` arguments only take effect on restart
`Source/SourceTerms/ERF_ForestDrag.H:59`

`define_drag_field` gained `need_frontal_area = true` and `level = -1`, and
`Source/ERF.cpp:666` passes real values. But the three other call sites in
`Source/ERF_MakeNewLevel.cpp` — lines 232 (`MakeNewLevelFromScratch`), 363
(`MakeNewLevelFromCoarse`) and 662 (`RemakeLevel`) — were not updated and so
use the defaults. Since the `ERF.cpp` call is inside
`if (!restart_chkfile.empty())`, the fresh-run and regrid paths go exclusively
through the un-updated sites.

Consequences:

1. On every fresh run and after every regrid, `m_frontal_area` is allocated and
   filled even when `forest_biophysics` is off — so the memory saving the flag
   was added for never applies except on restart. The paths also disagree
   observably: for identical inputs a fresh run has
   `get_frontal_area() != nullptr` while a restart has it `nullptr`.
2. `warn_if_forest_grid_does_not_cover_targets` prints "at level -1" for
   exactly those runs.

Fix: pass `solverChoice.forest_biophysics && solverChoice.forest_biophysics_heat`
and `lev` at all three `ERF_MakeNewLevel.cpp` sites, and consider dropping the
defaults so a future call site cannot silently regress this.

### 6. A regrid truncates the averaging window for the regridded level only
`Source/ERF_MakeNewArrays.cpp:273`

`init_stuff` is called from `MakeNewLevelFromScratch` (line 99),
`MakeNewLevelFromCoarse` (317) **and** `RemakeLevel` (621) in
`ERF_MakeNewLevel.cpp`, so on every regrid this unconditionally drops
`interval_means[lev]` and resets `t_mean_cnt[lev] = 0.0` for the affected level
while the other levels keep accumulating.

Concrete: a two-level run with `erf.compute_mean_vars = true` and
`erf.mean_vars_reset_mode = time` that regrids level 1 shortly before an output
writes a plotfile whose `uw_fluct` is averaged over ~1000 s on level 0 and ~5 s
on level 1, with nothing in the plotfile indicating the mismatch.

`vel_t_avg` has the same behaviour two lines above, so this is a carry-over of
existing design rather than a new regression. But the new fields are documented
as sharing one global window, and `ResetIntervalMeans()` already exists —
calling it here (or from the regrid path) would make each plotfile
self-consistent across levels. Worth a decision either way; a note in
`Plotfile3DReference.rst` is the minimum.


## Note: stale PR description

The summary claims "Add `profiles_xy` input handling", but `profiles_xy` does
not appear anywhere in the diff — the only related doc change is a reworded
sentence about `erf.destag_profiles` in `Inputs.rst`. Either the change was
dropped from the branch or the description needs updating.

PR #3906, opened 2026-08-24
Fix unfilled ghost-node thickness in particle deposition

### LINKED ISSUE #3903: [Bug]: `dz_cell` in `ERFPCParticleToMesh` reads unfilled `z_phys_nd` ghost nodes
### Summary

# Unfixed issue: `dz_cell` in `ERFPCParticleToMesh` reads unfilled `z_phys_nd` ghost nodes

Context: while fixing the `SDM_Box3D_IceAgg` SIGFPE (see the zero-CIC-weight
early return added to `deposit()` in `Source/Particles/ERFPCParticleToMesh.H`),
a deeper problem was identified but deliberately left alone, because fixing it
properly changes answers for nonzero-weight ghost deposits and belongs to the
SDM/particles owner.

## What is wrong

`ERFPCParticleToMesh` deposits each particle with CIC weights onto the eight
corners `(ix|ix+1, iy|iy+1, kz_lo|kz_hi)`. Each deposit normalizes by a
node-averaged cell thickness:

```cpp
auto dz_cell = [=] (int ic, int jc, int kc) -> Real {
    return Real(0.25) * (
          zheight(ic  , jc  , kc+1) - zheight(ic  , jc  , kc)
        + zheight(ic+1, jc  , kc+1) - zheight(ic+1, jc  , kc)
        + zheight(ic  , jc+1, kc+1) - zheight(ic  , jc+1, kc)
        + zheight(ic+1, jc+1, kc+1) - zheight(ic+1, jc+1, kc) );
};
```

For a particle in the outermost cell of the domain, the `ix+1` / `iy+1` corner
is a ghost cell, so `dz_cell` reads `z_phys_nd` at nodal indices one past the
last valid node. Those ghost nodes are **not filled** -- they hold 0 -- so the
thickness computed there is wrong.

Note that `kc` is clamped into `[0, k_max]` for the `dz` lookup, but `ic` and
`jc` are not clamped at all.

## How it showed up

In `SDM_Box3D_IceAgg` on 2 MPI ranks (`ERF_TEST_NRANKS=2`, the CI setting) a
particle at `pos = (187.5, 12.5, 62.5)` deposits into the corner
`ic = 8, jc = 0, kc_dz = 3`, where the four node-difference terms were

```
(8,0): 100 - 75 =  25
(9,0):   0 - 75 = -75     <-- unfilled ghost node
(8,1): 100 - 75 =  25
(9,1): 100 - 75 =  25
```

summing to **exactly 0**. The CIC weight for that corner is also exactly 0
(the particle sits on a cell-center plane), so the contribution evaluated as
`0.0 / 0.0` -> NaN. The test input sets `amrex.fpe_trap_invalid = 1`, so this
raised SIGFPE and aborted during `ERF::InitData_post` ->
`SuperDropletsMoist::FinishInit` -> `SuperDropletPC::cloudRainDensity` ->
`ERFPCParticleToMesh`. The same run passed on 1 and 4 ranks, where the
accidental cancellation to exactly zero did not occur.

## What was fixed, and what was not

Fixed: `deposit()` now returns early when either CIC weight is exactly zero.
That removes the `0/0` and is answer-preserving, since such a corner
contributed exactly `0.0` before.

Not fixed: the thickness itself. If a deposit with a **nonzero** weight ever
lands on a corner whose `dz_cell` stencil reaches outside the valid nodal
region, the normalization is still garbage -- and a nonzero numerator over a
zero thickness yields `inf` rather than a trapped NaN, which is worse because
it can propagate silently.

## Suggested directions

1. Fill the `z_phys_nd` ghost nodes outside the domain (extrapolate, or copy
   the periodic image where the direction is periodic), so the existing
   stencil always reads valid heights; or
2. Clamp `ic` and `jc` into the valid nodal range for the `dz` lookup only,
   the way `kc` is already clamped, so a ghost deposit borrows the thickness
   of the cell it will be folded into by `SumBoundary`.

Option 2 is local and cheap but changes answers wherever a nonzero-weight
ghost deposit currently uses ghost heights, so it needs a regression pass and
probably regenerated gold files for the affected SDM cases. Option 1 is the
more principled fix.

### Steps to reproduce

1.  must build with particles

### Expected behavior

this is what claude found

### Git commit hash

0808596ea273dcd7f87da2

### Environment

gcc

### Relevant logs or output

```shell
N/A
```

debog:
Clamps the horizontal indices of the CIC deposit thickness stencil into the valid nodal range, the way `kc` already was, so a ghost-cell deposit no longer normalizes by unfilled `z_phys_nd` ghost nodes. Fixes #3903, the half of the problem left open by the zero-weight guard added in #3902.

### DIFF
--- a/Source/Particles/ERFPCParticleToMesh.H
+++ b/Source/Particles/ERFPCParticleToMesh.H
@@ -34,6 +34,7 @@ void ERFPC::ERFPCParticleToMesh(amrex::MultiFab& a_mf,
     const int  k_max  = domain.bigEnd(AMREX_SPACEDIM-1);
     const int  i_lo   = domain.smallEnd(0), i_hi = domain.bigEnd(0);
     const int  j_lo   = domain.smallEnd(1), j_hi = domain.bigEnd(1);
+    const int  k_lo   = domain.smallEnd(AMREX_SPACEDIM-1);
     const bool per_x  = geom.isPeriodic(0);
     const bool per_y  = geom.isPeriodic(1);
 
@@ -164,9 +165,9 @@ void ERFPC::ERFPCParticleToMesh(amrex::MultiFab& a_mf,
                 // folds into; only the non-periodic exterior nodes need clamping.
                 const int ic_dz = per_x ? ic : amrex::max(i_lo, amrex::min(i_hi, ic));
                 const int jc_dz = per_y ? jc : amrex::max(j_lo, amrex::min(j_hi, jc));
-                const int kc_dz = amrex::max(0,  amrex::min(k_max, kc));
+                const int kc_dz = amrex::max(k_lo, amrex::min(k_max, kc));
                 const Real dz = dz_cell(ic_dz, jc_dz, kc_dz);
-                AMREX_ASSERT(dz > Real(0.0));
+                AMREX_ALWAYS_ASSERT(dz > Real(0.0));
                 const Real contrib = pval * w_horiz * w_z * inv_dxdy / dz;
                 Gpu::Atomic::AddNoRet(&rho(ic, jc, kc, a_comp), contrib);
             };
--- a/Source/Particles/ERFPCParticleToMesh.H
+++ b/Source/Particles/ERFPCParticleToMesh.H
@@ -34,6 +34,8 @@ void ERFPC::ERFPCParticleToMesh(amrex::MultiFab& a_mf,
     const int  k_max  = domain.bigEnd(AMREX_SPACEDIM-1);
     const int  i_lo   = domain.smallEnd(0), i_hi = domain.bigEnd(0);
     const int  j_lo   = domain.smallEnd(1), j_hi = domain.bigEnd(1);
+    const bool per_x  = geom.isPeriodic(0);
+    const bool per_y  = geom.isPeriodic(1);
 
     a_mf.setVal(0.0);
 
@@ -156,13 +158,13 @@ void ERFPC::ERFPCParticleToMesh(amrex::MultiFab& a_mf,
                 // an invalid-operation FPE.
                 if (w_horiz == Real(0.0) || w_z == Real(0.0)) { return; }
 
-                // Clamp only for the dz lookup; still write to the ghost rho cell. The
-                // z_phys_nd nodes outside the domain are not reliably filled, so a ghost
-                // deposit borrows the thickness of the outermost valid cell, which is the
-                // one SumBoundary folds it into.
-                const int ic_dz = amrex::max(i_lo, amrex::min(i_hi, ic));
-                const int jc_dz = amrex::max(j_lo, amrex::min(j_hi, jc));
-                const int kc_dz = amrex::max(0,    amrex::min(k_max, kc));
+                // Clamp only for the dz lookup; still write to the ghost rho cell. In a
+                // periodic direction the exterior nodes hold the wrapped image, so the
+                // unclamped stencil already gives the thickness of the cell SumBoundary
+                // folds into; only the non-periodic exterior nodes need clamping.
+                const int ic_dz = per_x ? ic : amrex::max(i_lo, amrex::min(i_hi, ic));
+                const int jc_dz = per_y ? jc : amrex::max(j_lo, amrex::min(j_hi, jc));
+                const int kc_dz = amrex::max(0,  amrex::min(k_max, kc));
                 const Real dz = dz_cell(ic_dz, jc_dz, kc_dz);
                 AMREX_ASSERT(dz > Real(0.0));
                 const Real contrib = pval * w_horiz * w_z * inv_dxdy / dz;
--- a/Source/Particles/ERFPCParticleToMesh.H
+++ b/Source/Particles/ERFPCParticleToMesh.H
@@ -32,6 +32,8 @@ void ERFPC::ERFPCParticleToMesh(amrex::MultiFab& a_mf,
     const Real dy  = geom.CellSize(1);
     const Box  domain = geom.Domain();
     const int  k_max  = domain.bigEnd(AMREX_SPACEDIM-1);
+    const int  i_lo   = domain.smallEnd(0), i_hi = domain.bigEnd(0);
+    const int  j_lo   = domain.smallEnd(1), j_hi = domain.bigEnd(1);
 
     a_mf.setVal(0.0);
 
@@ -154,9 +156,15 @@ void ERFPC::ERFPCParticleToMesh(amrex::MultiFab& a_mf,
                 // an invalid-operation FPE.
                 if (w_horiz == Real(0.0) || w_z == Real(0.0)) { return; }
 
-                // Clamp kc only for dz lookup; still write to ghost rho cell.
-                const int kc_dz = amrex::max(0, amrex::min(k_max, kc));
-                const Real dz = dz_cell(ic, jc, kc_dz);
+                // Clamp only for the dz lookup; still write to the ghost rho cell. The
+                // z_phys_nd nodes outside the domain are not reliably filled, so a ghost
+                // deposit borrows the thickness of the outermost valid cell, which is the
+                // one SumBoundary folds it into.
+                const int ic_dz = amrex::max(i_lo, amrex::min(i_hi, ic));
+                const int jc_dz = amrex::max(j_lo, amrex::min(j_hi, jc));
+                const int kc_dz = amrex::max(0,    amrex::min(k_max, kc));
+                const Real dz = dz_cell(ic_dz, jc_dz, kc_dz);
+                AMREX_ASSERT(dz > Real(0.0));
                 const Real contrib = pval * w_horiz * w_z * inv_dxdy / dz;
                 Gpu::Atomic::AddNoRet(&rho(ic, jc, kc, a_comp), contrib);
             };


asalmgren:
# Code review: PR #3906 — "Fix unfilled ghost-node thickness in particle deposition"

- **Author:** Debojyoti Ghosh (`debog`)
- **Branch:** `dg/fix_particle_deposit_dz` → `development`
- **Head commit reviewed:** `2ada53e3ffb7fd73d602140832d55344feed9639`
- **Scope:** 1 file, ~10 lines — `Source/Particles/ERFPCParticleToMesh.H`
- **Stated purpose:** clamp the horizontal indices of the CIC deposit thickness stencil into the
  valid nodal range, the way `kc` already was, so a ghost-cell deposit no longer normalizes by
  unfilled `z_phys_nd` ghost nodes. Fixes #3903, the half left open by #3902.

## Verification performed

- Built the PR branch (Release, particles ON) and ran the six affected regression tests —
  `ParticleAdvect`, `ParticleWoA`, `ParticleAdvect_AMR1_box`, `ParticleAdvect_AMR1_pcount`,
  `ParticleAdvect_AMR2_pcount`, `SDM_Bubble2D_WoA`. **All pass unchanged**, so no gold files under
  `Tests/ERFGoldFiles` need regeneration. That is not evidence the change is inert — it is because
  no test reaches the changed branch (finding 5).
- Also built the pre-PR header and ran a hand-made periodic-x Witch-of-Agnesi terrain case with
  `place_randomly_in_cells = true`. `amrex_fcompare` shows PR vs. baseline differ **only** in
  `tracer_particles_mass_density`, all other fields bit-identical — confirming the mechanism in
  finding 1.
- Repo left on `development` with `build/erf_exec` rebuilt to match.

## Summary judgement

The premise stated in the new comment — that the exterior `z_phys_nd` ghost nodes "are not
reliably filled" — holds only in **non-periodic** directions, and those are exactly the ghost cells
whose deposit `SumBoundary` discards. In **periodic** directions the ghost nodes *are* correctly
filled with the periodic image, and there the new clamp actively introduces an error in the
deposited density. The minimal fix is to gate each clamp on `!geom.isPeriodic(dir)`.

---

## 1. `Source/Particles/ERFPCParticleToMesh.H:163` — medium/high — the horizontal clamp corrupts `dz` in *periodic* directions, the only case whe
re a ghost deposit reaches the solution

`z_phys_nd` is filled with `FillBoundary(geom.periodicity())`
(`Source/ERF_MakeNewArrays.cpp:767`, `Source/Utils/ERF_TerrainMetrics.cpp:100,269`) with
`ngrow = ComputeGhostCells+2 >= 3`, so in a periodic direction the exterior ghost nodes hold the
correct periodic image. For a domain of `N` cells, `dz_cell(-1)` reads nodes `{-1,0} ≡ {N-1,N}`,
i.e. exactly `dz_cell(i_hi)` — and `a_mf.SumBoundary(geom.periodicity())` (line 175) folds ghost
cell `-1` into cell `i_hi`. The old code was already consistent. Symmetrically `dz_cell(i_hi+1)`
reads nodes `≡ {0,1}` = `dz_cell(i_lo)`, folded into cell `0`.

Clamping replaces both with the **near**-side thickness: a deposit that lands in cell `i_hi` is
normalized by `dz(i_lo)`.

**Failure scenario.** Periodic x (or y) with terrain that differs between the two ends of that
direction, and particles not sitting at cell centers (`place_randomly_in_cells = true`). The
deposited density is normalized by the wrong cell thickness, error `O(Δdz/dz)` — tens of percent
for realistic terrain. Confirmed empirically above: only `tracer_particles_mass_density` changed.

**Fix.** Gate the clamp, e.g.

```cpp
int ic_dz = ic, jc_dz = jc;
if (!is_per_x) { ic_dz = amrex::max(i_lo, amrex::min(i_hi, ic)); }
if (!is_per_y) { jc_dz = amrex::max(j_lo, amrex::min(j_hi, jc)); }
```

with `is_per_x/y` captured by value from `geom.isPeriodic(0/1)` outside the lambda. This is
option 1/2 in issue #3903 ("copy the periodic image where the direction is periodic").

---


## 2. `Source/Particles/ERFPCParticleToMesh.H:159` — medium — the new comment's justification is wrong for non-periodic directions, and the real conservation bug is untouched

The comment says the ghost deposit "borrows the thickness of the outermost valid cell, which is the
one `SumBoundary` folds it into". But `SumBoundary(geom.periodicity())` at line 175 is the only
`SumBoundary` on `a_mf` (verified: it is the sole `SumBoundary` call in `Source/`), and a
domain-exterior ghost cell in a **non-periodic** direction has no valid counterpart — its deposit is
silently dropped.

**Consequence.** Particle mass within half a cell of an inflow/outflow/wall boundary (and of
`zlo`/`zhi`) never appears in the deposited density, whichever `dz` is used. So in exactly the case
the clamp targets, the clamp changes only a value that is then discarded; in the periodic case
(finding 1) it changes a value that matters, for the worse.

**Fix.** Either reword the comment to drop the `SumBoundary` claim, or make the fix genuinely
conservative by folding the out-of-domain weight back into the outermost valid cell (deposit into
the clamped index rather than the ghost index) in non-periodic directions.

---

## 3. `Source/Particles/ERFPCParticleToMesh.H:165` — low — the same periodicity flaw applies to the pre-existing `kc` clamp, which also hardcodes `0`

`const int kc_dz = amrex::max(0, amrex::min(k_max, kc));` uses a literal `0` rather than
`domain.smallEnd(AMREX_SPACEDIM-1)`, inconsistent with the new `i_lo`/`j_lo`, and carries the same
periodic-image problem.

**Failure scenario.** `SDM_Box3D_IceAgg` runs with `geometry.is_periodic = 1 1 1`, so `kc = k_max+1`
is folded into cell `0`; with a stretched or `z_levels` vertical grid `dz(k_max) != dz(0)` and the
clamp supplies the wrong thickness. It is harmless today only because that test has uniform `dz`.
Since this PR is generalizing the clamp, handle `isPeriodic(2)` at the same time.

---

## 4. `Source/Particles/ERFPCParticleToMesh.H:167` — low — `AMREX_ASSERT(dz > Real(0.0))` gives no protection in the builds that matter

`AMREX_ASSERT` is compiled out in Release, which is this repo's CI/`CMAKE_BUILD_TYPE` default and
precisely where issue #3903's stated worry lives: "a nonzero numerator over a zero thickness yields
`inf` rather than a trapped NaN, which is worse because it can propagate silently". As written, the
`inf` still propagates silently in production runs.

**Fix.** Use `AMREX_ALWAYS_ASSERT`, or a runtime guard/clamp, if the check is meant to protect
production.

---

## 5. `Source/Particles/ERFPCParticleToMesh.H:172` — low (coverage note, no code defect) — the changed path is not exercised by any regression test

Every particle input in the suite sets `place_randomly_in_cells = false`, so particles sit exactly
at cell centers, `lx = ix+1` exactly, and `wx1 == wy1 == 0.0`. The `w_horiz == 0` early return added
in #3902 then skips every horizontal ghost corner, so `ic`/`jc` never leave `[i_lo,i_hi]` at deposit
time in CI. That is why all six affected tests pass bit-identically.

This is also exactly the corner from the issue report: particle at `x = 187.5`, `dx = 25` →
`ic = 8` with weight `0`. A test with `place_randomly_in_cells = true` near a lateral boundary would
be needed to lock in the new behavior — and would also expose finding 1.


---

## Checked and found correct

Recorded so they don't get re-reviewed:

- `i_lo`/`i_hi`/`j_lo`/`j_hi` are read from `geom.Domain()` on the host and captured by value into
  the device lambda — no host-pointer or `Box` capture problem.
- The zero-weight early return from #3902 is still first, so `dz_cell` is never evaluated for a
  corner that contributes nothing; the `0/0` FPE that motivated #3902 stays fixed.
- The `dz_cell` lambda's node stencil (`ic..ic+1`, `jc..jc+1`, `kc..kc+1`) is the correct
  node-averaged cell thickness, and the clamped indices keep all eight node reads inside the
  nodal box that `zheight` owns — so the PR does remove the genuine read-of-unfilled-node in the
  non-periodic case.
- `Gpu::Atomic::AddNoRet` still writes to the **unclamped** `(ic, jc, kc)`, so the clamp does not
  alter which cell receives mass — only the normalization. That is what makes finding 1 a value
  error rather than a misplacement.
- `kz_lo`/`kz_hi` and the `wz_lo`/`wz_hi` vertical weighting are untouched by this PR.

---

debog:
Thanks - 1 and 4 were right.

- 1: the clamp is now gated on `!geom.isPeriodic(dir)`.
- 2: comment reworded; it no longer claims the non-periodic ghost deposit survives `SumBoundary`.
- 3: `kc` now uses `domain.smallEnd` instead of the literal `0`, but I did not gate it on `isPeriodic(2)` - see below.
- 4: now `AMREX_ALWAYS_ASSERT`.
- 5: agreed. Running `SDM_Bubble2D_WoA` with `place_randomly_in_cells = true` gives 418240 nonzero-weight ghost deposits, so a regression test that reaches this path is possible if you want one.

One correction to the premise. The exterior nodes are not always the periodic image. In `SDM_Box3D_IceAgg` (`is_periodic = 1 1 1`, flat) node `z(9,j,0)` reads 100 where the wrapped value is 0, giving `dz_raw` of 0, -25 and -50 across 51200 ghost events - all with zero weight, so the early return from #3902 catches them. That is why the `kc` clamp is left ungated: gating it would expose those nodes where the clamp currently masks them. Happy to gate it too if you would rather have consistency.

Where the nodes are filled the review is right, and the unclamped stencil already gives the thickness of the cell `SumBoundary` folds into - in `SDM_Bubble2D_WoA` all 418240 ghost deposits have correctly filled nodes and `dz` is identical clamped or not.

All 40 particle/SDM tests are unchanged either way. The 4 failures (`SDM_Bubble2D_PerBoxInjection`, `SDM_SineMassFlux`, `SDM_FreezingShaft`, `SDM_RimingShaft`) reproduce on `development` with bit-identical diffs, so they are pre-existing and not from this PR.

PR #3907, opened 2026-08-24
Vertical Pressure Gradient

AMLattanzi:
# Summary
1. Use z_cc for vertical pressure gradient. This reduces the global memory access, is consistent with how lateral pressure gradients are computed, and reduces the number of variables passes around

## Expected changes
The compute metric at kface call **should** be equivalent since the nodes heights at `k` will cancel when computing the gradient. If this behavior is not observed then further investigation is needed.


Source/SourceTerms/ERF_MakeGradP.cpp
```diff
@@ -415,7 +411,7 @@ compute_gradp_z (const MultiFab& p,
 
             ParallelFor(tbz, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept
             {
-                Real met_h_zeta = (l_use_terrain_fitted_coords) ? Compute_h_zeta_AtKface(i, j, k, dxInv, z_nd_arr) : 1;
+                Real met_h_zeta = (l_use_terrain_fitted_coords) ? dxInv[2] * (z_cc_arr(i,j,k) - z_cc_arr(i,j,k-1)) : one;
```
asalmgren, line 414: **medium** — the "should be equivalent" claim holds only where `z_phys_cc` is still the 8-node average from `make_zcc`. It is not, on multi-level two-way-coupled terrain runs.

`average_down(*z_phys_cc[crse_lev+1], *z_phys_cc[crse_lev], ...)` overwrites the coarse-level `z_cc` under the fine grid with the fine-cell average — see `Source/ERF.cpp:596`, `Source/ERF_MakeNewLevel.cpp:355` and `Source/ERF_MakeNewLevel.cpp:654`. For non-linear terrain that average is *not* the coarse 8-node average, so `dxInv[2]*(z_cc(k)-z_cc(k-1))` no longer equals `Compute_h_zeta_AtKface`. Meanwhile `z_phys_nd` is never averaged down, and neither is `base_state`/`p_hse` (there is no `average_down` on `base_state` anywhere in `Source/`).

Concrete failure: `ERF::InitData_post` runs the average-down block at `ERF.cpp:590-603` and then, with `verbose > 0`, calls `compute_max_pressure_gradient_diagnostic(0)` at `ERF.cpp:1317`. That routine now forms `gpz` from the averaged `z_cc` while `p_hse` still reflects the un-averaged coarse geometry, and aborts if `|dp0/dz + rho0*g| > 1e-8` (`Source/ERF_Diagnostics.cpp:180-197`). Since `dp0/dz ~ rho*g ~ O(10) Pa/m`, even a `1e-9` *relative* error in the metric blows that tolerance — so a two-level terrain-fitted case with `coupling_type = TwoWay` can hard-`Abort("Base state is too far out of HSE")` where it passed before. The same substitution at line 655 perturbs the actual `gradp_type == 1` solve on coarse levels for the same reason.

Suggested fix: either keep the nodal metric here (it is the quantity that is guaranteed consistent with `p_hse`), or stop `average_down`-ing `z_phys_cc` / additionally average down the base state so the two stay consistent.

Source/SourceTerms/ERF_MakeGradP.cpp
```diff
@@ -659,7 +652,7 @@ compute_gradp_interpz (const MultiFab& p,
         [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept
         {
             // Note: identical to gradp_type == 0
-            Real met_h_zeta = (l_use_terrain_fitted_coords) ? Compute_h_zeta_AtKface(i, j, k, dxInv, z_nd_arr) : 1;
+            Real met_h_zeta = (l_use_terrain_fitted_coords) ? dxInv[2] * (z_cc_arr(i,j,k) - z_cc_arr(i,j,k-1)) : one;
```
asalmgren, line 655: **low** — stale ghost values now feed the vertical metric after a regrid.

In `Source/ERF_MakeNewLevel.cpp:355` and `Source/ERF_MakeNewLevel.cpp:654` the `average_down` of `z_phys_cc` is **not** followed by a `FillBoundary` (unlike `Source/ERF.cpp:601`, which does call it). So after a regrid a coarse box's `z_phys_cc` ghost layer still holds pre-`average_down` values while the neighbouring box's valid layer holds post-`average_down` values.

At the lowest `k` of each box, `z_cc_arr(i,j,k-1)` is exactly that ghost cell, so `met_h_zeta` is formed from a mismatched pair and `gpz` is wrong on the inter-box k-faces inside the refined footprint (and likewise at line 414). This path previously read `z_phys_nd`, which is never averaged down and so was always self-consistent.

Suggested fix: add `z_phys_cc[crse_lev]->FillBoundary(geom[crse_lev].periodicity());` after the two `average_down` loops in `ERF_MakeNewLevel.cpp`, mirroring what `ERF.cpp:601` already does.

asalmgren (commented):
Code review of the `z_phys_nd` -> `z_phys_cc` switch for the vertical pressure gradient.

The algebra checks out exactly, not just approximately: `z_cc(k) - z_cc(k-1) = 0.125*[sum_4 z_nd(k+1) - sum_4 z_nd(k-1)]`, which is precisely `Compute_h_zeta_AtKface * dz`. Ghost reach is also safe since `tbz` is shrunk away from `domain_klo` / `domain_khi+1`, and all six call sites were updated.

The equivalence holds **only where `z_phys_cc` is still the 8-node average produced by `make_zcc`**. Two comments below on where that assumption breaks.

Source/SourceTerms/ERF_MakeGradP.cpp
AMLattanzi: All of the base state quantities are averaged down in `ERF.cpp` and `ERF_MakeNewLevel.cpp`. Additionally, a fillboundary call is added for the base state to follow what is proposed in the comment below.

Source/SourceTerms/ERF_MakeGradP.cpp
AMLattanzi: Fillboundary of the z_phys_cc was added to `ERF_MakeNewLevel.cpp`.

PR #3911, opened 2026-08-25
Add large scale forcing docs

ckendrick:
This updates the docs to add more details about the large scale forcing option and fixes the time used for interpolating when a date time is used.

- Add a section describing the added data log for lsf and nudging terms
- Add a section describing the large scale forcing and nudging options with an example text file.
- Fix the time used in the large scale forcing interpolation when start time is used (e.g., start_datetime or real data case)
- Fix doc formatting issues added from sfc/sst MOST forcing option.
- Clean up large scale forcing file with more doxygen comments.
- Add print out of the lsf grid when ERF is ran in verbose mode.

### DIFF
--- a/Docs/sphinx_doc/Inputs.rst
+++ b/Docs/sphinx_doc/Inputs.rst
@@ -1677,6 +1677,53 @@ with an ``erf.abl_geo_wind_table``.
   or ``-DERF_ENABLE_WINDFARM`` (cmake) at build time.
   See :ref:`sec:WindFarmModels` for theory and examples.
 
+
+Large Scale Forcing
+--------------------
+
+Large-scale forcing can be used to prescribe time-varying vertical profiles for
+temperature, water vapor, horizontal wind, and vertical subsidence.
+A text file containing the large scale forcing data is provided using the ``erf.large_scale_forcing_file``
+option. The file contains one or more time blocks, where each time block is denoted with
+a header containing the day, number of levels in the block, and reference pressure for that time.
+Each row of data in the block contains ``z, p, tls, qls, uls, vls, wls`` corresponding to level
+height, pressure, large scale temperature and moisture tendency, large scale zonal and meridonal velocity,
+and large scale vertical subsidence velocity.
+
+- Either a height grid ``z[m]`` or pressure grid ``p[mb]`` can be provided, where one can be a negative number
+  to indicate the other grid type should be used (e.g. if the ``z`` values are negative, the file is
+  interpreted as pressure-level data instead of height-level data). If both grids are provided, then the ``z`` grid is used.
+
+- The large scale forcing grid is interpolated to the ERF grid, similar to the input sounding.
+
+- Large scale tendencies are applied based on a time period in seconds specified by the ``erf.forcing_timescale`` option.
+
+- Times are converted to elapsed seconds relative to the first timestamp in the file.
+
+- Temperature and moisture nudging can be restricted to a custom interval instead of entire domain with the
+  ``erf.nudging_t_z1`` and ``erf.nudging_t_z2``, and ``erf.nudging_q_z1`` and ``erf.nudging_q_z2`` options, respectively.
+
+- **NOTE:** When both large scale forcing and ``erf.nudging_from_input_sounding`` are used, the u and v velocities are nudged
+  according to the observed large scale velocity ``uls`` and ``vls`` instead of the input sounding (using the lsf forcing timescale).
+  Temperature and moisture are nudged from input sounding as normal (using the input sounding tau_nudging timescale).
+
+Example file format for large scale forcing:
+
+.. code-block:: text
+
+   z[m] p[mb]  tls[K/s] qls[kg/kg/s] uls  vls  wls[m/s]
+   0.0,   4, 1000.0  day, levels, pres0
+      0.0  1000.00  0.000e+00  0.000e+00  0.0  0.0  0.000e+00
+    100.0   990.00  0.000e+00  0.000e+00  0.5  0.2  0.000e+00
+   1000.0   900.00 -5.000e-06  0.000e+00  2.0  1.0  0.000e+00
+   5000.0   500.00 -1.000e-05  0.000e+00  6.0  3.0  0.000e+00
+   0.5,   4, 1000.0  day, levels, pres0
+      0.0  1000.00  0.000e+00  0.000e+00  0.2  0.1  0.000e+00
+    100.0   990.00  0.000e+00  0.000e+00  0.7  0.3  0.000e+00
+   1000.0   900.00 -2.500e-06  0.000e+00  2.5  1.2  0.000e+00
+   5000.0   500.00 -8.000e-06  0.000e+00  6.5  3.2  0.000e+00
+
+
 Boundary Plane I/O (Coupling Support)
 =====================================
 
--- a/Docs/sphinx_doc/Inputs.rst
+++ b/Docs/sphinx_doc/Inputs.rst
@@ -918,9 +918,9 @@ The requested output files have the following columns:
 
   #. Height (m)
 
-  #. Large-scale temperature tendency, :math:`\partial \theta / \partial t` (K/s)
+  #. Total applied temperature tendency, :math:`\partial \theta / \partial t` (K/s)
 
-  #. Large-scale water vapor tendency, :math:`\partial q_v / \partial t` (kg/kg/s)
+  #. Total applied water vapor tendency, :math:`\partial q_v / \partial t` (kg/kg/s)
 
   #. Large-scale subsidence velocity, :math:`w_{sub}` (m/s)
 
@@ -941,6 +941,11 @@ The requested output files have the following columns:
   #. X-velocity nudging tendency, :math:`u_{nudge}` (m/s2)
 
   #. Y-velocity nudging tendency, :math:`v_{nudge}` (m/s2)
+
+.. note::
+
+   The vertical tendency columns are only populated for interior cells between
+   the bottom and top boundaries; they are zero at the domain edges.
 
 
 Data Sampling Outputs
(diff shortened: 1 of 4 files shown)


Docs/sphinx_doc/Inputs.rst
```diff
@@ -1646,6 +1677,53 @@ with an ``erf.abl_geo_wind_table``.
...
+height, pressure, large scale temperature and moisture tendency, large scale zonal and meridonal velocity,
+and large scale vertical subsidence velocity.
+
+- Either a height grid ``z[m]`` or pressure grid ``p[mb]`` can be provided, where one can be a negative number
+  to indicate the other grid type should be used (e.g. if the ``z`` values are negative, the file is
+  interpreted as pressure-level data instead of height-level data). If both grids are provided, then the ``z`` grid is used.
+
+- The large scale forcing grid is interpolated to the ERF grid, similar to the input sounding.
+
+- Large scale tendencies are applied based on a time period in seconds specified by the ``erf.forcing_timescale`` option.
```
asalmgren: `erf.forcing_timescale` doesn't do this. `tau_lsf` appears in exactly two places, both annotated by their own comment as `// only applies to u,v LSF nudging` (`ERF_AdvanceDycore.cpp:448`, `ERF_MakeMomSources.cpp:708`). The large-scale θ, qv and subsidence tendencies are applied directly over the timestep with no relaxation timescale at all:

```cpp
cell_data(i,j,k,RhoTheta_comp) += cell_data(i,j,k,Rho_comp) * lsf_arr(i,j,k,0) * dt_advance;
```

(`ERF_AdvanceDycore.cpp:344-350`)

As written, a reader would reasonably conclude that lowering `forcing_timescale` strengthens the temperature and moisture forcing — it does nothing to them. Suggested replacement:

> `erf.forcing_timescale` is the relaxation time scale used when nudging the horizontal velocities toward the `uls`/`vls` profiles. The temperature, moisture and subsidence tendencies read from the file are applied directly and are not scaled by it. It must be positive whenever `erf.nudging_from_input_sounding` is also set.

(The assert enforcing that last sentence is at `ERF.cpp:908`.)

Docs/sphinx_doc/Inputs.rst
```diff
@@ -1646,6 +1677,53 @@ with an ``erf.abl_geo_wind_table``.
   or ``-DERF_ENABLE_WINDFARM`` (cmake) at build time.
   See :ref:`sec:WindFarmModels` for theory and examples.
 
+
+Large Scale Forcing
+--------------------
+
+Large-scale forcing can be used to prescribe time-varying vertical profiles for
```
asalmgren, line 1689: This section never mentions the flag that actually turns the feature on. The whole block — read, interpolate, apply — is gated on `solverChoice.large_scale_forcing` (`ERF.cpp:905`), so setting only `erf.large_scale_forcing_file` does nothing at all.

Since this is now the narrative how-to, it would help to open with the full minimal set:

```
erf.large_scale_forcing      = true
erf.large_scale_forcing_file = lsf.txt
erf.forcing_timescale        = 3600.0
```

The flag is in the parameter table at line 1524, but a reader landing on this section won't necessarily scroll back to it.

Docs/sphinx_doc/Inputs.rst
```diff
@@ -911,6 +912,36 @@ The requested output files have the following columns:
 
   #. SGS turbulence dissipation, :math:`\epsilon` (m2/s3)
 
+* Large scale forcing and nudging profiles
+
+  #. Time (s)
+
+  #. Height (m)
+
+  #. Large-scale temperature tendency, :math:`\partial \theta / \partial t` (K/s)
```
asalmgren: Columns 3 and 4 are the *total applied* tendency, not just the large-scale part. `lsf_arr` comps 0 and 1 accumulate both pieces: comps 3/4 are set to the horizontal part, then comps 0/1 have `tvtend`/`qvtend` added (`ERF_AdvanceDycore.cpp:302-303` and `:330-331`). So column 3 = column 6 + column 8 and column 4 = column 7 + column 9, and columns 3/4 are what is actually applied to the state.

"Total applied temperature tendency" / "Total applied water vapor tendency" would remove the ambiguity against the separate Horizontal and Vertical entries below.

Also worth a note: comps 5/6/7 are only written for `kmin <= k <= kmax`, so the vertical columns are identically zero in the bottom and top cells.

Source/DataStructs/ERF_LargeScaleForcingData.H
```diff
@@ -276,16 +339,19 @@ public:
...
         // helper function to get the current time indices of forcing data to use and
-        // coefficients based on the current simulation time
+        // coefficients based on the current simulation time. Historical callers
+        // pass elapsed time, while datetime-aware paths may pass total/epoch
+        // time, so normalize to elapsed time relative to the simulation start.
         curr = 0;
         next = 0;
         coeff_curr = one;
         coeff_next = zero;
+        const amrex::Real rel_time = (time >= start_time) ? (time - start_time) : time;
```
asalmgren: The fix is right, but this leaves `start_time` dead, and the guard that makes it dead is magnitude-based.

Every caller passes elapsed time, so this evaluates to `time` in both branches: when `start_time == 0` the subtraction is a no-op, and when `start_time` is an epoch value (~1.7e9 s, i.e. 54 years) elapsed time can never reach it. So the ternary, the `start_time` member, the assignment at `ERF.cpp:912`, and the comment above claiming "datetime-aware paths may pass total/epoch time" all describe a caller that doesn't exist.

Not wrong today, but fragile in a specific way: the branch is selected by the *magnitude* of the argument, so a future caller that genuinely passed absolute time would get different behavior depending on how far into the run it is. Deleting `start_time` from `LargeScaleForcingData` entirely and using `time` directly would be shorter, would put the "callers pass elapsed time" contract in the signature comment where it belongs, and would remove the trap.

Source/DataStructs/ERF_LargeScaleForcingData.H
```diff
@@ -173,9 +192,30 @@ public:
         AMREX_ALWAYS_ASSERT_WITH_MESSAGE(z_lsf.size() == num_times,
                                          "repeated timestamp in the large-scale forcing file");
         amrex::Print() << "  Read " << nz_lsf << " levels at " << num_times << " times" << std::endl;
+
+        // debugging print
+        if (verbose_print) {
+            for (int i = 0; i < times_lsf.size(); i++)
```
asalmgren: Nit (applies to the matching block at line 312 too): this loop always breaks at the end of its first iteration, so it's really `if (times_lsf.size() > 0) { const int i = 0; ... }` — which says the same thing without making the reader hunt for the `break`.

The `int i` vs unsigned `size()` comparison will also draw `-Wsign-compare`. Harmless, since `-Werror` is commented out in every CI workflow.

Bounds are fine, for what it's worth: `nz_lsf + 1` matches the asserted block size at `:97`/`:163`, and the interpolated loop's `k <= Nz` matches the `resize(Nz+1)` at `:210-215`.

asalmgren (commented):
**Verdict: no blockers.** The functional change is correct and fixes a real bug — with `erf.start_datetime` set, every LSF lookup was previously off by an epoch offset of ~1.7e9 s and therefore always clamped to the first time block. The `SurfaceLayer.rst` edits are genuine rendering fixes, not cosmetics: the bare `..` before the two input snippets is an RST *comment*, so those two configuration blocks were being silently dropped from the rendered page entirely, and `Notes` underlined with `-----` was being promoted to a top-level section (the `^^^^^` fix is right).

A few inline comments below — two doc statements that would mislead someone setting this up for the first time, and some cleanup suggestions.

### What I checked

- All three call sites pass **elapsed** time (`ERF_AdvanceDycore.cpp:270`, `:450` pass `old_time = t_old[level]`; `ERF_MakeMomSources.cpp:710` gets `old_stage_time`/`old_substep_time`), so the new `rel_time` is correct.
- The new debug prints are in bounds: the raw loop runs to `nz_lsf + 1` and every pushed block is asserted to be exactly that size (`:97`, `:163`); the interpolated loop runs `k <= Nz` and all six arrays are `resize(Nz+1)`.
- The 14 documented data-log columns match the write order in `ERF_Write1DProfiles.cpp:184-190` and `ERF_Write1DProfiles_stag.cpp:290-296` exactly, and the units are right against the fill sites (`lsf_arr` comps 0/3/5 K/s, 1/4/6/7 kg/kg/s; `nudge_arr` comps 0/1 K/s and kg/kg/s, 2/3 m/s²).
- The example forcing file actually parses: 7-column header, 6-column block headers, `std::stod("0.0,")`/`std::stoi("4,")` both succeed on the trailing commas, and `nlev = 4` with 4 data rows satisfies the `z_in.size() == nz_lsf + 1` asserts.
- Every parameter name in the new prose exists in the code.
- Both changed `.rst` files lint clean under docutils across the edited ranges.

### Pre-existing, not raised against this PR

- **Pressure-level forcing files require an input sounding, unguarded.** The new bullet presents the negative-`z` pressure-grid mode as a first-class option, but `interp_forcing` reads `sounding.pm_integ[k]`, which is only sized/filled by `InputSoundingData::calc_rho_p`/`calc_rho_p_hse`. A pressure-level LSF file with no input sounding indexes an empty `Vector`. Worth a sentence in the doc and ideally an `AMREX_ALWAYS_ASSERT`.
- **The staggered writer plots cell-centered profiles against face heights.** `derive_forcing_profiles_stag` returns `nz` cell-centered values regardless of its name, and both writers call it. `ERF_Write1DProfiles.cpp:180-184` uses cell centers correctly, but `ERF_Write1DProfiles_stag.cpp:291-292` uses `zlevels_stag[0][k]` — the *lower face* of cell `k`. The new doc describes column 2 as just "Height (m)", which is a half-cell off in the staggered output.
- **The file's first timestamp is assumed to be the simulation start.** The bullet "Times are converted to elapsed seconds relative to the first timestamp in the file" is accurate, but for a `start_datetime` run it means the day values are never compared against the simulation start date — a file beginning six hours early is silently applied from `t = 0`. Since this PR is specifically about the datetime case, that caveat belongs right next to that bullet.
- **No test exercises this path.** No `Exec/` or `Tests/` input enables `erf.large_scale_forcing`; the only mention is commented out at `Tests/test_files/ABL_MOST_SST/ABL_MOST_SST.i:44-46`. Uncommenting it (with a small `lsf.txt` and a fifth `erf.data_log` entry) would give both this fix and the new data log some coverage.

ckendrick:
@asalmgren The issues should be resolved now - the docs were updated to correctly describe the `erf.forcing_timescale` option and provide a full example of enabling the large scale forcing (flag, text file, and time scale). The unused `start_time` path in the large scale forcing struct was also removed and the debugging print loops were updated.

PR #3920, opened 2026-08-25
Issue 3912

### LINKED ISSUE #3912: [Bug]: Native Shoc Momentum host_diffusion
# Native SHOC: blanket vertical-implicit disable

## Summary

Native SHOC run with `erf.shoc.transport_mode = state_update` and
`erf.shoc.momentum_transport = host_diffusion` (the documented default pairing)
is unstable. The same case is stable with
`erf.shoc.momentum_transport = state_update`.

In the mixed mode, the
scalars get an unconditionally stable implicit column solve while momentum gets
a fully explicit host diffusion using SHOC's `K_m`, with **no implicit option
available because the vertical implicit solve is disabled for the whole SHOC
family.**

This issue proposes (a) a short-term fix that re-enables the host vertical
implicit solve for the momentum-only export, and (b) a third transport mode in
which SHOC's implicit increment is handed to the dycore as a tendency added to
the RHS in each RK stage, rather than as a pre-dycore state update.

### The vertical implicit solve is off for all SHOC configurations

`Source/DataStructs/ERF_DataStruct.H`:

```cpp
// Implicit vertical diffusion (not available with SHOC family)
if (uses_shoc_family()) {
    amrex::Print() << "Turning off native vertical implicit solve at all levels "
                      "with SHOC family PBL scheme." << std::endl;
    for (int lev = 0; lev <= max_level; lev++) {
        vert_implicit_fac[lev][0] = zero;
        vert_implicit_fac[lev][1] = zero;
        vert_implicit_fac[lev][2] = zero;
    }
}
```

This is unconditional on the transport mode. In the mixed mode,
`ShocDriver::set_eddy_diffs()` exports only `EddyDiff::Mom_v`:

```cpp
} else if (uses_momentum_host_diffusion()) {
    MultiFab::Copy(*m_eddy_diffs_ptr, m_eddy_coeffs_cc,
                   EddyDiff::Mom_v, EddyDiff::Mom_v, 1, 0);
}
```

so the host diffuses momentum explicitly, subject to `K_m * dt / dz^2 <~ 1/2`.
SHOC's convective `K_m` (order 10^2 m^2/s) against near-surface `dz` of 10-20 m
exceeds that at typical `dt`.

### Why `momentum_transport = state_update` is stable

`col.u` and `col.v` are included in the same tridiagonal solve as the scalars
(`ShocImplicit::advance_implicit_state`, `Source/PBL/Shoc/ERF_ShocImplicit.cpp`),
so that path has no diffusive stability limit. The two working modes differ in
whether momentum is treated implicitly, which is consistent with an explicit
diffusive CFL violation rather than a coupling bug.

## Proposal A (short term): allow the host vertical implicit solve in mixed mode

Narrow the `uses_shoc_family()` zeroing so it applies only to the components
SHOC actually owns. In mixed mode SHOC exports only `Mom_v`, so the host
implicit solve would handle exactly the term that is going unstable, and unlike
a frozen tendency it uses the current-stage state.

Suggested settings once the gate is relaxed:

```text
erf.vert_implicit_fac = 1.0          # default 1.0 1.0 0.0 leaves stage 3 explicit
erf.implicit_momentum_diffusion = true
erf.implicit_thermal_diffusion  = false
erf.implicit_moisture_diffusion = false
erf.implicit_ke_diffusion       = false
```

### Steps to reproduce

N/A

### Expected behavior

N/A

### Git commit hash

N/A

### Environment

N/A

### Relevant logs or output

```shell

```

AMLattanzi:
# Summary
This PR resolves #3912


Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1188,17 +1189,50 @@ struct SolverChoice {
...
         if (uses_shoc_family()) {
-            amrex::Print() << "Turning off native vertical implicit solve at all levels with SHOC family PBL scheme." << std::endl;
-            for (int lev = 0; lev <= max_level; lev++) {
-               vert_implicit_fac[lev][0] = zero;
-               vert_implicit_fac[lev][1] = zero;
-               vert_implicit_fac[lev][2] = zero;
-            }
+            bool uses_host_diffusion = ( host_owns_vertical_scalar_diffusion() ||
+                                         host_owns_vertical_momentum_diffusion() );
+            if (use_native_shoc && uses_host_diffusion) {
```
asalmgren: **High -- the anelastic guard is bypassed on this path.**

The `else` branch below zeroes `vert_implicit_fac` on any level with `anelastic[lev]` ("Turning off native vertical implicit solve flag because anelastic at level ..."), but this new branch never applies that check. Since `shoc_momentum_transport` defaults to `HostDiffusion`, *any* run with `erf.anelastic = 1` plus native SHOC now takes this branch and gets `vert_implicit_fac = (1,1,0)` with `implicit_momentum_diffusion = true`.

Worse, the anelastic MRI loop (`Source/TimeIntegration/ERF_MRI.H:364`) only runs `nrk = 0,1` -- exactly the two stages weighted `1.0` -- so `ERF_ImplicitPre.H` fires on *every* anelastic stage, mutating `rho`, `rhotheta`, and the momenta before the projection, on a code path the rest of ERF explicitly declares unsupported.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1188,17 +1189,50 @@ struct SolverChoice {
...
+                for (int lev = 0; lev <= max_level; lev++) {
+                    vert_implicit_fac[lev][0] = one;
```
asalmgren: **Medium -- the EB guard is bypassed too.**

`ERF_DataStruct.H:1309` disables the implicit solve for `terrain_type == TerrainType::EB` ("Implicit diffusion is not supported with EB; turning off at level ..."), but that check also lives only in the `else` branch. So `erf.terrain_type = EB` + native SHOC with the default `momentum_transport = host_diffusion` now enables the implicit momentum solve on EB, which the code elsewhere states is unsupported.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1188,17 +1189,50 @@ struct SolverChoice {
...
+                implicit_ke_diffusion       = host_owns_vertical_scalar_diffusion();
+            } else {
+                amrex::Print() << "Turning off native vertical implicit solve at all levels with SHOC family PBL scheme." << std::endl;
+                for (int lev = 0; lev <= max_level; lev++) {
+                    vert_implicit_fac[lev][0] = zero;
+                    vert_implicit_fac[lev][1] = zero;
+                    vert_implicit_fac[lev][2] = zero;
+                }
+            } // eamxx shoc
         } else {
```
asalmgren: **Medium -- no user escape hatch for the newly-enabled implicit solve.**

`pp.query("vert_implicit", ...)`, `pp.countval("vert_implicit_fac")`, `pp.query("implicit_thermal_diffusion", ...)`, `implicit_moisture_diffusion`, `implicit_ke_diffusion`, `implicit_momentum_diffusion`, and `implicit_before_substep` are all read inside this `else` branch only.

Under native SHOC the implicit solve is now on by default, yet a user who sets `erf.vert_implicit = false` (or any `vert_implicit_fac` / `implicit_*_diffusion` override) has it silently ignored and still gets the implicit solve -- the opposite of what they asked for, with no warning. It also means the anelastic and EB cases above cannot be worked around from the inputs file.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1188,17 +1189,50 @@ struct SolverChoice {
...
 
         // Implicit vertical diffusion (not available with SHOC family)
         if (uses_shoc_family()) {
-            amrex::Print() << "Turning off native vertical implicit solve at all levels with SHOC family PBL scheme." << std::endl;
-            for (int lev = 0; lev <= max_level; lev++) {
-               vert_implicit_fac[lev][0] = zero;
-               vert_implicit_fac[lev][1] = zero;
-               vert_implicit_fac[lev][2] = zero;
-            }
+            bool uses_host_diffusion = ( host_owns_vertical_scalar_diffusion() ||
```
asalmgren: **Low -- silent default-behavior change.**

The `else` branch prints "Turning off native vertical implicit solve at all levels...", but this branch prints nothing. Because `ShocMomentumTransport::HostDiffusion` is the default, every existing native-SHOC run silently flips from fully explicit to `vert_implicit_fac = (1,1,0)` with `implicit_momentum_diffusion = true` and `implicit_thermal/moisture/ke_diffusion = false`.

A `Print()` describing the chosen configuration would make the change visible in the run log -- `display()` only prints the numbers, not the reason.

asalmgren (commented):
## Code review

This refactor (moving the SHOC transport enums into `ERF_ShocTransportStruct.H` and parsing `erf.shoc.*` in `SolverChoice::init_params`) looks clean: the `erf.shoc` prefix matches, the defaults match the old `ShocRuntimeOptions` defaults, `ShocDriver` is the only consumer of `read_shoc_runtime_options`, and the transitive include chain `ERF_ShocTypes.H -> ERF_DataStruct.H -> ERF_ShocTransportStruct.H` keeps everything compiling.

The behavior change is the new `vert_implicit_fac = (1,1,0)` path for native SHOC with host diffusion. The `(1,1,0)` weighting itself is *not* a stability problem -- the composite amplification factor `1 + z/((1-z/3)(1-z/2))` stays in (0,1) for all `z = dt*lambda < 0`, so the scheme is unconditionally stable even though the final stage is explicit. My concerns are about what the new branch *skips* relative to the `else` branch; see the inline comments.

### Additional finding (outside the diff)

**`Docs/sphinx_doc/Inputs.rst:2933` -- documentation now stale.** The "Implicit Vertical Diffusion" section states the implicit solve "is turned off automatically if there is no molecular or turbulent diffusion, if the level is anelastic, if `terrain_type` = `EB`, **or if a SHOC-family PBL scheme is active**." That last clause is no longer true for native SHOC with host diffusion. The section should also note that the `erf.vert_implicit*` / `erf.implicit_*` parameters are ignored under a SHOC-family scheme. `Docs/sphinx_doc/theory/PBLschemes.rst:311` (the `momentum_transport = host_diffusion` description) would likewise benefit from stating that this mode now enables the host implicit solve.

### Minor notes

- `use_native_shoc &&` in the new condition is redundant: `host_owns_vertical_*_diffusion()` already return `false` whenever `use_eamxx_shoc` is true, so `uses_host_diffusion` can only be true when native SHOC is active. Harmless, but it obscures the logic.
- The `implicit_*_diffusion` flags are global while `turbChoice` is per-level, so in a mixed multi-level config (e.g. MYNN on level 0, native SHOC on level 1) the non-SHOC level also has its implicit thermal/moisture/KE diffusion decided by the SHOC level. Not a regression (previously it lost everything), but a latent per-level correctness gap.
- `ERF_ShocTypes.H` dropped its `<string>` / `<algorithm>` / `<cctype>` includes and now gets them transitively through `ERF_DataStruct.H -> ERF_ShocTransportStruct.H`, even though it still uses `std::string` directly. It compiles, but a direct include would be more robust.

Source/PBL/Shoc/ERF_ShocDriver.cpp
```diff
@@ -348,6 +348,9 @@ ShocDriver::ShocDriver (int lev, const SolverChoice& solver_choice)
       m_moisture_type(solver_choice.moisture_type),
       m_moisture_indices(solver_choice.moisture_indices)
 {
+    m_opts.transport_mode     = solver_choice.shoc_transport_mode;
```
asalmgren, line 351: **HIGH — the SHOC unit tests can no longer set the transport modes; four tests will fail.**

The ctor now takes `transport_mode` / `momentum_transport` from `solver_choice`, and `read_shoc_runtime_options()` no longer reads `erf.shoc.transport_mode` / `erf.shoc.momentum_transport` at all — that parsing moved into `SolverChoice::ReadParameters`.

But `Tests/Unit/Shoc/ERF_ShocDriverTests.cpp` sets those keys via `ScopedParmParseString` and then constructs a **default-constructed** `SolverChoice` (it never runs `ReadParameters`), so the driver silently gets the defaults `StateUpdate` / `HostDiffusion`. Concretely:

- `MomentumTransportNoneLeavesVelocitiesUntouched…` (line 1093) — `EXPECT_TRUE(driver.disables_momentum_transport())` fails
- `MomentumTransportStateUpdateUpdatesFaceVelocities…` (line 1166) — `EXPECT_TRUE(driver.uses_momentum_state_update())` fails
- `HostDiffusionModeExportsDiffusivitiesWithoutInternalTransport` (line 1656) — `EXPECT_FALSE(driver.uses_state_update())` fails
- `MultiStepHostDiffusionModeKeepsExportsStable` (line 2027) — same

Fix either way: keep the ParmParse fallback in `read_shoc_runtime_options` (only overriding from `solver_choice` when the ParmParse key is absent), or update the tests to set `solver_choice.shoc_transport_mode` / `shoc_momentum_transport` directly.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1188,115 +1189,155 @@ struct SolverChoice {
...
+            }
+            if (!host_owns_momentum && user_set_imp_momentum && implicit_momentum_diffusion) {
+                amrex::Print() << "Ignoring erf.implicit_momentum_diffusion = true; "
+                               << "SHOC owns vertical momentum diffusion." << "\n";
+            }
+
+            implicit_thermal_diffusion  = (implicit_thermal_diffusion  && host_owns_scalars);
+            implicit_moisture_diffusion = (implicit_moisture_diffusion && host_owns_scalars);
+            implicit_ke_diffusion       = (implicit_ke_diffusion       && host_owns_scalars);
+            implicit_momentum_diffusion = (implicit_momentum_diffusion && host_owns_momentum);
```
asalmgren, line 1275: **MEDIUM — silent default-on of the implicit momentum solve for every native-SHOC run.**

With the documented default pairing (`transport_mode = state_update`, `momentum_transport = host_diffusion`), `host_owns_momentum` is true, `implicit_momentum_diffusion` keeps its default `true`, and `vert_implicit_fac` keeps its default `{1,1,0}`. So an existing input deck that previously ran fully explicit now runs the tridiagonal momentum solve, allocates `Tau_corr`, and produces different answers **with no input change**. Similarly `transport_mode = host_diffusion` now switches on implicit thermal/moisture/KE.

Issue #3912's Proposal A asked users to *opt in* via `erf.implicit_momentum_diffusion = true`. The in-repo SHOC regression tests all use `momentum_transport = state_update`, so CI will not catch this change.

Consider requiring an explicit opt-in, or at minimum calling this out in the PR description / release notes.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1188,115 +1189,155 @@ struct SolverChoice {
             }
             if (turbChoice[lev].uses_native_shoc()) {
                 use_native_shoc = true;
+                amrex::ParmParse pp_shoc("erf.shoc");
```
asalmgren: **MEDIUM — the SHOC transport mode and the ownership flags are global, but `pbl_type` is per level.**

`shoc_transport_mode` / `shoc_momentum_transport` are single scalars parsed inside a `for (lev …)` loop, and `host_owns_vertical_*_diffusion()` short-circuits on `use_eamxx_shoc` (true if *any* level uses EAMXX SHOC) before consulting the native mode.

Scenario: level 0 = `NATIVE_SHOC` with `momentum_transport = host_diffusion`, level 1 = `EAMXX_SHOC` (or vice versa). `use_eamxx_shoc` is true, so `host_owns_vertical_momentum_diffusion()` returns false and `implicit_momentum_diffusion` is forced off on *all* levels — including level 0, which is exactly the level this PR is trying to stabilize. The same applies to a mixed native-SHOC / non-SHOC hierarchy: the non-SHOC level loses its implicit scalar solve.

Since `implicit_*_diffusion` are already per-solver rather than per-level this may be acceptable, but the ownership test should at least be evaluated per level, or the mixed configuration should be rejected.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1978,6 +2023,28 @@ struct SolverChoice {
         return use_eamxx_shoc || use_native_shoc;
     }
 
+    /**
+     * @brief Test whether the host owns momentum diffusion.
+     * @return True if ERF dycore does vertical diffusion of momentum.
+     */
+    bool host_owns_vertical_momentum_diffusion () const noexcept
```
asalmgren, line 2050: **LOW/MEDIUM — `momentum_transport = none` is treated as "SHOC owns momentum diffusion", which is false.**

This returns `shoc_uses_momentum_host_diffusion(...)`, i.e. false for `None`. But in `none` mode `ShocDriver::set_eddy_diffs()` exports nothing and SHOC performs no momentum transport at all — the host is the *only* party doing vertical momentum diffusion (molecular / any non-SHOC contribution).

Scenario: `erf.pbl_type = NATIVE_SHOC`, `erf.shoc.momentum_transport = none`, `erf.molec_diff_type = Constant` with a large viscosity, `erf.implicit_momentum_diffusion = true`. The request is silently discarded and the run prints *"Ignoring erf.implicit_momentum_diffusion = true; SHOC owns vertical momentum diffusion"*, which is factually wrong for this mode.

`none` should either keep the host implicit solve available, or the message should be reworded.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1188,115 +1189,155 @@ struct SolverChoice {
...
+            implicit_thermal_diffusion  = (implicit_thermal_diffusion  && host_owns_scalars);
+            implicit_moisture_diffusion = (implicit_moisture_diffusion && host_owns_scalars);
+            implicit_ke_diffusion       = (implicit_ke_diffusion       && host_owns_scalars);
+            implicit_momentum_diffusion = (implicit_momentum_diffusion && host_owns_momentum);
+
+            // Make the resulting configuration visible in the run log
+            if (!host_owns_scalars && !host_owns_momentum) {
+                amrex::Print() << "Turning off vertical implicit solve for all components since SHOC owns diffusion." << "\n";
+            } else {
+                amrex::Print() << "SHOC hands vertical diffusion of"
```
asalmgren, line 1281: **LOW — run-log messages are emitted before the checks that can invalidate them.**

This "keeping the vertical implicit solve on" message is printed here, but `erf.vert_implicit = false` (line 1296), the anelastic check (1307), the EB check (1317) and the all-components-off check (1329) all zero `vert_implicit_fac` afterwards.

With `erf.pbl_type = NATIVE_SHOC` plus `erf.vert_implicit = false`, the log claims the solve is kept on and then immediately says it is turned off; with SHOC owning everything, two nearly-identical "turning off" messages are printed. Moving this summary print after the zeroing blocks would keep the log truthful.

Source/PBL/Shoc/ERF_ShocTypes.H
```diff
@@ -284,65 +165,40 @@ read_shoc_runtime_options (ShocRuntimeOptions& opts)
...
-    {
-        std::string error_message;
-        if (!parse_shoc_momentum_transport_string(momentum_transport,
-                                                  opts.momentum_transport,
-                                                  error_message)) {
-            amrex::Abort(error_message.c_str());
-        }
-    }
+    pp.query("lambda_low", opts.lambda_low);
+    pp.query("lambda_high", opts.lambda_high);
```
asalmgren, line 169: **LOW — unrelated wholesale `queryAdd` → `query` revert for 34 `erf.shoc.*` options.**

This is orthogonal to the PR's purpose and reverses the direction of the recent `queryAdd` work (#3922 / #3923 / #3925). The practical effect is that SHOC defaults are no longer recorded in the ParmParse table / job-info echo.

If it is deliberate (e.g. to stop the first `ShocDriver` from poisoning the table for later levels or for the unit tests), it deserves a comment explaining why; otherwise it should probably be dropped from this PR.

asalmgren (commented):
Automated code review of this PR (targeting issue #3912). Six findings below, one HIGH.

**What checked out:** re-ordering the `vert_implicit_fac` parse relative to the anelastic zeroing actually *fixes* a pre-existing bug (a user-supplied `erf.vert_implicit_fac` used to resurrect the implicit solve on anelastic levels); `pp.contains()` is prefix-aware and is correctly called before the `queryAdd` calls that would insert the keys; `SolverChoice::ReadParameters` is only invoked once (`Source/ERF_Constructors.cpp:122`) so the `user_set_*` probes are sound; the enum move compiles because `ERF_ShocTypes.H` already includes `ERF_DataStruct.H`; the new header is registered in `Source/DataStructs/Make.package` and CMake does not list headers; and the `:ref:`sec:Inputs`` target added in `PBLschemes.rst` exists at `Docs/sphinx_doc/Inputs.rst:4`.

PR #3926, opened 2026-08-26
MetGrid Nodal Reconstruction

AMLattanzi:
# Summary
This PR enables MetGrid to utilize the nodal reconstruction routine available for WRF input. The following are key changes

1. Input to do averaging versus nodal reconstruction changed to `erf.use_wrf_height_grid` changed to `erf.average_erf_height_grid`
2. Correction to averaging option in MetGrid where the bounds were incorrectly clamped.
3. Two new routines added to `ERF_NodalReconstruction.H` are `reconstruct_nodal_height_slice` and `fill_nodal_level_from_slice`. They are both now called from init from WRFInput and init from MetGrid

## Noted Benchmark Changes
1. MetGrid benchmarks will change since `erf.average_erf_height_grid = false`  is on by default.
2. MetGrid benchmarks will change with `erf.average_erf_height_grid = true` due to the correction of the averaging for boundary nodes.


asalmgren (commented):
Reviewed the terrain-construction changes; the clamping fix, the new `klo`/`klev` box guards, and the added `Gpu::streamSynchronize()` before `ParallelAllReduce::Sum` all look right. Five comments below, mostly about the migration path and the metgrid path now defaulting to reconstruction.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -867,15 +867,15 @@ struct SolverChoice {
         // Use lagged_delta_rt in the fast integrator?
         pp.queryAdd("use_lagged_delta_rt", use_lagged_delta_rt);
 
-        // Keep the default wrf grid or remake our own
-        pp.queryAdd("use_wrf_height_grid",use_wrf_height_grid);
+        // Average the input z-face heights onto the nodes, or reconstruct our own
+        pp.queryAdd("average_erf_height_grid",average_erf_height_grid);
```
asalmgren: **Renaming this input has no migration path, and the old name fails silently.**

`erf.use_wrf_height_grid` is gone and `erf.average_erf_height_grid` replaces it, but nothing rejects or warns about the old spelling. `ParmParse::QueryUnusedInputs()` is declared in `AMReX_ParmParse.H` and defined in `AMReX_ParmParse.cpp`, but it is never called anywhere in ERF or in the vendored AMReX, so an unrecognized `erf.*` entry is simply ignored.

Concrete scenario: a user has an existing deck containing `erf.use_wrf_height_grid = true` (the WRF z-face averaging path, with `rebalance_wrf_input` left at whatever they chose). After this PR that line is silently discarded, `average_erf_height_grid` falls back to its `false` default, and the run instead takes the nodal-reconstruction branch **and** hits the block just below, which force-sets `rebalance_wrf_input = true`. They get different terrain *and* a different hydrostatic base state, with no message telling them why their answers changed.

Suggested fix: keep a deprecation shim next to the new query, e.g.

```cpp
bool deprecated_flag = false;
if (pp.query("use_wrf_height_grid", deprecated_flag)) {
    amrex::Abort("erf.use_wrf_height_grid has been renamed to erf.average_erf_height_grid");
}
```

so stale decks fail loudly instead of quietly changing meaning.

Source/Initialization/ERF_InitFromMetgrid.cpp
```diff
@@ -666,39 +666,67 @@ ERF::init_from_metgrid (int lev)
...
+        // metgrid heights into managed memory for it
+        FArrayBox z_slice_hgt(hgt_box, 1, The_Managed_Arena());
+        const Array4<Real>& z_slice_hgt_arr = z_slice_hgt.array();
+        ParallelFor(hgt_box, [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept
+        {
+            z_slice_hgt_arr(i,j,k) = nc_hgt_arr(i,j,k);
+        });
+
+        // Solve for the nodal heights of the surface
+        FArrayBox z_slice_erf = reconstruct_nodal_height_slice(hgt_box, geom, z_slice_hgt,
```
asalmgren, line 724: **The metgrid reconstruction path has no post-construction grid check, unlike the wrfinput path.**

In `init_terrain_from_wrfinput` the corresponding `else` branch ends with a reduction over the first layer that computes `max_avg_err` and `min_dz` and aborts on either a non-finite value or `min_dz <= zero`, with a message pointing the user at `erf.average_erf_height_grid = true`. Nothing equivalent runs here. `get_dzmin_terrain()` is only called from `ERF_MakeNewArrays.cpp` and `ERF_Checkpoint.cpp`, i.e. before `init_from_metgrid` overwrites `z_phys_nd`, so it does not cover this either.

`reconstruct_nodal_height_slice` only guarantees that the nodal heights stay within `slack = max(0.5 * relief, 10)` of the range of `HGT_M`; it explicitly permits over/undershoot of the cell values where the terrain is under-resolved. Concrete scenario: a met_em file with steep, grid-scale terrain (the exact case that motivated the regularized solve) yields nodal surface heights that pass the range check but whose first nodal layer is degenerate or inverted. Because reconstruction is now the **default** for metgrid (`average_erf_height_grid` defaults to `false`), that mesh flows straight into `make_terrain_fitted_coords` and `make_J` and the user gets a bad Jacobian rather than an abort.

Suggested fix: after `fill_nodal_level_from_slice`, run the same first-layer `min_dz`/finiteness reduction the wrfinput path uses (it can be factored into `ERF_NodalReconstruction.H` alongside the two new helpers), and abort with the same "consider `erf.average_erf_height_grid = true`" advice.

Source/Utils/ERF_NodalReconstruction.H
```diff
@@ -832,4 +836,131 @@ private:
...
+ *
+ * @param[in] cc_slice_box Box covering the domain, cell-centered in (x,y) and a single k = 0 slab.
+ * @param[in] geom Geometry of the level, used for the grid spacing.
+ * @param[in] z_cc_slice Heights on \p cc_slice_box.
+ * @param[in] klev Vertical index of the slice; used only in messages.
+ * @param[in] src_name Name of the source data; used only in messages.
+ * @return The nodal heights on surroundingNodes(cc_slice_box), indexed at k = 0.
+ */
+inline amrex::FArrayBox
+reconstruct_nodal_height_slice (amrex::Box const& cc_slice_box,
```
asalmgren, line 860: **This helper can propagate C++ exceptions out of initialization with no handler.**

Two throw sites are reachable from here and neither is caught anywhere on the init path:

* the `NodalReconstruction` constructor throws `std::invalid_argument` when `face_box.length(2) != 1` or `face_box.smallEnd(2) != 0`;
* `conjugateGradient` throws `std::runtime_error("CG curvature failure.")`.

Under MPI an uncaught exception is `std::terminate`, not an `amrex::Abort`, so the user gets a bare `terminate called after throwing...` and no indication of which file or which level failed.

Concrete scenario on the newly-enabled metgrid path: `read_from_metgrid` pushes `HGT_M` into `BuildFABsFromNetCDFFile` but never inspects the returned `success` flag for it. If a met_em file lacks `HGT_M`, `NC_hgt_fab` is left default-constructed with an empty `Box` and null data. `init_terrain_from_metgrid` then computes `makeSlab(NC_hgt_fab.box(), 2, 0)`, whose `length(2)` is 0, and the constructor throws → terminate. (Pre-PR this silently produced zero terrain, so failing is an improvement — it just should not fail this way.)

Suggested fix: guard the entry point, e.g. `AMREX_ALWAYS_ASSERT_WITH_MESSAGE(cc_slice_box.ok() && cc_slice_box.length(2) == 1, ...)` in `reconstruct_nodal_height_slice`, check the `HGT_M` success flag in `read_from_metgrid`, and convert the `throw`s in the solver to `amrex::Abort` so failures surface as normal AMReX aborts.

Source/Initialization/ERF_InitFromMetgrid.cpp
```diff
@@ -666,39 +666,67 @@ ERF::init_from_metgrid (int lev)
...
+                int jj = std::max(std::min(j  ,jhi),jlo);
+                int jm = std::max(std::min(j-1,jhi),jlo);
+                z_arr(i,j,k) = fourth * ( nc_hgt_arr(ii,jj,klo) + nc_hgt_arr(im,jj,klo) +
+                                          nc_hgt_arr(ii,jm,klo) + nc_hgt_arr(im,jm,klo) );
+            });
+        } // mfi
+    } else {
+        // The reconstruction runs on the host, so stage the (already global)
+        // metgrid heights into managed memory for it
+        FArrayBox z_slice_hgt(hgt_box, 1, The_Managed_Arena());
```
asalmgren, line 716: **Cost note: this now puts a serial, whole-domain host solve on the default metgrid init path, redundantly on every rank.**

`reconstruct_nodal_height_slice` stages the full global slice and hands it to `NodalReconstruction`, whose loops (`solve`, `conjugateGradient`, `dot`, `applyJacobi`, `applyA`, `applyM`, ...) are plain host `for` loops over the entire domain, executed identically by every MPI rank. Per rank that is several full-domain managed `FArrayBox` allocations (`reference`, `misfit`, `rhs`, `E`, `S`, `m_cc_scratch`, `m_nd_scratch`, plus `diag`, `r`, `z`, `p`, `Mp` inside CG), and `applyDTD` allocates two more edge-box fabs on **every** matrix-vector product, i.e. once per CG iteration (up to `max_iterations = 1000` per attempt, times the mu-refinement attempts).

The wrfinput path already paid this, but it was opt-out there and it ran on the two surface levels of a domain the user had already accepted. Here it becomes the default for metgrid, where met_em domains can be large. Concrete scenario: a 2000x2000 met_em domain on 512 ranks does 512 identical copies of a 4M-node serial CG solve at startup, with per-rank managed allocations proportional to the whole domain — a startup stall and a memory spike that scales with rank count rather than with the decomposition.

Worth at least measuring on a realistic met_em case before making it the default; solving on one rank and broadcasting the resulting nodal slice would remove the redundancy cheaply, and hoisting the `difx`/`dify` fabs in `applyDTD` out of the iteration would remove the per-iteration allocation.

Source/Initialization/ERF_InitFromMetgrid.cpp
```diff
@@ -666,39 +666,67 @@ ERF::init_from_metgrid (int lev)
...
-       z_arr(i,j,k) =  fourth * ( nc_hgt_arr (ii,jj  ,k) + nc_hgt_arr(ii-1,jj  ,k) +
-                                nc_hgt_arr (ii,jj-1,k) + nc_hgt_arr(ii-1,jj-1,k) );
-   });
+    Print() << "Constructing nodal heights (z_phys_nd)" << std::endl;
+
+    // HGT_M is a single level of surface heights on the (cell-centered) mass grid
+    const Box hgt_box = makeSlab(NC_hgt_fab.box(), 2, 0);
+
+    // Nodal index of the surface
+    const int klo = hgt_box.smallEnd(2);
```
asalmgren, line 687: **`klo` is `0` by construction, but it is used as if it were derived from the data.**

`hgt_box` on the line above is `makeSlab(NC_hgt_fab.box(), 2, 0)`, which forces the k range to `[0,0]`, so `hgt_box.smallEnd(2)` is unconditionally `0` regardless of what `NC_hgt_fab` actually spans. That value is then used for two different things: as the k index into `nc_hgt_arr` in the averaging branch below, and as the *nodal* surface index of `z_phys_nd` passed to `fill_nodal_level_from_slice`.

Both uses happen to be correct today, but only because two independent facts line up: AMReX level domains start at k = 0, and `BuildFABsFromNetCDFFile` shifts the fab by `boxes_at_level[lev][0].smallEnd(2)`, which is also 0. The comment "Nodal index of the surface" reads as though the value came from the data, which makes the coupling invisible. If the z origin ever moved, `nc_hgt_arr(ii,jj,klo)` would read out of bounds silently rather than fail.

Suggested fix: derive the nodal surface index from `z_phys_nd` / the level domain, take the source k index from `NC_hgt_fab.box().smallEnd(2)` before the `makeSlab`, and assert the two agree — or at minimum reword the comment to say the slice is pinned to k = 0.

PR #3932, opened 2026-08-27
Nudge sounding option

wang1202:
Add options to nudge momentum and theta separately, and to customize the nudging range of height.

### DIFF
--- a/Docs/sphinx_doc/Inputs.rst
+++ b/Docs/sphinx_doc/Inputs.rst
@@ -1810,8 +1810,8 @@ List of Parameters
 +-------------------------------------+------------------------+-------------------+---------------------+
 | **erf.nudging_u**                   | Nudge u,v towards the  | Boolean           | true                |
 |                                     | sounding when          |                   |                     |
-|                                     | ``nudging_from_input_``|                   |                     |
-|                                     | ``sounding`` is true.  |                   |                     |
+|                                     | ``nudging_from_input_\ |                   |                     |
+|                                     | sounding`` is true.    |                   |                     |
 |                                     | Does not affect        |                   |                     |
 |                                     | LSF-based wind nudging |                   |                     |
 |                                     | (see                   |                   |                     |
@@ -1821,14 +1821,14 @@ List of Parameters
 | **erf.nudging_t**                   | Nudge potential        | Boolean           | true                |
 |                                     | temperature towards    |                   |                     |
 |                                     | the sounding when      |                   |                     |
-|                                     | ``nudging_from_input_``|                   |                     |
-|                                     | ``sounding`` is true   |                   |                     |
+|                                     | ``nudging_from_input_\ |                   |                     |
+|                                     | sounding`` is true     |                   |                     |
 +-------------------------------------+------------------------+-------------------+---------------------+
 | **erf.nudging_q**                   | Nudge water vapor      | Boolean           | true                |
 |                                     | towards the sounding   |                   |                     |
 |                                     | when                   |                   |                     |
-|                                     | ``nudging_from_input_``|                   |                     |
-|                                     | ``sounding`` is true   |                   |                     |
+|                                     | ``nudging_from_input_\ |                   |                     |
+|                                     | sounding`` is true     |                   |                     |
 +-------------------------------------+------------------------+-------------------+---------------------+
 | **erf.large_scale_forcing**         | Apply time-varying     | Boolean           | false               |
 |                                     | large-scale tendencies |                   |                     |
@@ -3297,8 +3297,8 @@ Custom, Nudging and Numerical Diffusion Forcing
 |                                                | ``large_scale_forcing``)                               |                                |                        |
 +------------------------------------------------+--------------------------------------------------------+--------------------------------+------------------------+
 | **erf.nudging_t**                              | nudge potential temperature towards the sounding       | Boolean                        | true                   |
-|                                                | (takes effect only when ``nudging_from_input_``        |                                |                        |
-|                                                | ``sounding`` is true)                                  |                                |                        |
+|                                                | (takes effect only when ``nudging_from_input_sounding``|                                |                        |
+|                                                | is true)                                               |                                |                        |
 +------------------------------------------------+--------------------------------------------------------+--------------------------------+------------------------+
 | **erf.nudging_q**                              | nudge water vapor towards the sounding (takes          | Boolean                        | true                   |
 |                                                | effect only when ``nudging_from_input_sounding``       |                                |                        |
--- a/Docs/sphinx_doc/Inputs.rst
+++ b/Docs/sphinx_doc/Inputs.rst
@@ -1785,7 +1785,7 @@ List of Parameters
 |                                     | initial sounding       |                   |                     |
 |                                     | profile                |                   |                     |
 +-------------------------------------+------------------------+-------------------+---------------------+
-| **erf.nudging_u_z1**                | Bottom of the height   | Real [m]          | 0.0                 |
+| **erf.nudging_u_z1**                | Bottom of the height   | Real [m]          | -1.0e36             |
 |                                     | range over which       |                   |                     |
 |                                     | u,v are nudged         |                   |                     |
 +-------------------------------------+------------------------+-------------------+---------------------+
@@ -1810,8 +1810,8 @@ List of Parameters
 +-------------------------------------+------------------------+-------------------+---------------------+
 | **erf.nudging_u**                   | Nudge u,v towards the  | Boolean           | true                |
 |                                     | sounding when          |                   |                     |
-|                                     | ``nudging_from_input_\ |                   |                     |
-|                                     | sounding`` is true.    |                   |                     |
+|                                     | nudging_from_input_\   |                   |                     |
+|                                     | sounding is true.      |                   |                     |
 |                                     | Does not affect        |                   |                     |
 |                                     | LSF-based wind nudging |                   |                     |
 |                                     | (see                   |                   |                     |
@@ -1821,14 +1821,14 @@ List of Parameters
 | **erf.nudging_t**                   | Nudge potential        | Boolean           | true                |
 |                                     | temperature towards    |                   |                     |
 |                                     | the sounding when      |                   |                     |
-|                                     | ``nudging_from_input_\ |                   |                     |
-|                                     | sounding`` is true     |                   |                     |
+|                                     | nudging_from_input_\   |                   |                     |
+|                                     | sounding is true       |                   |                     |
 +-------------------------------------+------------------------+-------------------+---------------------+
 | **erf.nudging_q**                   | Nudge water vapor      | Boolean           | true                |
 |                                     | towards the sounding   |                   |                     |
 |                                     | when                   |                   |                     |
-|                                     | ``nudging_from_input_\ |                   |                     |
-|                                     | sounding`` is true     |                   |                     |
+|                                     | nudging_from_input_\   |                   |                     |
+|                                     | sounding is true       |                   |                     |
 +-------------------------------------+------------------------+-------------------+---------------------+
 | **erf.large_scale_forcing**         | Apply time-varying     | Boolean           | false               |
 |                                     | large-scale tendencies |                   |                     |
@@ -1990,10 +1990,12 @@ zonal and meridional velocity, and large scale vertical subsidence velocity.
   and ``erf.nudging_q_z1``/``erf.nudging_q_z2`` options, respectively. ``erf.nudging_u``, ``erf.nudging_t``, and
   ``erf.nudging_q`` turn momentum, temperature, and moisture nudging on/off independently (all default to true).
 
-- **NOTE:** When both large scale forcing and ``erf.nudging_from_input_sounding`` are used, the u and v velocities are nudged
+- **NOTE:** All of the nudging described here, including LSF-based wind nudging, requires
+  ``erf.nudging_from_input_sounding = true``; none of it has any effect otherwise. When both
+  ``erf.nudging_from_input_sounding`` and ``erf.large_scale_forcing`` are true, the u and v velocities are nudged
   according to the observed large scale velocity ``uls`` and ``vls`` instead of the input sounding, using
-  ``erf.forcing_timescale``. Temperature and moisture are nudged from input sounding as normal, using the input
-  sounding ``tau_nudging`` timescale. This LSF-based wind nudging is controlled solely by
+  ``erf.forcing_timescale``. Temperature and moisture are still nudged from the input sounding as normal, using the
+  input sounding ``tau_nudging`` timescale. In this case, wind nudging is controlled solely by
   ``erf.large_scale_forcing`` and is unaffected by ``erf.nudging_u``, which only turns off/on
   sounding-based wind nudging (and, likewise, ``erf.nudging_u_z1``/``erf.nudging_u_z2`` only restrict the
   sounding-based branch).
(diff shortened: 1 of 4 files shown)


asalmgren:
PR 3932 ("Nudge sounding option", wang1202) is a small, mostly-clean change:
  four new inputs, defaults that preserve existing behavior, well-formed RST tables, documentation included. Three real issues:

  1. nudging_z_lo/nudging_z_hi are redundant and a misconfiguration trap — Source/TimeIntegration/ERF_AdvanceDycore.cpp:419

  if (nudge_theta && z >= t_z1 && z <= t_z2 && z >= nudge_z_lo && z <= nudge_z_hi) {

  The new pair is ANDed with the pre-existing nudging_t_z1/nudging_t_z2 on the same single if, and both pairs are absolute-height windows
  on the same quantity. So the new parameters can only narrow the theta range, never widen it — they add no capability that
  nudging_t_z1/t_z2 doesn't already provide. A user who reads the doc ("Additional top of the same range", default 1.0e36) and sets
  nudging_z_lo = 500, nudging_z_hi = 15000 silently gets nothing nudged above 10 km, because nudging_t_z2 still defaults to 10000.0. No
  warning. Best fix is to drop them; if they must stay, print the effective window [max(t_z1,z_lo), min(t_z2,z_hi)] at startup.

  Secondary: the names nudging_z_lo/z_hi read as global (all nudged fields), but they apply to theta only — not qv, not momentum. Worth
  renaming to nudging_t_* if kept.

  2. The u/v diagnostic block isn't gated by nudging_mom — Source/TimeIntegration/ERF_AdvanceDycore.cpp:437-467

  The block writing nudge_arr(i,j,k,2) and (,3) was left ungated, unlike the theta block above it and unlike the site that actually
  applies the momentum forcing (ERF_MakeMomSources.cpp:676). These components feed only the 1D profile diagnostics
  (Source/IO/ERF_Write1DProfiles_stag.cpp:841-842, the unudge/vnudge columns). With erf.nudging_mom = false, the profile files report
  non-zero momentum nudging tendencies at every level while no momentum nudging is applied anywhere — a tendency budget that doesn't
  close. Note theta is consistent here (nudge_data is zeroed at line 356, so nudge_theta = false correctly yields zero diagnostics); it's
  only momentum that's inconsistent. Adding the guard also skips the plane-average work.

  3. nudging_mom = false also silently disables large-scale-forcing wind nudging — Source/SourceTerms/ERF_MakeMomSources.cpp:676

  The guard was added to the outer if, which encloses the else branch at lines 703-715. That branch does not nudge toward the input
  sounding — it nudges toward the LSF winds uls/vls with the separate tau_lsf timescale, exactly the behavior the existing NOTE at
  Inputs.rst:1982 calls out. So a run with large_scale_forcing = true plus nudging_from_input_sounding = true that sets nudging_mom =
  false to switch off sounding-based wind nudging also loses its LSF wind nudging. The new doc text ("Nudge u,v towards the sounding")
  doesn't describe that. Either restrict the guard to the !large_scale_forcing branch, or document the coupling.

asalmgren:
# Code Review: PR #3932 — `nudge_sounding_option`

**Branch:** `nudge_sounding_option` → `development`
**Files:** `Docs/sphinx_doc/Inputs.rst`, `Source/DataStructs/ERF_DataStruct.H`,
`Source/SourceTerms/ERF_MakeMomSources.cpp`, `Source/TimeIntegration/ERF_AdvanceDycore.cpp`

## What the PR does

Adds three boolean switches — `erf.nudging_u`, `erf.nudging_t`, `erf.nudging_q` (all default
`true`) — that independently turn momentum, potential-temperature, and water-vapor nudging on and
off when `erf.nudging_from_input_sounding` is enabled. Also adds a height window for momentum
nudging, `erf.nudging_u_z1` / `erf.nudging_u_z2` (defaults `0.0` and `1.0e36`), matching the
existing `nudging_t_z*` and `nudging_q_z*` options. The bulk of the `ERF_AdvanceDycore.cpp` diff is
re-indentation from wrapping two existing blocks in the new gates.

## What checks out

- The restructuring in `ERF_AdvanceDycore.cpp` is brace-balanced and behavior-preserving at
  defaults — `git diff -w` shows the changes are almost entirely re-indentation.
- `nudge_data[level]->setVal(0.0)` at `ERF_AdvanceDycore.cpp:356` precedes the newly-skippable
  writes to `nudge_arr` components 2 and 3, so disabling momentum nudging leaves zeros rather than
  stale values.
- All new gates are host-side and rank-invariant, so there is no MPI-collective divergence between
  ranks.

## Findings

### 1. HIGH — momentum nudging height window is wrong on stretched vertical grids

`Source/SourceTerms/ERF_MakeMomSources.cpp:729` (and the matching `tby` kernel at `:738`)

```cpp
const Real zlo  = geom.ProbLo(2);
const Real dz   = geom.CellSize(2);
...
Real z = zlo + (k+0.5)*dz;
if (l_lsf || (z >= u_z1 && z <= u_z2)) { ... }
```

This flat-grid height formula is only valid for `MeshType::ConstantDz`. The plane-average block
that guards this function asserts only `mesh_type != MeshType::VariableDz`
(`ERF_MakeMomSources.cpp:173`), so `MeshType::StretchedDz` reaches the kernel. For stretched grids
`ERF_InitZLevels.cpp` builds `zlevels_stag` geometrically from `erf.initial_dz` and
`erf.grid_stretching_ratio` (or from `erf.terrain_z_levels`), with no relation to
`geom.CellSize(2)`, which remains the nominal `(prob_hi - prob_lo)/n_cell`.

**Failure scenario:** `erf.grid_stretching_ratio = 1.05`, `erf.initial_dz = 10`,
`erf.nudging_from_input_sounding = true`, `erf.nudging_u_z2 = 1000`. Momentum nudging is cut off at
whatever `k` satisfies the nominal formula rather than at 1000 m, and it disagrees with the
companion diagnostic in `ERF_AdvanceDycore.cpp:476`, which correctly writes
`(z_cc_arr) ? z_cc_arr(i,j,k) : zlo + (k+0.5)*dz` (`l_use_terrain_fitted_coords` is true whenever
`mesh_type != ConstantDz`, which includes `StretchedDz`).

**Fix:** use `z_cc_arr`, already declared unconditionally at `ERF_MakeMomSources.cpp:335`, in both
kernels — no null guard needed there.


### 2. MEDIUM — plane-average gate not updated for the new `nudging_u` flag

`Source/SourceTerms/ERF_MakeMomSources.cpp:169`

The block is still entered on `solverChoice.nudging_from_input_sounding` alone. The parallel gate in
`ERF_AdvanceDycore.cpp:204` *was* updated to `nudging_u || use_lsf`; this one was not.

**Failure scenario:** a terrain run (`erf.terrain_type = StaticFittedMesh` → `VariableDz`) with
`erf.nudging_from_input_sounding = true` and `erf.nudging_u = false` — wanting theta/qv-only
nudging, which lives entirely in `ERF_AdvanceDycore.cpp` and has no such assert — still aborts at
`AMREX_ALWAYS_ASSERT(solverChoice.mesh_type != MeshType::VariableDz)` on line 173, even though
`make_mom_sources` no longer consumes the u/v averages. On non-terrain runs it is three wasted MPI
plane-average reductions per slow step.

**Fix:** add `&& (solverChoice.large_scale_forcing || solverChoice.nudging_u)` to the
`nudging_from_input_sounding` term of the condition.

### 3. LOW — no validation that `nudging_*_z1 < nudging_*_z2`

`Source/DataStructs/ERF_DataStruct.H:1974`

`queryAdd` accepts any pair of values.

**Failure scenario:** a user transposes them (`erf.nudging_u_z1 = 1000`,
`erf.nudging_u_z2 = 100`); every `z >= z1 && z <= z2` test is false, so momentum nudging is
silently off everywhere and the run looks like a physics problem rather than a typo. Because
`nudging_u_z2` now defaults to `1.0e36` (unbounded) rather than a physical value, a typo in `z2`
alone has the same effect.

**Fix:** an `AMREX_ALWAYS_ASSERT_WITH_MESSAGE(z1 < z2)` for each of the u/t/q pairs.

### 4. LOW — split inline literals render the parameter name incorrectly

`Docs/sphinx_doc/Inputs.rst:1813` (also `:1824`, `:1830`, `:3300`)

The table cells wrap the name as ``` ``nudging_from_input_`` ``` on one line and
``` ``sounding`` ``` on the next. RST joins the two code spans with a space, so the rendered docs
read `nudging_from_input_` `sounding` — a parameter name the user cannot copy or search for. The
same file writes the name unbroken elsewhere, so the tables are internally inconsistent.

**Fix:** use a single unbroken ``` ``erf.nudging_from_input_sounding`` ``` (widening the column if
needed), or drop the literal markup on the wrapped occurrences.

### 5. LOW — `nudging_u` and its height window are silently ignored under LSF

`Source/SourceTerms/ERF_MakeMomSources.cpp:680` (also `ERF_AdvanceDycore.cpp:448` / `:477`)

The gate is `(solverChoice.large_scale_forcing || solverChoice.nudging_u)` and the in-kernel test is
`l_lsf || (z >= u_z1 && z <= u_z2)`, so with LSF enabled neither the flag nor the height window has
any effect.

**Failure scenario:** a user on an LSF case sets `erf.nudging_u = false` (or
`erf.nudging_u_z2 = 500`) expecting wind nudging off or limited; the run still applies full-domain
nudging toward `uls`/`vls` with no warning.

This is deliberate and documented in the new `Inputs.rst` prose, so it is a hardening suggestion
rather than a bug — but a `Print` (or `Abort`) when `nudging_u` / `nudging_u_z*` is set together
with `erf.large_scale_forcing` would turn a silent surprise into a clear message.

## Suggested order of work

1. Fix the `z_cc_arr` height computation (finding 1) — this one changes results.
2. Extend the plane-average gate (finding 2) — unblocks theta-only nudging on terrain.
3. Fix the RST literals (finding 4) — trivial, user-visible.
4. Optionally add the `z1 < z2` asserts and the LSF-conflict message (findings 3 and 5).

asalmgren:
# Code Review: PR #3932 — `nudge_sounding_option` (round 2)

**Branch:** `nudge_sounding_option` → `development`
**Files reviewed:** `Docs/sphinx_doc/Inputs.rst`, `Source/DataStructs/ERF_DataStruct.H`,
`Source/SourceTerms/ERF_MakeMomSources.cpp`, `Source/TimeIntegration/ERF_AdvanceDycore.cpp`

This round supersedes `pr3932_review.md`, which was written against an earlier state of the
branch — several of its findings (the plane-average gate, the missing `z1 < z2` asserts, the
`z_cc_arr` height computation) have since been addressed.

## Findings

### 1. MEDIUM — `nudging_u_z1 = zero` default is not behavior-preserving

`Source/DataStructs/ERF_DataStruct.H:1993`

Before this PR, sounding-based momentum nudging was applied at every `k` unconditionally. Now
`ERF_MakeMomSources.cpp:730` requires `z >= 0.0`, so any domain whose cell-center heights can be
negative silently loses nudging in those cells.

**Failure scenario:** `geometry.prob_lo = ... -1` — used in `Exec/RegTests/InitFromNCFile/inputs`,
`IsentropicVortex/inputs_advecting`, and `Couette_Poiseuille/inputs_poiseuille_x` — or
terrain-fitted coordinates with terrain below the datum. Cells below `z = 0` lose momentum nudging
with no diagnostic.

**Fix:** default `nudging_u_z1` to `-1.0e36`, mirroring the `1.0e36` upper sentinel, so the
default window is a true no-op.

### 2. LOW/MEDIUM — `erf.nudging_u = false` is silently overridden by LSF

`Source/SourceTerms/ERF_MakeMomSources.cpp:683`

The gate is `(large_scale_forcing || nudging_u)`; inside, `if (!l_lsf)` takes the LSF branch, and
the kernel's `if (l_lsf || ...)` short-circuits true. So under LSF, wind nudging *and* the height
window are applied in full regardless of the flag.

**Failure scenario:** a user running LSF sets `erf.nudging_u = false` to disable wind nudging.
Nothing changes; the only signal is a startup `amrex::Print()` warning.

**Fix:** make `nudging_u = false` skip the block regardless of LSF, or rename the option so its
scope is unambiguous.

### 3. LOW — documentation contradicts the code gate

`Docs/sphinx_doc/Inputs.rst:1996`

The new prose states that "This LSF-based wind nudging is controlled solely by
``erf.large_scale_forcing``", but both code gates (`ERF_MakeMomSources.cpp:682` and
`ERF_AdvanceDycore.cpp:359`) also require `nudging_from_input_sounding == true`.

**Failure scenario:** `erf.large_scale_forcing = true` with
`erf.nudging_from_input_sounding = false` produces no wind nudging at all, contradicting the doc.

### 4. LOW — broken line continuation inside inline literals

`Docs/sphinx_doc/Inputs.rst:1813` (also `:1824`, `:1830`)

The cells wrap the name as ``nudging_from_input_\`` followed by ``sounding`` on the next line,
relying on a backslash line continuation. Backslash escapes are not processed inside
reStructuredText inline literals. Verified with docutils 0.23: all three new table cells render as
`nudging_from_input_\ sounding`, with a visible stray backslash.

**Fix:** widen the column, or drop the literal markup across the line break.


### 5. LOW — new asserts abort inputs that previously ran

`Source/DataStructs/ERF_DataStruct.H:1102`

The new `AMREX_ALWAYS_ASSERT_WITH_MESSAGE(nudging_t_z1 < nudging_t_z2)` and
`(nudging_q_z1 < nudging_q_z2)` checks hard-abort at startup on input files that used to run.

**Failure scenario:** the natural pre-PR way to disable moisture nudging was
`erf.nudging_q_z2 = 0.0` with `q_z1` defaulting to `0`. That combination now aborts instead of
running.

Now that `erf.nudging_q` / `erf.nudging_t` exist this is arguably the desired direction, but it is
an undocumented breaking change — worth a release note, or relaxing the comparison to `<=`.

### 6. LOW — sentinel value hardcoded in two places

`Source/DataStructs/ERF_DataStruct.H:1118`

`if (nudging_u_z1 != zero || nudging_u_z2 != amrex::Real(1.0e36))` re-hardcodes the sentinel that is
also written at line 1994. If either default changes, the "did the user set this?" heuristic
silently inverts — the warning either fires on every LSF run or never fires.

**Fix:** use a named `static constexpr` sentinel, or detect an explicit setting via
`pp.contains("nudging_u_z1")` / the return value of `queryAdd`.

## Verified as correct — no need to re-check

- The new `if (solverChoice.nudging_u || use_lsf)` gate at `ERF_AdvanceDycore.cpp:206` matches the
  gate on the only consumer of `dptr_u_plane` / `dptr_v_plane` (the `nudge_arr` comps 2/3 kernel at
  line 448). `nudge_data` is diagnostic-only (`Write1DProfiles_stag.cpp:838-842`) and is
  `setVal(0.0)`'d first, so skipped writes leave zeros, not garbage. Immersed forcing needs only
  `r_plane` / `t_plane`, which remain unconditional.
- The `make_mom_sources` plane-average gate (line 169) is exactly consistent with the section-5
  nudging gate (line 682), so `dptr_r/u/v_plane` are never null-dereferenced.
- `z_cc_arr` at `ERF_MakeMomSources.cpp:729/738` is in bounds: `z_phys_cc[lev]` is always allocated
  with 2 ghost cells (`ERF_MakeNewArrays.cpp:69`) and ghosts are filled by `make_zcc`
  (`growntilebox` + `FillBoundary`), so the `i = hi+1` face index on `nodaltilebox(0)` is safe.
- All gates are uniform across ranks, so the skipped `PlaneAverage` reductions cannot deadlock.
- The `zlo + (k+0.5)*dz` fallback in `AdvanceDycore` is reached only for `MeshType::ConstantDz`
  (`l_use_terrain_fitted_coords = mesh_type != ConstantDz`), where it is exact.
- Default behavior is otherwise preserved: all three new booleans default `true` and `nudging_u_z2`
  defaults to `1.0e36`.

## Suggested order of work

1. Change the `nudging_u_z1` default to `-1.0e36` (finding 1) — this one changes results.
2. Decide the LSF / `nudging_u` precedence and fix code or docs to match (findings 2 and 3).
3. Fix the RST literals (finding 4) — trivial, user-visible.
4. Optionally relax or document the new asserts, and factor out the sentinel (findings 5 and 6).

PR #3944, opened 2026-08-29
allow refinement in vertical when given wrfinput refined laterally

asalmgren:


asalmgren:
This passes all the biollante regression tests.

PR #3945, opened 2026-08-29
Explore using symmetry not foextrap for nodal values of zphys

asalmgren:


AMLattanzi:
CPU regtest results:
```
ParticlesOverWoA FAILED
WPS_Test_Terrain FAILED
WPS_Test_Terrain-OMP FAILED
```

PR #3947, opened 2026-08-29
Corrections for Dynamic AMR

asalmgren:
# Summary
Modifications in support of dynamic AMR with realistic simulations.

## 1. The core defect

`init_stuff()` rebuilds a level's 2D/surface arrays every time that level is **created or remade**, and initializes them to placeholders:

| Array | Placeholder left by `init_stuff` | Only real source |
|---|---|---|
| `mapfac[lev][*]` (MAPFAC_M/U/V) | `1.0` | `init_from_wrfinput` / `init_from_metgrid` |
| `mf_PSFC[lev]` | uninitialized | `init_from_wrfinput` |
| `lmask_lev`, `land_type_lev`, `soil_type_lev` | uniform (`erf.is_land`, type 0) | `init_from_wrfinput` |

Both real sources require an init file **at that level** (`erf.nc_init_file_<lev>`). Consequently:

* a level created by ERF's own tagging/refinement ran with a map factor of 1, no surface pressure, and
  a uniform land mask, while its parent used the true WRF fields;
* **any** regrid — including the level-0 regrid `restart()` forces whenever the checkpoint has fewer
  grids than ranks — silently discarded the WRF fields at that level.

This is not cosmetic. `mfsq = mf_mx*mf_my` divides the horizontal fluxes in the continuity and scalar
equations (`AdvectionSrcForState`) and in the acoustic substep, so the fine level was solving a
*different mass equation* from the one that produced the momenta handed to it across the coarse/fine
boundary. The land mask selects the land vs. sea flux branch in `SurfaceLayer::compute_fluxes`, gates
which cells `update_sst_tsk` applies SST to, and (with the land/soil types) drives the LSM.

## 2. New interpolation helpers (`Source/ERF.cpp`, declared in `Source/ERF.H`)

* **`has_own_init_file(lev)`** — true iff `nc_init_file[lev]` is non-empty; the gate that distinguishes
  "this level read its own file" from "everything must come from the parent."
* **`interp_mapfac_from_coarse(lev)`** — interpolates each `mapfac` entry from `lev-1`, picking the
  interpolater by index type (`cell_cons_interp` for `m_x`, `face_cons_linear_interp` + the matching
  `xvel_bc`/`yvel_bc` for the face-based `u_x`/`v_x`), with `rr2d = (rr_x, rr_y, 1)` because these live
  on `ba2d`. Exterior ghost cells **are** filled (advection reads map factors on grown boxes), followed
  by `FillBoundary` so a fine neighbour's own data wins. Justified physically: the map factor is smooth,
  varying by O(1e-5) across a coarse cell.
* **`interp_psfc_from_coarse(lev)`** — same shape, `cell_cons_interp`, but **does not** fill exterior
  ghosts (the coarse PSFC is only copied over the domain intersection, so there is nothing to
  interpolate; `update_sst_tsk` clamps its index). No-ops unless `init_type == WRFInput`.
* **`interp_land_masks_from_coarse(lev)`** — piecewise-constant **injection** (`pc_interp`) of the land
  mask and land/soil types, routed through a temporary real-valued `MultiFab` (values are small integers;
  `pc_interp` does no arithmetic, so the round trip is exact) and `std::round`ed back to `int`. Exterior
  ghosts are filled because `compute_fluxes` reads the mask on grown boxes without clipping. A fine
  land-mask value of `2` (building, stamped by `init_zphys`) is preserved rather than overwritten by the
  parent. No-ops unless `init_type == WRFInput`.

`Interp2DArrays` now calls all three up front, guarded by `!has_own_init_file(lev)`; the ad-hoc PSFC
interpolation block that used to live in `ERF::init_only` was deleted in favour of the helper.

## 3. Ghost cells outside the domain for other 2D fields

In `Interp2DArrays`, four `InterpFromCoarseLevel` calls switched from `IntVect(0,0,0)` to `ngv` for the
exterior-ghost argument:

* **`sinPhi_m`, `cosPhi_m`** — the variable-Coriolis momentum sources average them onto the faces of the
  nodal tilebox (`ERF_MakeMomSources.cpp`), so the x-momentum source at `i = nx` reads index `nx`. Before
  this change a refined level computed its hi-face Coriolis source from uninitialized memory.
* **`sst_lev`, `tsk_lev`** — `SurfaceLayer` populates `t_surf` on grown tileboxes, reading past a
  non-periodic face.

Correctness of these rests on a stated precondition: **all** coarse ghost cells (inside and outside the
domain) must already be filled. The comments argue this holds — `init_from_wrfinput`/`init_from_metgrid`
fill the grown box by clamping into the valid region at a level with a file, and the new interpolation
fills it at a level without one.

## 4. Preserving surface data across a regrid (`Source/ERF_MakeNewLevel.cpp`)

`RemakeLevel` now stashes `mapfac`, `mf_PSFC`, `lmask_lev[lev][0]`, `land_type_lev[lev][0]`,
`soil_type_lev[lev][0]` *before* `init_stuff`, then:

1. at `lev > 0`, interpolates from the parent, so cells the new grids **added** (never covered by the
   pre-regrid arrays) are defined;
2. `ParallelCopy`s the retained values on top wherever they exist, then `FillBoundary`.

This is what carries the arrays across a regrid on non-WRF pathways too, since the helpers no-op there.
It also fixes a case unrelated to WRF: `init_zphys` marks cells under a building with land mask 2, but a
regrid goes through `remake_zphys`, which does not redo that marking — the retain/restore keeps it.

## 5. Restart robustness (`ERF.cpp`, `ERF_InitFromWRFInput.cpp`)

`wrf_bsp` (six WRF reference-state parameters) is filled by `init_from_wrfinput` at level 0, which only
runs on a cold start, and is **not** in the checkpoint. Level 0 never noticed, because its base state is
read back from the checkpoint — but the first regrid that creates or remakes a refined level calls
`rebuild_base_state_from_wrfinput`, which aborted on the `is_set` assertion.

New **`restore_base_state_params_on_restart()`** re-reads them from `nc_init_file[0][0]` and calls
`set_layer_interfaces()`. It is invoked in `ERF::restart()` immediately after `ReadCheckpointFile()` and
before the forced level-0 regrid. Guarded by `ERF_USE_NETCDF` and `init_type == WRFInput`.

## 6. LSM refinement-ratio bug (`ERF_MakeNewLevel.cpp`, Noah-MP)

`lsm.Init(...)` was passed `refRatio(lev-1)` — the ratio across the **last** interface only — while
`NOAHMP::interp_from_lev0` interpolates from `Geom(0)`. Both call sites (`MakeNewLevelFromScratch`,
`MakeNewLevelFromCoarse`) now pass the **product across every interface beneath the level**. Wrong for
any run with three or more levels; correct by coincidence for two.

Separately, `NOAHMP::interp_from_lev0` now builds `rr2d = (rr_x, rr_y, 1)`: the LSM fields have no
vertical extent (soil layers are carried in components, not cells), and a z-ratio > 1 makes the
interpolater demand a layer of coarse z ghost cells these MultiFabs cannot carry.

## 7. MOST / surface-layer averaging (`ERF_MOSTAverage.{H,cpp}`)

The largest behavioural block. Two things are being fixed: (a) a fine patch's 2D averages had undefined
ghost cells, and (b) the old exterior-ghost logic keyed off `boxArray().minimalBox()`, which is meaningless
at a refined level.

* **Policy override in the constructor.** If `m_geom.size() > 1` (i.e. `amr.max_level > 0`) and the policy
  is not local, ERF prints a note and forces `erf.most.average_policy = 1`, because a plane average over a
  fine patch does not span the domain and the surface layer would see a different mean on either side of a
  level boundary. Under `TerrainType::EB` it only warns ("proceed with caution"). Keyed on levels
  *allowed*, not levels in use, since the policy cannot change once the time-filter state is built.
* **`get_ng_fill(lev)`** (new) — how many 2D ghost cells are computable: `min` of the allocated growth over
  averages and fields, minus `m_radius + 1` (an average reads `m_radius` cells each way; the tangential
  velocity magnitude and the interpolation stencil reach one further), floored at 0, zero in z.
* **`extrap_ghost_cells(lev, iavg, ng_fill)`** (new) — zeroth-order extrapolation of the nearest computed
  value into the outer ghost layers that `get_ng_fill` excludes, so no average is left undefined; run
  *before* `FillBoundary`, which then overwrites anything shared with a real neighbour.
* **`compute_region_averages`** — the three averaging loops changed from `mfi.tilebox()` to
  `mfi.growntilebox(ng_fill)`, each followed by `extrap_ghost_cells` then `FillBoundary`. An isolated fine
  patch has no neighbour to communicate with, but the underlying fields *do* carry valid ghost data filled
  from the coarse level, so the averages can be computed there directly.
* **Index/position stencils grown** — `set_k_indices_T`, `set_norm_indices_T`, `set_z_positions_T`,
  `set_norm_positions_T` now size their nodal boxes from `get_ng_fill` (positions additionally grown by
  `m_radius`), so the data the widened averages read is actually defined.
* **`set_rotated_fields`** — rotates into the ghost region as well (stopping one cell short, since the
  rotation reaches one cell ahead in `w` and one node ahead in `z_phys`).
* **Exterior-domain fill rewritten** — the old test `domain.contains(cc_bnd_bx) || not_per_x || not_per_y`
  with clamping to `fields[imf]->boxArray().minimalBox()` is replaced by: clamp to the **domain box
  converted to the average's index type**, and only in directions that are actually non-periodic
  (per-direction `d_not_per_x` / `d_not_per_y` inside the kernel). Interior ghosts of an isolated patch
  are now deliberately left alone, having been computed above.

## 8. Build, style, and test input

* **`CMake/SetERFCompileFlags.cmake`** — `set_cuda_architectures(AMReX_CUDA_ARCH)` is now guarded by
  `if(DEFINED AMReX_CUDA_ARCH)`, so a CUDA configure that does not set the arch no longer errors in that
  macro.
* **`Exec/ERF_Prob.cpp`, `Source/Prob/ERF_InitCustomPert{,Vels}_ABL.H`** — no functional change: the
  `ParmParse` objects in `init_custom_pert` / `init_custom_pert_vels` are scoped in braces and renamed
  (`pp` → `pp_prob` / `pp_erf`) to remove a shadowing warning between the caller and the `#include`d
  problem body, plus `amrex::ParmParse` → `ParmParse` for consistency.
* **`Exec/RegTests/WPS_Test/inputs_wps_ml`** (new, 120 lines) — the exercise case for all of the above: WRF
  input at level 0 only, `amr.max_level = 1`, `amr.ref_ratio_vect = 3 3 1`, `erf.regrid_int = 1`,
  qc-threshold tagging, `surface_layer` with `average_policy = 1` and `radius = 0`, variable + 3D Coriolis,
  `StaticFittedMesh` terrain, Kessler moisture, MRF PBL. Reads `wrfinput_chisholmview_d01` /
  `wrfbdy_chisholmview_d01_width1`.

---

## Implications

**Physics / results.** Any multi-level WRF-input run in which a level was created by tagging, or in which a
regrid occurred, was wrong before this PR — map factor 1 on the fine level against real map factors on the
coarse level is an inconsistent mass equation across the coarse/fine interface, and the land mask reverting
to uniform silently switched surface-flux branches. Expect changed answers for such cases; single-level
runs and runs where every level reads its own `nc_init_file_<lev>` are untouched by §2–§4.

**Restart.** Restarting a WRF-input case with refinement no longer aborts on the first regrid. Note the new
implicit requirement: the level-0 `wrfinput` file must still be present and readable at restart time. The
PR takes this deliberately ("the file is still the authoritative source") over adding the parameters to the
checkpoint — cheaper and back-compatible with existing checkpoints, but it makes restarts depend on an
external file that previously was not needed.

**User-visible behaviour change.** `erf.most.average_policy` is now overridden to 1 (local) in *any* run
with `amr.max_level > 0`, whether or not a fine level is ever created. Users with existing multi-level
inputs specifying plane averaging will get different answers and a printed note. Worth checking whether
any regression gold files depend on plane averaging at `max_level > 0`.

**Cost.** Region averages, rotated fields, and the index/position stencils are now computed over grown
boxes instead of valid boxes; the extra work scales with the ghost width, so it is a small constant-factor
increase in the surface-layer setup and per-step averaging, not an asymptotic change. `interp_land_masks_from_coarse`
allocates two temporary real `MultiFab`s per array per call, but only at level creation/regrid.

**Known gaps, deliberately left.**
* `init_from_metgrid` also reads `LANDMASK`, but as a time series (one `iMultiFab` per met_em time) rather
  than the single entry handled here; the metgrid pathway is explicitly not covered. Both
  `interp_psfc_from_coarse` and `interp_land_masks_from_coarse` return immediately unless
  `init_type == WRFInput`, so metgrid multi-level runs still carry the `init_stuff` placeholders.
* MOST with `TerrainType::EB` and `max_level > 0` only warns; the plane-average inconsistency is left in
  place there.
* The CMake guard means that with CUDA enabled but `AMReX_CUDA_ARCH` unset, `CUDA_ARCHITECTURES` is set from
  an empty `AMREX_CUDA_ARCHS` rather than from a detected default.

**Review focus.** The exterior-ghost-cell changes (§3) all depend on the claim that every coarse ghost cell
is pre-filled, inside the domain and out; that invariant is asserted in comments rather than in code, and it
is the thing most likely to bite later. Second, `get_ng_fill`'s `m_radius + 1` margin and the grown stencils
in §7 are coupled — a future change to the averaging stencil width has to move both.


AMLattanzi (approved):

AMLattanzi:
Tagging was tested on `qc` criteria (below). The new inputs file `inputs_wps_ml` now marches 2 steps successfully with a regrid.
image

PR #3950, opened 2026-09-01
Add two-stream radiation model with simplified surface energy balance (force-restore)

hgopalan:
## Summary

Adds a **two-stream atmospheric radiation model** as an alternative to RRTMGP, together with an optional **simplified surface energy balance (SEB)** module that evolves surface temperature and moisture by a force-restore method where Noah-MP is not driving them. The model is selected with `erf.radiation_model = "TwoStream"`, the same selector as RRTMGP and Simple, and runs through its own class rather than the RRTMGP interface.

Everything new is off by default, so a run that does not ask for two-stream radiation is unaffected.

## Shortwave

- Direct beam by Beer-Lambert attenuation, `F_dir(z) = S0 cos(zenith) exp(-tau_cum(z) / cos(zenith))`.
- Diffuse field from the Meador-Weaver (1980) two-stream solution with Zdunkowski PIFM coefficients, parameterized by single-scattering albedo and asymmetry factor, with independent clear-sky and cloud values.
- The layer solutions are combined by the adding method, so the surface albedo reflects the beam back into the column rather than being applied only as an absorbed fraction. Direct and diffuse albedo can differ (`surface_albedo_sw_diffuse`).
- Optional Earth-Sun distance factor (Spencer 1971) scaling `S0` by day of year.

## Longwave

- Gray-gas two-stream sweep on layer interfaces (Toon et al. 1989), with Stefan-Boltzmann surface emission at a configurable emissivity and reflection of the downwelling flux at the surface.
- Cooling to space is set by the flux divergence, with no isothermal override.

## Optical depth

Two models, selected by `erf.radiation.tau_model`:

- `per_layer` (default): a fixed optical depth per layer, so the column value scales with the number of vertical cells.
- `mass`: built from each layer's mass path `rho dz`, so the column optical depth is set by the atmosphere rather than by the grid. Shortwave mixes dry absorption, Rayleigh scattering, water vapour and cloud water with extinction-weighted single-scattering albedo and asymmetry factor. Longwave uses the corresponding mass absorption coefficients, with the cloud term expressed as the Stephens (1978) emissivity.

On top of either model: a prescribed cloud band, moisture- and cloud-aware dynamic optical depth diagnosed from `qv` and `qc`, prognostic cloud fraction diagnosed from relative humidity and cloud water, and a prescribed bulk aerosol optical depth that is constant or exponential with height.

## Coupling

- Heating rates are stored as **potential-temperature** tendencies, `dtheta/dt = (dT/dt) / pi`, matching the convention the RhoTheta source term and the RRTMGP path already use. They are written to `qheating_rates` and exposed as the `qsrc_sw` and `qsrc_lw` plot variables.
- Applied once per slow step, during slow-RHS construction only, so the tendency is consistent across all fast substeps.
- Time-varying solar zenith angle from simulation time, latitude, longitude and day of year, with a static fallback.
- Per-column surface albedo, emissivity and temperature from an LSM when present, falling back to scalars.
- Non-uniform vertical grids: layer thicknesses are the distances between the layer interfaces of the nodal heights `z_phys_nd`, with the top layer's own thickness.

## Simplified surface energy balance

- Allocates the surface flux fields, passing Noah-MP values through when it is active and using scalar fallbacks otherwise.
- Diagnostic mode reports the residual `R_net - H - LE - G`.
- Prognostic mode evolves surface temperature and moisture by force-restore with bounded explicit Euler updates, and switches itself off at any level where Noah-MP is already driving the surface state.

## Grid requirement

The column sweep integrates a whole atmospheric column in one pass, so each grid must span the domain in the vertical. ERF only decomposes in `z` when `amr.max_grid_size_z` is smaller than the number of cells in `z`; setting it at least that large is sufficient. A vertically decomposed grid aborts with a message naming the input rather than returning heating rates that look plausible and are wrong. Results do not depend on the horizontal decomposition or on the MFIter tile size.

## Input validation

Several misconfigurations used to be accepted in silence. None of them aborts, so existing inputs keep running, but each now prints a warning naming the input, the valid values, and what the run will actually do instead:

- an unrecognized `tau_model`, `tau_profile_type`, `aerosol_profile_type` or `diag_callsite_mode` falls back to its default, which is stated in the warning;
- `aerosol_profile_type = table` is not implemented and contributes no aerosol optical depth;
- `seb_prognostic_enable` reports that it also switches on `seb_enable` and `seb_diagnostic_enable`, which it depends on;

## Cost and scaling

`Exec/CanonicalTests/Radiation/TwoStream_vs_RRTMGP_Timing/` measures what each solver costs per radiation call and how that cost scales, with everything except the solver held fixed.

Radiation solver cost and scaling

Measured on one CPU core, marginal cost per call:

| Grid | Cells | Two-stream | RRTMGP |
|---|---|---|---|
| 42 x 4 x 42 | 7,056 | 1.08 ms | 433 ms |
| 84 x 4 x 42 | 14,112 | 2.07 ms | 881 ms |
| 84 x 4 x 84 | 28,224 | 4.27 ms | 1,812 ms |
| 168 x 4 x 84 | 56,448 | 8.55 ms | 3,581 ms |

The two solvers compute different things: RRTMGP resolves 224 shortwave and 256 longwave g-points with gas optics from lookup tables, while the two-stream solver does a single gray sweep per column. The difference is the cost of spectral detail, so this is a guide to choosing a solver for a given run rather than a ranking. Cost per cell is flat for both across a factor of eight in problem size, which means either one extrapolates predictably to a production grid.

Both radiation paths previously had no `BL_PROFILE` region, so radiation cost could not be separated from the dycore for either solver. One is now added at each entry point, and the measurement reads exclusive time from AMReX's TinyProfiler.

Keeping it fair: both inputs pull in one shared settings block through AMReX's `FILE` include, pinning the grid, timestep, sounding, moisture model and surface temperature; radiation is called every slow step on both sides; plotfiles and both diagnostic logs are off so file I/O stays outside the measured region. The reported figure is the marginal cost across two step counts, which cancels one-time work, notably the roughly 150 ms RRTMGP spends reading its lookup tables on the first call. `check_timing_fairness.py` enforces those invariants and needs no RRTMGP build.

Caveats: a single CPU core, RRTMGP targets GPUs, and cost per call is not cost per unit of accuracy.

## Testing

- 34 GoogleTest cases covering the shortwave and longwave kernels and the column sweep, including energy-budget closure, the Exner convention, surface reflection, and resolution independence of the mass model.
- Two ctest regressions: a column-heating check that verifies the vertical structure and orientation of the heating rates, and a plotfile check that `qsrc_sw` and `qsrc_lw` are written.
- 17 canonical cases under `Exec/CanonicalTests/Radiation/`, each with a Python checker, covering clear-sky and cloudy shortwave, scattering, longwave, non-uniform spacing, dynamic optical depth, prognostic cloud fraction, aerosol, diurnal solar geometry, surface heterogeneity, the mass optical-depth model, RhoTheta and YSUNew coupling, and all three SEB modes.

## Documentation

- `Docs/sphinx_doc/theory/Radiation.rst` covers the shortwave and longwave formulations, the optical-depth models, the potential-temperature convention, the SEB equations and the grid requirement.
- `Docs/sphinx_doc/Inputs.rst` documents every input.
- `Exec/CanonicalTests/Radiation/inputs_twostream_master` is the reference input file and lists all 79 `erf.radiation.*` parameters with defaults, units and descriptions.
- The development notes that accompanied this work (`RAD_DEVELOPMENT.md`, `RAD_MPI_SKILLS.md`) and the benchmark suite's generated outputs (`benchmark_summary.*`, the `radiation_*_diag.dat` files under its cases) are not part of the PR; they are listed in `.gitignore`, as the Canonical_RANS notes were handled in #3993. The theory and inputs pages above are the documentation of record, and nothing in the tree refers to the notes.

## Changes since the development merge of 11 September

An audit of the branch, and the follow-ups it produced, in the commits after a6dbe9a44.

**Defects fixed**

- The development merge had dropped the `HOL_abs` declaration in the MRF scheme that #3972 relies on, so every CI build failed; the file is development's again.
- The force-restore surface update multiplied its tendency by the absolute simulation time instead of the step size (2x to 72x the Euler step over a 36 s run). The driver now takes `time` and `dt` separately.
- The column sweep ran twice per step: the post-dycore call re-swept the same old state. It now runs once, at the pre-dycore call; the post-dycore call reports that sweep's cached flux diagnostics and steps the surface state, so the diagnostics CSV is unchanged.
- The prognostic surface temperature and moisture were not in the checkpoint, so a restart reset them to the scalar defaults. They are written as `Level_<lev>/TwoStream_TSfc` and `TwoStream_QSfc` and read back when present.
- `Tests/Unit/Utils/ERF_GTestTerrainFineColumns.cpp` (from #3976) odr-used an anonymous-namespace constexpr inside a device lambda, which nvcc rejects; the lambda now captures a local copy. This is the CUDA failure that showed on the merged head and will show on development until this merges.

**SEB coupling and inputs**

- New opt-in `erf.radiation.seb_use_radiation_fluxes`: where the LSM does not supply them, the net surface shortwave and longwave fluxes of the energy balance are the sweep's own per-column surface fluxes rather than the `seb_*_default` constants. Noah-MP fields are mapped as `sav + sag` for absorbed shortwave and `-fira` for absorbed longwave; the deep-soil reservoir values are the scalar defaults, since no LSM exposes them by name (the previous mapping had read soil-moisture fields).
- `erf.radiation_model = TwoStream` with `amr.max_level > 0` now stops at start-up with the reason (the sweep needs whole columns of its level) instead of asserting on `max_grid_size_z` at the first fine-level call. The theory page gains a Limitations section (single level, solar time base of 00:00 UTC on `day_of_year`, the diagnostics file written by default).
- `erf.ysu_rad_tend_smooth_strength` was read and never used and is removed; a non-positive `ysu_rad_tend_limiter_magnitude` warns instead of silently becoming 1. Both YSUNew keys and the new SEB key are in `Inputs.rst` and the master deck.

**Structure**

- The model is now a class, `TwoStreamRadiation` (`Source/Radiation/ERF_TwoStreamRadiation.H/.cpp`), owning the per-level surface and SEB fields, the flux cache and the diagnostics writer. `ERF` keeps one member and calls `define_level`, `advance` (pre- and post-dycore), `write_checkpoint` and `read_checkpoint`, each behind an explicit `solverChoice.radChoice.rad_type == RadType::TwoStream` test; every method re-tests it. The heating-rate array stays in `ERF` because RRTMGP fills the same one.
- Assertions where a broken contract would otherwise surface as a wrong number elsewhere: call site, level defined, the 2D fields laid out like the 3D state (the sweep indexes them with the state's MFIter), the heating array's shape, SEB fields present, positive finite dt; aborts naming the level and step on a non-finite state into the sweep, non-finite heating rates out of it, or a non-finite surface state after the force-restore step.
- Ghost and fill audit: every read in the sweep is at the column's own (i, j) with k inside the box; the 2D fields are read on valid cells; no `FillBoundary` or `FillPatch` is needed and none is done.

**Tests**

- `TwoStream_ColumnHeating` is registered only with MPI and not on Windows, like the other `cmake -P` tests; the RRTMGP `Radiation` regtest that the `ERF_ENABLE_RRGMTP` typo fix had re-enabled, with no input files, is additionally gated on its input directory.
- The SEB prognostic case gains a time-integration check (each step must change `T_s` by `dt` times the tendency), a `restart` mode with its own deck, and a `coupled` mode for the new flux option. All 17 canonical cases, the benchmark suite and the two CTests pass on the refactored code, with no unused `erf.radiation` or YSUNew key in any run.

**Tree hygiene**

- The development notes and the run outputs that had been committed (benchmark summaries, diagnostics CSVs, data logs) are untracked and in `.gitignore`, as #3993 did for the Canonical_RANS notes; the stray `build/cmake_uninstall.cmake` and the googletest submodule bump are gone; development-phase wording is removed from the branch's files.

## Review response (13 September)

All ten inline findings are addressed in 5efc5eb42; each thread has a reply saying what changed. In short: every domain-mean and maximum in the diagnostics is MPI-reduced, with a terrain-fitted CTest (`TwoStream_ColumnHeating_Terrain`) whose runner requires the diagnostics CSV to be identical on 1 rank and on the CI rank count; the prognostic cloud fraction no longer discards the dynamic optical depth; `cloud_fraction_qc_scale` is the saturating cloud water, `cf += min(1, qc / scale)`; layer thicknesses come from the interface heights, the top layer included; the azimuth formula is corrected (and its sine sign convention) with gtests; the per-column scratch is a per-box FArrayBox instead of ~37 kB of device stack;  the diagnostics writer persists; `ERF_SimplifiedSEB.H` is in `Make.package`; the plotfile gtest drives the real capability predicate. The same defect classes were searched for elsewhere in the model: the 46 silent input clamps in `RadChoice::init_params` now warn with the key and the value used; no other rank-local diagnostic, fixed-size device array or unused exported function was found. The driver now also refuses a finite but non-positive density or rho·theta before the sweep (a collective minimum), so the kernel's placeholder substitutions are unreachable from a run. Five gtests were added (553 pass), and the 17 canonical cases, the benchmark suite and the three CTests pass on the fixed code.

**Selector unification (Slack request, 13 September).** The branch had introduced a second selector, `erf.radiation_type`, next to development's `erf.radiation_model`. That is gone: `TwoStream` is now a value of `erf.radiation_model` (the `RadiationType` enum), `RadType` and `erf.radiation_type` no longer exist, and the `erf.radiation.*` sub-options are read only when that value is selected. The IRadiation models (RRTMGP, Simple) are gated on `SolverChoice::rad_uses_interface()`, the two-stream model on `rad_type == RadiationType::TwoStream`, and everything that any radiation model shares (heating-rate storage, the RhoTheta source, the plotfile variables) on `rad_type != None`. Every deck, README, checker and doc page uses `erf.radiation_model = "TwoStream"`.


Source/Radiation/ERF_TwoStreamRadiation.cpp
```diff
@@ -0,0 +1,949 @@
...
+        }
+        // equivalent to a single-column value for spatially UNIFORM atmospheres
+        // (as in the current SW_ClearSky_Analytical / LW_Isothermal RegTests).
+        // Cloud layer and scattering tests are ALSO spatially uniform (identical
+        // cloud/tau/scattering parameters applied to every column), so the
+        // domain-averaged value still equals the true single-column flux there.
+        // True horizontal heterogeneity (e.g., patchy clouds varying by column)
+        // remains deferred to future work.
+        if (do_sweep) {
+            if (n_columns_total > 0) {
```
asalmgren, line 961: **[high] Every "domain mean" diagnostic is rank-local; no MPI reduction.**

`sw_surface_sum`, `sw_up_toa_sum`, `lw_net_sum`, `lw_up_toa_sum` and `n_columns_total` here — and `max_heating_global` (line 649), `seb_residual_sum`/`n_seb_columns` (line 730), `t_s_sum`/`q_s_sum`/`n_prog_columns` (lines 847-851) — are accumulated over the MFIter of the *local* boxes only, then divided by the *local* column count and printed/written by the IOProcessor. There is no `ParallelDescriptor::ReduceRealSum` / `ReduceRealMax` anywhere in this file.

Run on 4 ranks with `erf.radiation.seb_enable` and a heterogeneous surface: `SW_surface`, `LW_net_surface`, `T_s_mean` and `heating_rate_max` in `radiation_diag.dat` describe rank 0's subdomain only, and change when you change the decomposition — which contradicts the "results do not depend on the horizontal decomposition" claim in the PR description.

`heating_rate_max` is also the safety diagnostic, so a runaway column on rank 3 is invisible.

Source/Radiation/ERF_TwoStreamColumn.H
```diff
@@ -0,0 +1,1127 @@
...
+                    : diagnose_tau_dynamic(tau, qv, qc, rad_choice.tau_lw_coeff_qv, rad_choice.tau_lw_coeff_qc);
+    }
+
+    // Prognostic cloud fraction: scale the cloud-band enhancement by cf(k).
+    if (rad_choice.cloud_fraction_prog_enable && cloudy &&
+        rad_choice.tau_profile_type == TauProfileType::CloudLayer &&
+        is_cloud_level(z_center, rad_choice)) {
+        amrex::Real cf_prog = diagnose_cloud_fraction_prognostic(i, j, k, state_arr, rad_choice);
+        amrex::Real tau_clear_k = tau_layer_value(z_center, base, rad_choice, /*apply_cloud=*/false);
+        tau = tau_clear_k + cf_prog * rad_choice.cloud_tau_per_layer;
```
asalmgren: **[medium] The prognostic-cloud-fraction branch throws away the dynamic moisture optical depth.**

`diagnose_layer_tau` computes `tau` including `diagnose_tau_dynamic(...)`, and then this line unconditionally overwrites it with `tau_clear_k + cf_prog * cloud_tau_per_layer`, where `tau_clear_k` is the *base* value computed with no dynamic term.

With `tau_sw_dynamic_enable = true`, `cloud_fraction_prog_enable = true` and `tau_profile_type = cloud_layer`, every layer inside the cloud band silently loses its `coeff_qv*qv + coeff_qc*qc` contribution. The doc comment says the cloud fraction only "replaces the cloud-band enhancement", so code and documented intent disagree.

The same loss propagates into the mass model, where `additions = diagnose_layer_tau(..., 0.0, ...) - band` then evaluates to just the aerosol term.

Suggest adding the dynamic term back onto `tau_clear_k`, or computing the replacement relative to the already-computed `tau` rather than re-deriving from the base.

Source/Radiation/ERF_PrognosticCloudFraction.H
```diff
@@ -0,0 +1,147 @@
...
+        if (cf_rh > 1.0) cf_rh = 1.0;
+        cf += cf_rh;
+    } else if (rh >= rh_min) {
+        // rh_max == rh_min: step function
+        cf += 1.0;
+    }
+
+    // qc contribution: scaled by qc_scale
+    if (qc_scale > 0.0) {
+        amrex::Real cf_qc = qc_scale * qc;
```
asalmgren: **[medium] `cloud_fraction_qc_scale` is applied as a multiplier, making the qc term dead at its default.**

`cf_qc = qc_scale * qc` with the default `qc_scale = 1.0e-3` and a realistic `qc ~ 1e-3 kg/kg` gives `cf_qc = 1e-6` — i.e. no contribution at all.

`ERF_RadStruct.H:549` describes this as the factor that "converts qc [kg/kg] to dimensionless cloud fraction contribution", which implies `qc / qc_scale` (a threshold), not `qc * qc_scale`. As written, a user who turns on `cloud_fraction_prog_enable` for a cloudy column gets only the RH ramp; the liquid-water path is ignored.

Source/Radiation/ERF_TwoStreamColumn.H
```diff
@@ -0,0 +1,1127 @@
...
+    // ------------------------------------------------------------------
+    amrex::Real dz_level[MAX_RAD_LEVELS];
+    amrex::Real z_level[MAX_RAD_LEVELS];
+
+    for (int m = 0; m < nlev; ++m) {
+        dz_level[m] = dz_uniform;
+    }
+    if (z_phys_cc) {
+        for (int m = 0; m + 1 < nlev; ++m) {
+            amrex::Real dz_computed = z_phys_cc(i, j, kmin + m + 1) - z_phys_cc(i, j, kmin + m);
```
asalmgren: **[medium] Layer thickness on a non-uniform grid is the cell-centre spacing, not the layer thickness.**

`dz_computed = z_phys_cc(i,j,kmin+m+1) - z_phys_cc(i,j,kmin+m)` is the centre-to-centre distance, but `dz_level[m]` is then used both as the layer *thickness* in the flux divergence `dF/(rho*cp*dz)` — where `dF` is taken across the layer's two *interfaces* — and as the mass path `rho*dz` for the `tau_model = mass` optical depths.

On a stretched grid (`erf.terrain_z_levels`) those two differ by the stretch ratio, so heating is misallocated between layers and the column optical depth is biased.

`z_phys_nd` carries the true interface heights and is available at the call site.

Source/Radiation/ERF_TwoStreamColumn.H
```diff
@@ -0,0 +1,1127 @@
...
+            if (dz_computed > 0.0 && std::isfinite(dz_computed)) {
+                dz_level[m] = dz_computed;
+            }
+        }
+        // Top layer: no level above; reuse the thickness of the layer below.
+        if (nlev > 1) dz_level[nlev - 1] = dz_level[nlev - 2];
```
asalmgren: **[medium, same finding as above] The topmost layer's thickness is copied from the layer below.**

On a stretched grid this is simply wrong — and the top layer is the one the regression checker requires to be the strongest LW cooler, so the check is validating a thickness that doesn't correspond to the grid. Deriving the top layer's thickness from `z_phys_nd` interfaces fixes both this and the centre-spacing issue above.

Source/Radiation/ERF_SolarGeometry.H
```diff
@@ -0,0 +1,355 @@
...
+    amrex::Real sin_zenith = std::sin(zenith_rad);
+    amrex::Real cos_zenith = std::cos(zenith_rad);
+
+    // Guard against division by zero when sun is directly overhead (zenith ≈ 0)
+    // In this case, azimuth is undefined; we return 0 (north) as convention
+    if (sin_zenith < 1.0e-6) return 0.0;
+
+    // Compute numerator and denominator for sine/cosine of azimuth
+    amrex::Real sin_azimuth = std::sin(hour_angle_rad) * std::cos(declination_rad) / sin_zenith;
+    amrex::Real cos_azimuth = (std::sin(declination_rad) - sin_zenith * std::sin(latitude_rad))
```
asalmgren: **[medium] `compute_solar_azimuth_angle` has sin and cos of the zenith angle swapped.**

The code computes

```
cos_az = (sin(decl) - sin(Z)*sin(lat)) / (cos(Z)*cos(lat))
```

The standard formula is

```
cos_az = (sin(decl) - cos(Z)*sin(lat)) / (sin(Z)*cos(lat))
```

For lat 40N at solar noon on the solstice this returns a value outside [-1, 1] and the subsequent `atan2` result is meaningless.

The function is currently unused (`git grep` finds only the definition), so there is no runtime impact today — but it's exported alongside the zenith helpers and the doc comment repeats the wrong formula, so the next caller inherits the bug.

Source/Radiation/ERF_TwoStreamColumn.H
```diff
@@ -0,0 +1,1127 @@
...
+    const amrex::Real cp_air = Cp_d;              // Dry-air specific heat at constant pressure [J/(kg·K)]
+
+    // ------------------------------------------------------------------
+    // Layer thickness dz_level[m] and layer-center height z_level[m] above
+    // the surface. Both come from the cell-centered physical heights when
+    // z_phys_cc is available (stretched or terrain-following grids), and
+    // from the uniform spacing otherwise. z_level locates the cloud band
+    // and evaluates the exponential aerosol profile.
+    // ------------------------------------------------------------------
+    amrex::Real dz_level[MAX_RAD_LEVELS];
```
asalmgren: **[medium] ~37 KB of per-thread stack arrays inside a GPU device lambda.**

With `MAX_RAD_LEVELS = 512`, the arrays `dz_level`, `z_level`, `tau_sw_level`, `F_dir`, `A`, `S`, `F_lw_up_iface`, `F_lw_down_iface` and `tau_lw_level` total roughly 37 KB per thread — and `vertical_two_stream_sweep` is `AMREX_FORCE_INLINE`d twice into the same lambda (clear + cloudy), so the frame may double.

In a CUDA build every one of those accesses is local memory. Occupancy collapses, and the GPU path will likely be slower than the CPU one, defeating the cost comparison in the PR description. The arrays are also allocated at full capacity regardless of the actual `nlev`.

Worth either sizing to a realistic maximum, or hoisting the scratch into a device-allocated `FArrayBox`/`Gpu::DeviceVector` indexed by column.

Source/TimeIntegration/ERF_Advance.cpp
```diff
@@ -438,6 +438,16 @@ ERF::Advance (int lev, double time, double dt_lev, int iteration, int /*ncycle*/
         Time_Avg_Vel_atCC(dt[lev], t_avg_cnt[lev], vel_t_avg[lev].get(), U_new, V_new, W_new);
     }
 
+    // ***********************************************************************************************
+    // Two-stream radiation, post-dycore call: reports the cached flux
+    // diagnostics and advances the force-restore surface state by dt_lev.
+    // No column sweep runs here; that happened in advance_radiation above.
+    // ***********************************************************************************************
+    if (solverChoice.radChoice.rad_type == RadType::TwoStream) {
+        two_stream_rad.advance(lev, iteration, time + dt_lev, dt_lev, "post_dycore",
```
asalmgren, line 447: **[low] This call and the pre-dycore one are gated differently, so setting both solvers half-runs the two-stream model.**

The pre-dycore call in `ERF_AdvanceRadiation.cpp:113` is an `else if` on the RRTMGP branch; this post-dycore call is gated only on `radChoice.rad_type == RadType::TwoStream`.

Set `erf.radiation_model` (RRTMGP) *and* `erf.radiation_type = TwoStream` with `seb_prognostic_enable`: the column sweep never runs, `m_flux_diag[lev]` stays at its zero-initialised value, and this call still advances the force-restore surface state every step from those zeros — and, with `seb_use_radiation_fluxes`, from `m_sw_flux_sfc` values that are never written.

`RadChoice::init_params` warns about `radiation_type = RRTMGP` but says nothing about this combination. Either abort there, or give this call the same `rad_type == None` condition the pre-dycore branch effectively carries.

Source/Radiation/ERF_RadiationDiagnostics.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+  bool is_pre_site = (call_site.find("pre") != std::string::npos);
+  bool is_post_site = (call_site.find("post") != std::string::npos);
+
+  // Check if this call site should be filtered based on mode
+  if ((is_pre_site && !should_emit_pre) || (is_post_site && !should_emit_post)) {
+    return;
+  }
+
+  // Guard against duplicate writes using 3-tuple identity:
```
asalmgren, line 103: **[low] The duplicate-write guard is dead code.**

`RadiationDiagnostics rad_diag(...)` is constructed fresh inside every `TwoStreamRadiation::advance()` call (`ERF_TwoStreamRadiation.cpp:266`), so `m_last_write_step` / `m_last_write_call_site` / `m_last_write_time` always start at their defaults and this `(step, call_site, time)` dedup never fires.

`m_header_written` resets the same way, so the `ifstream` existence probe runs on every CSV row.

Either make the logger a member of `TwoStreamRadiation`, or drop the dedup state along with the `diag_dedup_tol` input it reads.

Source/Radiation/Make.package
```diff
@@ -0,0 +1,11 @@
...
+CEXE_sources += ERF_TwoStreamRadiation.cpp
+
+CEXE_headers += ERF_TwoStreamRadiation.H
+CEXE_headers += ERF_TwoStreamSW.H
+CEXE_headers += ERF_TwoStreamLW.H
+CEXE_headers += ERF_TwoStreamColumn.H
+CEXE_headers += ERF_RadiationDiagnostics.H
+CEXE_headers += ERF_PrognosticCloudFraction.H
+CEXE_headers += ERF_AerosolOpticalDepth.H
+CEXE_headers += ERF_SolarGeometry.H
```
asalmgren, line 11: **[low] `ERF_SimplifiedSEB.H` is missing from `CEXE_headers`.**

It's included by `ERF_TwoStreamRadiation.cpp`, and every other header in `Source/Radiation/` is listed here, so the GNUmake build doesn't track it for install/dependency listing.

(`ERF_RadiationCoupling.H` from the intermediate commits is correctly absent from the final tree.)

Tests/Unit/IO/ERF_GTestPlotfileSelection.cpp
```diff
@@ -483,6 +483,32 @@ TEST(Plotfile3DSelection, OptionalStorageGroupsAreExplicit)
     EXPECT_TRUE(erf_plotfile::plot3d_fixed_variable_available("walldist", caps));
 }
 
+// Motivation: verify that radiation heating variables (qsrc_sw, qsrc_lw)
+// are available when either RRTMGP or TwoStream radiation is active, but not when
+// radiation is disabled. This regression test ensures the capability check correctly
+// recognizes both solverChoice.rad_type (RRTMGP) and solverChoice.radChoice.rad_type
+// (TwoStream) as sources of radiation heating.
+TEST(Plotfile3DSelection, RadiationHeatingStorageRecognizesTwoStreamAndRRTMGP)
```
asalmgren: **[low] This regression test doesn't exercise the thing it claims to.**

`RadiationHeatingStorageRecognizesTwoStreamAndRRTMGP` sets `caps.radiation_heating_storage = true` by hand for both the "RRTMGP" and "TwoStream" cases and then asserts the same predicate twice. It never touches `setPlotVariables`, which is where the new `|| radChoice.rad_type == RadType::TwoStream` mapping actually lives — so the test would still pass if that line were reverted.

To be a regression test it needs to drive `setPlotVariables` (or whatever factory populates `caps`) from a `SolverChoice` with `radChoice.rad_type = TwoStream`, and assert that `radiation_heating_storage` comes back true.

asalmgren (commented):
## Review: two-stream radiation + simplified SEB

Reviewed the ~6,400 lines of new code under `Source/`, `Tests/`, `CMake/`, `Exec/Make.ERF`. The remaining ~75k lines (canonical test decks, Python checkers, docs) were skimmed rather than audited line by line.

**The core physics holds up.** The Meador-Weaver layer solution and the adding-method upward/downward passes match the RRTMGP formulation, including the `src(sfc) = alpha_dir * F_dir(0)` boundary condition and the placement of the `denom = 1 - R_dif * A` factor. The LW downward-then-upward ordering correctly precedes the gray-surface reflection term. Orientation (k=0 at the surface), the Exner conversion, and the once-per-slow-step injection into `make_sources` all look right.

10 findings inline: 1 high, 5 medium, 4 low. The high one is #1 — it's the claim the PR makes about itself in the description.

Also verified as correct: the diffuse/direct albedo split vs. RRTMGP's `sw_solver_2stream`; `&hetero_alb_sw_arr` taken inside the `[=]` device lambda refers to the closure copy rather than a host stack address; `ba2d[lev]` has `setRange(2,0)` and preserves box order, so `(i,j,0)` indexing of the 2D fields against the 3D state's MFIter is valid; `std::swap(vars_old, vars_new)` happens at the top of `ERF::Advance` so the post-dycore call sees the intended old state; `iteration` equals `istep[lev]`, so the two CSV call sites agree on the step index; `read_checkpoint` is placed after `MakeNewLevelFromScratch` so `define_level` has run on restart; `get_qv_from_state`/`get_qc_from_state` guard correctly against `RhoQ1_comp`/`RhoQ2_comp` when moisture is off; and the `ERF_ENABLE_RRGMTP` -> `ERF_ENABLE_RRTMGP` typo fix uses the real option name from `CMakeLists.txt:141`.

Source/Radiation/ERF_TwoStreamRadiation.cpp
hgopalan: Fixed in 5efc5eb42. Every accumulator formed over the MFIter (the four flux sums, the column count, max heating, the SEB residual sum/max/count, the prognostic T_s/q_s sums/maxima/count) is now reduced with ParallelDescriptor::Reduce{Real,Long}{Sum,Max} before the means and maxima are formed; the collectives sit under input-driven conditions that are identical on all ranks.

To keep it that way, RunTwoStreamRadiation.cmake now also runs the deck on one rank and requires the diagnostics CSV to match the NRANKS run byte for byte, and a new CTest TwoStream_ColumnHeating_Terrain runs the column over a Witch-of-Agnesi hill on a terrain-fitted mesh with two boxes of unequal width, so the columns differ and rank 0's share has a different mean from the whole. On the previous binary the two CSVs differ from the first row (3.193070e+02 vs the global SW_surface); on this one they are identical. A centred hill on a 4x4 domain did not discriminate: the two boxes are mirror images.

Source/Radiation/ERF_TwoStreamColumn.H
hgopalan: Fixed in 5efc5eb42. The cloud-band enhancement is now scaled in place, tau += (cf - 1) * cloud_tau_per_layer, on the tau that already carries the dynamic term, so nothing is re-derived from the clear-sky base and the mass model's additions term keeps the moisture contribution. gtest DynamicOpticalDepthSurvivesPrognosticCloudFraction checks that switching the dynamic term on changes a cloud-band layer by exactly coeff_qv * qv with the prognostic cloud fraction active, and that it equals the fixed-cloud-fraction value at cf = 1.

Source/Radiation/ERF_PrognosticCloudFraction.H
hgopalan: Fixed in 5efc5eb42: cf += min(1, qc / cloud_fraction_qc_scale), so the input is the cloud water at which the liquid-water term alone saturates (1e-3 kg/kg by default). The RadStruct comment, Inputs.rst, the theory page's formula and the master deck now say that; gtest CloudFractionLiquidWaterTermIsAThreshold pins 1e-3 -> 1, 5e-4 -> 0.5. The ProgCloudFraction canonical case still passes its checker.

Source/Radiation/ERF_TwoStreamColumn.H
hgopalan: Fixed in 5efc5eb42. The sweep now takes z_phys_nd: the height of face k over column (i,j) is the mean of its four nodes, the layer thickness is the distance between the layer's two interfaces, and the layer centre is their midpoint. That thickness feeds both the heating divergence and the mass path. TwoStreamRadiation::advance takes z_phys_nd instead of z_phys_cc (both call sites updated). gtest MassModelIsIndependentOfTheStretching runs a constant-density column with a geometric stretching of 1.25 against the uniform layering of the same height and requires the same SW and LW fluxes to 1e-9 of the incident flux; with the centre-to-centre thickness the mass path does not add up to the column and this fails.

Source/Radiation/ERF_TwoStreamColumn.H
hgopalan: Fixed together with the one above in 5efc5eb42: the top layer's thickness is now the distance between its own two interfaces from z_phys_nd, not a copy of the layer below. The new TwoStream_ColumnHeating_Terrain CTest runs the column-structure checker on a terrain-fitted mesh, so the top-layer cooling check now sees a thickness that belongs to the grid.

Source/Radiation/ERF_SolarGeometry.H
hgopalan: Fixed in 5efc5eb42: cos(azimuth) = (sin(decl) - cos(Z) sin(lat)) / (sin(Z) cos(lat)), and the doc comment with it. Writing the tests also showed the sin term had the wrong sign for the clockwise-from-north convention the function documents (the hour angle is negative before solar noon, so sin(azimuth) = -sin(H) cos(decl) / sin(Z)); that is fixed too, plus a guard at the pole. New Tests/Unit/Radiation/ERF_GTestSolarGeometry.cpp pins due south at solar noon at 40N on the solstice, east-of-south in the morning and west-of-south in the afternoon (symmetric), due east at equinox sunrise, and the pole. Still unused by the model, but the next caller gets the right answer.

Source/Radiation/ERF_TwoStreamColumn.H
hgopalan: Fixed in 5efc5eb42. The nine per-column arrays are now a scratch FArrayBox the driver allocates once per box (TwoStreamScratch::NCOMP = 9 components on the box grown by one cell at the top, so nlev + 1 interface entries per column) and the sweep addresses as scratch(i, j, kmin + m, field) through small reference-returning accessors, so the kernel body is unchanged. The clear and cloudy evaluations of a column share the buffer; MAX_RAD_LEVELS and its assert are gone and any nlev works; the unit-test harness allocates the same buffer. The device lambda now holds only scalars.

Source/TimeIntegration/ERF_Advance.cpp
hgopalan: Fixed in 5efc5eb42. SolverChoice::init_params aborts at start-up when erf.radiation_model and erf.radiation_type = TwoStream are both set (Inputs.rst says so), and the post-dycore call in ERF::Advance now carries the same rad_type == None condition the pre-dycore else-if branch implies, so the two gates match even without the abort.

Source/Radiation/ERF_RadiationDiagnostics.cpp
hgopalan: Fixed in 5efc5eb42: the writer is a member of TwoStreamRadiation (std::unique_ptr<RadiationDiagnostics> m_diag), created on the first advance() call, so m_last_write_* and m_header_written carry across calls and the header probe runs once.

Source/Radiation/Make.package
hgopalan: Fixed in 5efc5eb42: ERF_SimplifiedSEB.H is in CEXE_headers.

Tests/Unit/IO/ERF_GTestPlotfileSelection.cpp
hgopalan: Fixed in 5efc5eb42. The predicate is now a free function, erf_plotfile::radiation_heating_storage_available(rad_model, rad_type), which ERF::setPlotVariables uses to populate the capability, and the test drives it with the two selector enums: (None, None) -> false, (RRTMGP, None) -> true, (None, TwoStream) -> true, and the resulting capabilities through plot3d_fixed_variable_available. Dropping the TwoStream half of the mapping now fails the test.

hgopalan:
Following up on the Slack request about the two selectors: the branch now uses `erf.radiation_model` only. `TwoStream` is a value of the existing `RadiationType` enum, `RadType` and `erf.radiation_type` are gone, and the `erf.radiation.*` sub-options are read only when that value is selected. Gates: `SolverChoice::rad_uses_interface()` (RRTMGP, Simple) for everything that goes through `IRadiation`, `rad_type == RadiationType::TwoStream` for the two-stream class, `rad_type != None` for what every model shares (heating-rate storage, the RhoTheta source, `qsrc_sw`/`qsrc_lw`). All 17 canonical decks, the CTest decks, the master deck, the READMEs, the checkers and the docs use `erf.radiation_model = "TwoStream"`; a deck still carrying the old key runs without radiation and the key appears in AMReX's unused-parameter report. Verified with 553 gtests, the three CTests and the full canonical sweep, on top of today's development merge (commit 3eb26c636, merged with the branch head as e2a67adbc).

PR #3961, opened 2026-09-05
Keep the MRF scheme from raising invalid floating-point flags it never uses

hgopalan:
## Problem

Every MRF run dies at its first step under `amrex.fpe_trap_invalid = 1` in an optimised build (Apple clang, Release), on a plain neutral ABL whose results are fine, while a debug build of the same case runs clean. The canonical MRF cases all carry `amrex.fpe_trap_invalid = 0` for this reason.

The cause is the optimiser rather than the physics. Three guarded expressions in `ComputeDiffusivityMRF` are evaluated speculatively, ahead of the guard that makes them valid, and the NaN they produce is discarded but raises the invalid-operation flag the trap watches. Compiling just this file with `-ffp-model=strict` makes the trap disappear, which pins it on speculation; the debugger then identified each site from the faulting instruction and its registers.

## Changes (`Source/PBL/ERF_PBLScaleAwareBlending.H`, `Source/PBL/ERF_ComputeDiffusivityMRF.cpp`)

1. **Blending factor.** `pbl_blend_factor` computed `(dx/L)²/(1+(dx/L)²)` behind an early return for `L <= 0`. Being loop-invariant it was hoisted out of the K loop above the caller's `L > 0` test; with blending off `L = 0`, `dx/L` is inf and `inf/inf` is NaN (the registers showed the 200 m `dx` of the case divided by a zero `L`). It is now `dx²/(L²+dx²)`: the same function, a denominator of at least `dx²` for every `L`, and exactly 1 when `L <= 0`. A select around the division does not help, the division gets folded into the select's arms.
2. **Unstable stability functions.** `sqrt(max(-Ri, 0))` sat inside the `Ri <= 0` arm. Knowing `Ri <= 0` there, the optimiser removed the max and hoisted the bare sqrt above the selection, where `Ri` is positive. `min(Ri, 0)` moved the problem into the arms of the min. The argument is now `sqrt(|Ri|)`, evaluated before the selection; it equals `sqrt(-Ri)` in the arm that uses it and has nothing to fold.
3. **Unstable phi functions.** `pow(max(1 - 16 HOL, 0.01), -1/4)` inside the `L < 0` arm is the same construct; the base is now `1 + 16 |HOL|`, equal to `1 - 16 HOL` in that arm and at least 1 everywhere.

## Verification

Neutral ABL, doubly periodic, 40 x 40 x 64 on 8 km x 8 km x 2 km, input sounding 300 K and 10 m/s, `erf.pbl_type = MRF`, `dt = 0.1`.

| case | before | after |
|---|---|---|
| trap on, blending off, 10 steps | aborts at step 1 (blend factor) | runs |
| trap on, `pbl_blend_length = 750`, 10 steps | aborts at step 1 (sqrt) | runs |
| trap on, 4 ranks, 200 steps | aborts | runs |
| trap on, surface heating +3 K/h and -3 K/h, 60 steps each | aborts | runs |
| trap off, blending off: this PR vs development at step 10 | | `PLOTFILE AGREE` (bitwise) |
| trap off, blending 750 m: this PR vs development at step 10 | | `PLOTFILE AGREE` |
| `Exec/CanonicalTests/ABL/MRF_Enhancements/blending_active`, this PR vs development at step 5 | | `PLOTFILE AGREE` |

YSUNew and MYNN2.5 run the same case with the trap on before and after. YSU could not be checked on this grid (its 10 m reference height is below the first cell centre); it has a similar `sqrt(-Ri)` inside an `Ri < 0` branch in `ERF_ComputeDiffusivityYSU.cpp` that may need the same treatment.

### DIFF
--- a/Source/PBL/ERF_ComputeDiffusivityMRF.cpp
+++ b/Source/PBL/ERF_ComputeDiffusivityMRF.cpp
@@ -351,10 +351,14 @@ ComputeDiffusivityMRF (const MultiFab& xvel,
             // HOL computed from predictor height (WRF uses predictor height for wstar/VPERT)
             const Real HOL = sf * pblh_pred_arr(i, j, 0) / obuk_val;
             const Real HOL_bounded = amrex::max(amrex::min(HOL, Real(100.0)), Real(-100.0));
+            // |HOL|, so the base 1 + 16*HOL_abs of the unstable arm below equals its
+            // 1 - 16*HOL there and is at least 1 for every HOL; a guard written inside the
+            // arm gets folded away by the optimiser and the pow hoisted (see sqrt_neg_Ri).
+            const Real HOL_abs = std::abs(HOL_bounded);
             const Real one_quarter = Real(0.25);
             const Real phiM = (obuk_val > 0)
                             ? (1 + 5 * HOL_bounded)
-                            : std::pow(amrex::max(1 - 16 * HOL_bounded, Real(0.01)), -one_quarter);
+                            : std::pow(1 + 16 * HOL_abs, -one_quarter);
             const Real phiM_safe = amrex::max(phiM, Real(0.01));
 
             // wstar = u* / phi_m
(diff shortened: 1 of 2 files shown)


hgopalan:
Removed the extra blank lines in the comments and cut the comment in ERF_PBLScaleAwareBlending.H down to one line (93f36ef).

Issue #3962, opened 2026-09-08
NATIVE_SHOC: GPU and CPU results diverge far above roundoff


asalmgren:
## Summary

Running the GABLS1 ABL case with `erf.pbl_type=NATIVE_SHOC`, a CUDA run and a
pure-CPU run diverge by ~1e-5 relative in velocity and ~1e-6 relative in the
thermodynamic fields after only 50 steps. The identical comparison with
`erf.pbl_type=MYNN25` stays at ~1e-10, i.e. at amplified-roundoff level.

The cause is **not** a race, an uninitialized read, or a reduction-ordering
problem. It is a strict `>` comparison against an exact clamp floor inside the
SHOC energy fixer. That comparison decides an integer (`shoc_top`), and a
1-ulp host/device difference flips it, which changes both the magnitude and the
vertical extent of the energy-fixer temperature increment. SHOC therefore
re-injects an O(1e-9) relative perturbation on essentially every step, whereas
MYNN's differences start at true roundoff.

## Reproducer

Base input file: `Exec/CanonicalTests/ABL/inputs_GABLS1_mynn25`
(2 x 2 x 64 cells, 400 m cube, dry, stable, `zlo.type = surface_layer`).

```
mpiexec -n 4 ./ERF3d.gnu.TEST.MPI.CUDA.NC.ex inputs_GABLS1_mynn25 \
    erf.plot_file_1=ABL-SHOC_plt erf.check_file=ABL-SHOC_chk \
    amr.checkpoint_files_output=0 max_step=50 \
    amr.restart=chk00000_for_restart erf.fixed_mri_dt_ratio=4 \
    erf.pbl_type=NATIVE_SHOC \
    amrex.the_arena_init_size=0 amrex.the_arena_is_managed=true
```

`fcompare` of the CUDA plotfile against the CPU plotfile at step 50:

### `erf.pbl_type=NATIVE_SHOC`

| variable   | absolute error | relative error |
|------------|----------------|----------------|
| density    | 7.829528579e-07 | 5.922658378e-07 |
| rhoKE      | 1.587752609e-05 | 1.351273851e-06 |
| x_velocity | 0.0001405523562 | 1.749842017e-05 |
| y_velocity | 2.637293507e-07 | 2.532405919e-05 |
| z_velocity | 1.983492638e-05 | 0.05998526786   |
| theta      | 0.0001829564187 | 6.827578875e-07 |
| pressure   | 0.01551786579   | 1.540078535e-07 |
[...]

PR #3963, opened 2026-09-08
Keep the SHOC energy-fixer active top off a measure-zero comparison (…

asalmgren:
…#3962)

Above the boundary layer, both the explicit TKE update and the implicit TKE solve clamp TKE to exactly shoc::constants::min_tke(), so the bare "tke > min_tke()" test in diagnose_active_top() was settled by the last bit of the tridiagonal solution's exponentially decaying tail. Host and device builds differ there by one ulp, since the CUDA build contracts to FMA (--use_fast_math is on by default in the GNUmake path) and the host build does not. That flipped the diagnosed active top from step to step, changing both the magnitude of delta_tabs and the depth over which the energy fixer spreads it.

For the GABLS1 ABL case with erf.pbl_type=NATIVE_SHOC this made a CUDA run and a pure-CPU run differ by ~1e-5 relative in velocity after 50 steps, against ~1e-10 for the same comparison with MYNN25.

Compare against a threshold carrying a small relative margin above the floor instead. The margin is precision-aware so it stays well clear of one ulp in single-precision builds, and both the Array4 and Vector overloads of diagnose_active_top() use it so the device path and the unit-tested host path cannot diverge.

### DIFF
--- a/Source/PBL/Shoc/ERF_ShocConstants.H
+++ b/Source/PBL/Shoc/ERF_ShocConstants.H
@@ -9,6 +9,31 @@ namespace shoc::constants {
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 amrex::Real min_tke () noexcept { return amrex::Real(4.0e-4); }
 
+// Relative margin above min_tke() used to decide whether a level is still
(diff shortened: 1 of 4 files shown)


pressel (approved):
Thanks, @asalmgren. This looks good to me.

PR #3969, opened 2026-09-09
RemakeLevel Asymmetries

AMLattanzi:
# Summary
The following PR addresses asymmetries in the remakelevel pathway versus the makenewlevelfromcoarse. 

## Key changes
1. LSM needs to be initialized for all pathways
2. Dz_min needs to be initialized for all pathways
3. Set BA and set DM are called for all three cases at the top so any future routines added cannot use an empty or uninitialized BA/DM
4. When remaking a level, the define FillPatcher needs to be called for the current `lev` to register the new fine data. However, if `lev+1` exists, define fillpatcher must also be called for `lev+1` to register the correct coarse data.
5. Fix velocity projection with anelastic and a regrid (present for make from coarse but not remake).

### DIFF
--- a/Source/ERF.H
+++ b/Source/ERF.H
@@ -917,6 +917,9 @@ private:
     // Create the physbcs objects
     void make_physbcs (int lev);
 
+    // (Re)build the land surface model on this level's current grids
+    void make_lsm_at_level (int lev);
+
     // Initialize the microphysics object
     void initializeMicrophysics (const int&);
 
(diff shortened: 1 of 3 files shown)


AMLattanzi:
@asalmgren This allows the Joplin tornado to run with dynamic AMR. However, the `max_helicity` criteria will still cause issues. Specifically, it leads to a box array that contains boxes whose vertical extent is not the full domain. Since the Joplin WRF files use odd nz (like many WRF input files) the blocking factor, max grid size, and buffer, can't perfectly alleviate this issue. We should follow this PR up with a robust fix for the noted problem.

PR #3971, opened 2026-09-10
One-equation k RANS: wall condition, closure fixes, terrain wall distance, Canonical_RANS regression cases

hgopalan:
## One-equation k RANS: a minimal working model with regression cases

This PR makes the Axell & Liungman (2001) one-equation RANS closure (`erf.rans_type = kEqn`) a working, tested option on flat and terrain-fitted meshes, and adds `Exec/CanonicalTests/Canonical_RANS`: five cases laid out like `Canonical_LES`, each with a Python check script that compares numbers against stated targets, nine CTest entries (label `rans`), seven gtests, and a theory page. The development record with per-phase findings is `Exec/CanonicalTests/Canonical_RANS/PLAN.md`; the numbers are in `RESULTS.md`.

Every behaviour change is opt-in except two bug fixes noted below; the existing kEqn deck (Askervein) only gains `erf.dirichlet_k = true`.

### Fixes

| defect | before | after |
| --- | --- | --- |
| Wall value of k (`dirichlet_k`) was written into the first cell once per step, then the ext_dir BC applied a one-sided flux against a zero ghost cell and drained a fifth of it every step | KE(0)/u*^2 = 2.60 against the AL01 value 3.23 | 3.232; the value is held through every RK stage and in the implicit solve, wall flux of k is zero |
| Burchard & Petersen smoothing of Rt cancelled catastrophically | -2 at Rt = -1e16, +1.4e14 at -1e30 | exact rearrangement, holds Rt_min |
| Unstable length iterated the fixed-point map of AL01 Eq. 26, which has no fixed point in strong convection | length unbounded | Eq. 28 evaluated once with the smoothed Rt, bounded at about 1.31 l_g (matches the Kynema KLAxell implementation) |
| Poisson wall distance took the low-face fluxes as the cell gradient, half a cell off centre | every distance short by z dz/(2H), 0.8 % on 64 cells | cell-centred metric gradient: exact to 1e-6 m on a flat fitted mesh |

### New options (all default off)

`erf.tke_floor`, `erf.implicit_tke_dissipation`, `erf.rans_consistent_diffusivities`, `erf.rans_lscale_from_pblh` + `erf.rans_lscale_min`, `erf.wall_dist_type = poisson | terrain_height`; input validation for the closure constants (the stability functions have poles near Rt = -3.6); a start-up warning when kEqn runs under a surface layer without `dirichlet_k`; plot variables `Rt`, `cmu`, `cmu_prime`.

The `terrain_height` wall distance (height above the local surface on its normal, no linear solve) is exact to 1e-10 on a flat fitted mesh and closer to the true distance than the Poisson solve on both hills (mean 0.02 % and 0.01 % against 1.0 % and 0.3 %); the hill decks use it, and `_Poisson` CTest variants keep the Poisson path covered.

### Cases and results

| case | run | key results |
| --- | --- | --- |
| Neutral_ABL_Flat | 12 h Ekman layer | log law within 7 %, KE(0)/u*^2 = 3.232, 1 and 2 ranks agree to 2e-15 |
| Stable_ABL_Flat | GABLS1, 9 h | u* 0.244, jet 1.23 Ug at 154 m, depth 134 m (LES ensemble: 0.26-0.30, 150-200 m) |
| Convective_ABL_Flat | 0.24 K m/s, 4 h | column heat budget closes to 0.02 %, inversion at 1020 m, AL01 buoyancy term at the wall (3.82) |
| Neutral_Hill_2D | Witch of Agnesi ridge, 6 h | crest speed-up 0.55/0.45/0.39 vs Jackson-Hunt 2h/L = 0.4, upstream log law within 4 % |
| Neutral_Hill_3D | radial hill, 4 h | crest speed-up 0.41/0.28/0.22 vs 1.6 h/L = 0.32, upstream log law within 10 % |

Restart is bit-exact on the flat and both terrain decks.

### Findings outside this PR's scope

- Under the anelastic integrator every vertical diffusion is explicit (the implicit column solve is switched off for anelastic, PR #3329), so RANS runs are bound by dz^2/(2K): the convective deck aborts at dt = 5 s with K reaching 40 m2/s and runs at 2 s. The column tridiagonal works with the RANS diffusivities (compressible, explicit vs implicit at the same dt: 3e-6 m/s in wind). Enabling it under anelastic is planned separately (PLAN.md phases 9-10).
- The 1.788e139 pre-projection divergence first seen on the 3D hill deck at dz different from dx was not the aspect ratio: that mesh splits the BoxArray in z, and the initial projection read unfilled momenta ghost faces at the internal box faces while the planar surface-layer arrays were duplicated across the stacked boxes. Both are fixed in #3970. The same class of read was in this PR's `terrain_height` wall distance (surface nodes taken from boxes above the split, 220 m error); it now gathers the surface slab onto every box, with a guard. The Poisson wall-distance multigrid does diverge at dx = 2 dz with or without the split, so the `_Poisson` CTest variants keep dz = dx.
- A local-K closure keeps a superadiabatic lapse of about -2 K/km through a convective mixed layer (1.1 K spread where LES gives under 0.3 K); the check bounds it at 2 K.

### Limitations

Dry buoyancy only; no hybrid RANS-LES; the Poisson wall distance is not available with EB or thin bodies; the wall distance is not recomputed after a regrid.

### Verification

`ctest -L rans` (9 entries in an FFT build) and `erf_unit_tests --gtest_filter=RANSClosure*` (7 tests) pass; the Release build with all warnings on is clean in the changed files; single-precision build passes; codespell clean; `sphinx-build` reports no error on the touched pages.

**CI note (2026-09-11).** No CI configuration builds with `ERF_ENABLE_FFT`, and the decks use the FFT projection, so every RANS entry aborted at start-up in the 2026-09-10 CI run (the harness also swallowed the run log). The three flat entries now pass `erf.use_fft=false` (MLMG projection) and run in every build; the six terrain-fitted entries need the FFT-preconditioned general-terrain projection and are registered only under `if(ERF_ENABLE_FFT)`; the test command echoes the tail of the run log when the executable fails. Checked with a Debug build without FFT (3 entries registered and passing), a configure with MPI off (3 entries), and a Release FFT build (9 entries and the 7 gtests passing).


asalmgren:
Hi @hgopalan — I reviewed this and pushed one commit (98970c1a) directly to `claude-RANS` since maintainer-edits are on. Happy to back it out and hand you a patch instead if you'd rather apply it yourself. Details below so nothing is a surprise.

### `erf.dirichlet_k` without a surface layer (the substantive one)

`pin_klo` in `ImplicitDiffForStateLU_{N,S,T}` was gated on `rans_type == kEqn && dirichlet_k`, but the matching pin in `erf_slow_rhs_post` also requires `SurfLayer != nullptr`. Before this PR the Dirichlet-k handling lived entirely inside the `ERF_BC::surface_layer` branch of `init_bcs`, so `erf.dirichlet_k = true` with `zlo.type = NoSlipWall`/`SlipWall` was a no-op. With the new code that combination is quietly wrong: nothing ever writes the AL01 Eq. 16 value (`SurfaceLayer::update_fluxes` is never constructed), `erf_slow_rhs_post` doesn't pin, but the implicit solve still collapses the `klo` row to `x(klo) = phi(klo)` — vertical TKE diffusion is off in the first cell while advection, production and dissipation keep updating it.

I did **not** fix this by adding `use_SurfLayer` to the condition: that flag is hardcoded `false` for the TKE solve (`ERF_ImplicitPost.H:21`), so it would have made `pin_klo` never fire and disabled the feature. A BC-based gate doesn't work either — `surface_layer` sets RhoKE's zlo BC to `foextrap`, same as the wall types.

Instead `init_bcs` now errors if `dirichlet_k = true` without `zlo.type = surface_layer`. Since `m_SurfaceLayer != nullptr` ⟺ that BC (`ERF.cpp:1247`), the two paths agree by construction with no signature churn. Every deck here already sets both, so nothing breaks. There's a note at each `pin_klo` so neither side gets relaxed alone.

If you'd rather *support* `dirichlet_k` on a plain wall, that's a real feature and needs a wall value from somewhere other than MOST — happy to discuss, but it seemed out of scope for this PR.

### Wall value drifted with density

The re-imposition copied the conserved variable, `cur_cons(RhoKE) = old_cons(RhoKE)`, but `cur_cons(Rho_comp)` is already updated at that point, so the retained value was `k_wall · ρ_old/ρ_new` — Eq. 16 defines the primitive `k`. Now `cur_cons(RhoKE) = cur_cons(Rho) * old_cons(RhoKE)/old_cons(Rho)`. A no-op in the anelastic decks (ρ ≈ const), but it matters for a compressible run and for `l_moving_terrain`, where `cur_cons` was additionally rescaled by `detJ/detJ_new` just above. Both states carry the same `detJ` convention, so the ratio cancels it exactly.

### Docs

`Rt`, `cmu` and `cmu_prime` are user-typeable plotfile names, so they're now in `Plotfile3DReference.rst` — the variable table plus prose noting they need `use_kturb` and are identically zero unless the k-eqn closure is running.

In `Inputs.rst`, the `erf.rans_consistent_diffusivities` cell was split across two table lines. Inline markup *can* span lines inside a grid cell, so Sphinx emits no warning — it just renders as bold `erf.rans_consistent_diffusivit-` + newline + `ies`, i.e. a stray hyphen and space in the middle of the parameter name someone is about to copy. Widened the first column and joined the cell; `git diff -w` shows that row is the only non-whitespace change.

### Deck comments

`inputs_hill` said `erf.use_fft = false  # terrain-fitted mesh: MLMG projection`. On `MeshType::VariableDz` the projection never reads `use_fft` — `ERF_PoissonSolve.cpp:722` goes straight to `solve_with_gmres` and `#else`-aborts with "Rebuild with USE_FFT = TRUE". That's exactly why `CTestList.cmake:952` gates these tests on `ERF_ENABLE_FFT`, so the deck and the CMake comment contradicted each other. The line is a no-op today and becomes actively wrong if anyone flattens the deck to `ConstantDz`; it now says so.

Separately, the 2D deck header, both `check_hill*.py` tolerance comments, `check_flat_fitted.py` and `Neutral_Hill_2D/README.md` all described the Poisson wall distance as what runs, while the decks set `wall_dist_type = terrain_height` and `_Poisson` is a CTest override. The tolerances are unchanged — they cover both paths and are set by the looser one — the comments now say that. The README's opening paragraph also contradicted its own line 24, and its parameter table repeated the MLMG error.

### What I didn't touch

`PLAN.md` and `RESULTS.md` read as development notes rather than case documentation — open TODOs ("Decide default: `dirichlet_k = true` when…"), and `PLAN.md` describes the `dirichlet_k` ghost-cell handling differently from what shipped. Your call whether they belong in a user-facing `Exec/` directory; I left them alone.

### Verification

Compile-level only: the four changed `.cpp` files pass `-fsyntax-only` against this branch's headers, the check scripts pass `py_compile`, and both RST files parse clean under docutils. **I did not build ERF or run the Canonical_RANS cases.** The two things worth watching in CI are the `k`-vs-`ρk` pin (expected to be a no-op in these anelastic decks, but that's reasoning, not a test) and the new `init_bcs` abort.

Also checked and found correct, in case it saves you re-reading them: the `constB`/`rhs` sign flip in the wall-distance solve is algebraically a no-op and moves MLABecLaplacian to its documented positive-definite form; the `smooth_Rt` rewrite is exactly the old Burchard–Petersen expression and the dropped `std::max` was provably unreachable; the trapezoidal implicit-dissipation `1/2` is right because the reconstructed stage-1 increment carries the other half; dropping the `ngc` growth in `ComputeTurbulentViscosityRANS` is safe and fixes an OpenMP tile race; and the `mfi`-indexed `pblh`/`z_0` access is valid on z-split BoxArrays.

PR #3973, opened 2026-09-10
Keep 1/theta_ref from tripping fpe_trap_zero when theta_ref is unset

hgopalan:
## Defect

With `amrex.fpe_trap_zero = 1`, a Release build aborts in two places when `erf.theta_ref` is left unset (its default is 0). On macOS the FP trap reports as SIGILL.

- Step 1 in `DiffusionSrcForState_S`, called from `erf_slow_rhs_pre`, e.g. `Exec/CanonicalTests/ABL/mrf_unstable`. This affects any deck with diffusion.
- During `InitData_post` in `SurfaceLayer::update_fluxes`, with `erf.rans_type = kEqn` and `erf.dirichlet_k = true`.

The same runs complete with `fpe_trap_invalid` or `fpe_trap_overflow` alone. A Debug build does not trap.

## Cause

Four places compute

```cpp
const Real l_inv_theta0 = (use_ref_theta) ? one / theta_ref : one;
```

At -O3, clang turns the select into an unconditional `fdiv` followed by `fcsel`:

```
fmov  d0, #1.0
fdiv  d1, d0, d11      <- trap, d11 = theta_ref = 0
fcmp  d11, #0.0
fcsel d12, d1, d0, gt
```

The 1/0 result is discarded, but it still raises FE_DIVBYZERO. lldb stopped at this instruction with `d11 = 0`. Compiling only `ERF_DiffusionSrcForState_S.cpp` with `-ffp-model=strict` and relinking removes the trap. In a standalone -O3 test, GCC 14 kept a real branch for this pattern, so GCC builds may not show the defect.

| Site | Reached by |
|---|---|
| `Source/Diffusion/ERF_SetupDiff.H` (included by the S, N, T and EB diffusion kernels) | every run with diffusion |
| `Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp`, Deardorff branch | `les_type = Deardorff` |
| `Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp`, k-eqn RANS branch | `rans_type = kEqn` |
| `Source/BoundaryConditions/ERF_SurfaceLayer.cpp`, k-eqn Dirichlet update | `rans_type = kEqn`, `dirichlet_k = true` |

## Fix

Before the selection, clamp the divisor to the smallest normal `Real`:

```cpp
const bool use_ref_theta = (turbChoice.theta_ref > 0);
const Real inv_theta_ref = one / amrex::max(turbChoice.theta_ref, std::numeric_limits<Real>::min());
const Real l_inv_theta0  = (use_ref_theta) ? inv_theta_ref : one;
```

- **No zero division, in either precision.** The divisor is never zero, and `1/numeric_limits<Real>::min()` is finite in both `float` and `double`. There is no absolute constant to re-validate for a different precision.
- **Values unchanged.** The selected value is `one` when `theta_ref <= 0`. For every normal `theta_ref > 0` it is bit-identical to `1/theta_ref`, since the clamp is the identity there.
- **The guard stays in place.** `theta_ref > 0` does not imply `theta_ref >= min()` (denormals), so the optimiser cannot remove the `max` inside the selected arm.

Two forms that do not work:
- `one / (std::abs(theta_ref) + Real(1e-30))`, the first version of this PR, relied on an absolute epsilon.
- Selecting the divisor, `one / (use_ref_theta ? theta_ref : one)`, still traps. Clang folds `1/1` and moves the division back into the select arm, giving the same `fdiv`/`fcsel` as the original. An ERF build with that form trapped in three cases:
  - `mrf_unstable`, in `DiffusionSrcForState_S`
  - Deardorff with `theta_ref = 0`, in `ComputeTurbulentViscosity`
  - the k-eqn Dirichlet case, in `SurfaceLayer::update_fluxes`

## Verification

Release, AppleClang on arm64, MPI, `ERF_ENABLE_ALL_WARNINGS=ON`.

Reproduction with `amrex.fpe_trap_zero=1`. The unfixed binary was built from `development` 489a2aa64:

| Case | development | this branch |
|---|---|---|
| `ABL/mrf_unstable` | SIGILL at step 1 in `DiffusionSrcForState_S` | 300 steps clean |
| `mrf_unstable` + `pbl_type=None rans_type=kEqn dirichlet_k=true` | SIGILL at init in `SurfaceLayer::update_fluxes` | clean |

Results are unchanged in double and single precision. All runs use `amrex.fpe_trap_zero=1`, and fcompare uses `--abs_tol 0 --rel_tol 0` on all plotted fields. Each build is compared with the epsilon version of this branch at the same precision; that version was itself bit-identical to `development`.

| Case | Steps | `PRECISION=DOUBLE` | `PRECISION=SINGLE` |
|---|---|---|---|
| `ABL/mrf_unstable` (theta_ref unset) | 300 | clean, bit-identical | clean, bit-identical |
| `ABL/inputs_deardorff`, `n_cell = 32 32 32`, `theta_ref = 300` | 40 | clean, bit-identical | clean, bit-identical |
| `ABL/inputs_deardorff`, `n_cell = 32 32 32`, `theta_ref = 0` | 10 | clean, bit-identical | clean, bit-identical |
| k-eqn variant, `theta_ref = 300` | 2 | clean, bit-identical | clean, bit-identical |
| k-eqn variant, theta_ref unset | 2 | clean, bit-identical | clean, bit-identical |

`codespell` and `git diff --check` are clean, and the change adds no warnings in either precision.

## Notes

- The k-eqn variant is only a way to reach the RANS sites; it is not a supported deck. It went to NaN in step 5 with both the `development` binary and the first version of this branch, identically, so comparisons stop at step 2.
- The ERF-Hazard fork took the first version as hgopalan/ERF#396 and gets the clamp in a follow-up PR.

### DIFF
--- a/Source/BoundaryConditions/ERF_SurfaceLayer.cpp
+++ b/Source/BoundaryConditions/ERF_SurfaceLayer.cpp
@@ -253,8 +253,10 @@ SurfaceLayer::update_fluxes (const int& lev,
     }
 
     if (m_update_k_rans) {
+        // Nonzero divisor: both arms of the select are evaluated (fpe_trap_zero)
         const bool use_ref_theta = (theta_ref > 0);
-        const Real l_inv_theta0  = (use_ref_theta) ? one / theta_ref : one;
+        const Real inv_theta_ref = one / (std::abs(theta_ref) + Real(1e-30));
+        const Real l_inv_theta0  = (use_ref_theta) ? inv_theta_ref : one;
         const Real l_inv_Cmu2    = inv_Cmu2;
         const int klo = m_geom[lev].Domain().smallEnd(2);
         IntVect ng = u_star[lev]->nGrowVect(); ng[2] = 0;
--- a/Source/BoundaryConditions/ERF_SurfaceLayer.cpp
+++ b/Source/BoundaryConditions/ERF_SurfaceLayer.cpp
@@ -253,9 +253,10 @@ SurfaceLayer::update_fluxes (const int& lev,
     }
 
     if (m_update_k_rans) {
-        // Nonzero divisor: both arms of the select are evaluated (fpe_trap_zero)
+        // Clamped divisor: the select is if-converted, so 1/theta_ref runs even
+        //   when theta_ref = 0 and would trip fpe_trap_zero (see ERF_SetupDiff.H)
         const bool use_ref_theta = (theta_ref > 0);
-        const Real inv_theta_ref = one / (std::abs(theta_ref) + Real(1e-30));
+        const Real inv_theta_ref = one / amrex::max(theta_ref, std::numeric_limits<Real>::min());
         const Real l_inv_theta0  = (use_ref_theta) ? inv_theta_ref : one;
         const Real l_inv_Cmu2    = inv_Cmu2;
         const int klo = m_geom[lev].Domain().smallEnd(2);
--- a/Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp
+++ b/Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp
@@ -170,8 +170,10 @@ void ComputeTurbulentViscosityLES (Vector<std::unique_ptr<MultiFab>>& Tau_lev,
         const Real Ce_lcoeff    = amrex::max(zero, l_C_e - Real(1.9)*l_C_k);
         const Real l_abs_g      = const_grav;
 
+        // Nonzero divisor: both arms of the select are evaluated (fpe_trap_zero)
         const bool use_ref_theta = (turbChoice.theta_ref > 0);
-        const Real l_inv_theta0  = (use_ref_theta) ? one / turbChoice.theta_ref : one;
+        const Real inv_theta_ref = one / (std::abs(turbChoice.theta_ref) + Real(1e-30));
+        const Real l_inv_theta0  = (use_ref_theta) ? inv_theta_ref : one;
 
 #ifdef _OPENMP
 #pragma omp parallel if (Gpu::notInLaunchRegion())
@@ -610,8 +612,10 @@ void ComputeTurbulentViscosityRANS (Vector<std::unique_ptr<MultiFab>>& /*Tau_lev
         const Real l_g_max    = turbChoice.l_g_max;
         const Real abs_g      = const_grav;
 
+        // Nonzero divisor: both arms of the select are evaluated (fpe_trap_zero)
         const bool use_ref_theta = (turbChoice.theta_ref > 0);
-        const Real inv_theta0  = (use_ref_theta) ? one / turbChoice.theta_ref : one;
+        const Real inv_theta_ref = one / (std::abs(turbChoice.theta_ref) + Real(1e-30));
+        const Real inv_theta0  = (use_ref_theta) ? inv_theta_ref : one;
 
 #ifdef _OPENMP
 #pragma omp parallel if (Gpu::notInLaunchRegion())
--- a/Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp
+++ b/Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp
@@ -170,9 +170,10 @@ void ComputeTurbulentViscosityLES (Vector<std::unique_ptr<MultiFab>>& Tau_lev,
         const Real Ce_lcoeff    = amrex::max(zero, l_C_e - Real(1.9)*l_C_k);
         const Real l_abs_g      = const_grav;
 
-        // Nonzero divisor: both arms of the select are evaluated (fpe_trap_zero)
+        // Clamped divisor: the select is if-converted, so 1/theta_ref runs even
+        //   when theta_ref = 0 and would trip fpe_trap_zero (see ERF_SetupDiff.H)
         const bool use_ref_theta = (turbChoice.theta_ref > 0);
-        const Real inv_theta_ref = one / (std::abs(turbChoice.theta_ref) + Real(1e-30));
+        const Real inv_theta_ref = one / amrex::max(turbChoice.theta_ref, std::numeric_limits<Real>::min());
         const Real l_inv_theta0  = (use_ref_theta) ? inv_theta_ref : one;
 
 #ifdef _OPENMP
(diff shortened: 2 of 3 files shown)


asalmgren:
@hgopalan -- 1e-30 is not robust to single precision -- please modify to allow this to pass with both single and double precision

asalmgren (commented):
The diagnosis is right and well-evidenced: clang at `-O3` if-converts `cond ? one/theta_ref : one` into an unconditional `fdiv` plus `fcsel`, so the discarded `1/0` still raises `FE_DIVBYZERO` and trips `amrex.fpe_trap_zero`. All four sites that use this pattern are covered (`ERF_SetupDiff.H` reaches the `_S`, `_N`, `_T` and `_EB` kernels), and nothing else under `Source/` divides by `theta_ref`.

The problem is the shape of the fix, not the intent.

## What's wrong with the current fix

```cpp
const Real inv_theta_ref = one / (std::abs(theta_ref) + Real(1e-30));
const Real l_inv_theta0  = (use_ref_theta) ? inv_theta_ref : one;
```

This still performs the division on the `theta_ref <= 0` path and relies on an *absolute* magic epsilon to keep it finite. That constant's validity is precision-dependent and has to be re-argued for every `PRECISION` setting — the PR description already has to make that argument ("`1e-30` rather than `1e-300` keeps the guard from underflowing in single precision"). `1e-300` would flush to exactly `0.0f` under `AMREX_USE_FLOAT` and silently reintroduce the trap. `1e-30` happens to survive in both, but the code carries no protection against the next person tightening it.

Two lesser issues: `std::abs()` is dead — it only changes the value for `theta_ref < 0`, where the select discards it anyway — and adding it to `ERF_SetupDiff.H` puts a `<cmath>` dependency into a header that is textually included in five translation units that do not include `<cmath>` directly.

## The correct fix

Select the denominator, not the quotient:

```cpp
const bool use_ref_theta = (turbChoice.theta_ref > 0);
const Real l_inv_theta0  = one / (use_ref_theta ? turbChoice.theta_ref : one);
```

There is now exactly one division, and its divisor is the output of the select — either `theta_ref` (positive by the guard) or `1`. If-conversion is then harmless: `fcsel` followed by `fdiv` can never see a zero divisor, because no code path divides by `theta_ref` when `theta_ref <= 0`. GCC at `-O3` emits a single guarded `divsd` for this form, identical to the original.

Apply it at all four sites, adjusting the variable name at `ERF_ComputeTurbulentViscosity.cpp:614` (`inv_theta0`) and the member name at `ERF_SurfaceLayer.cpp:257` (`theta_ref`).

## Why this runs correctly in both single and double precision

- No absolute constant, so there is nothing to re-validate against the float exponent range — the fix is correct by construction for `double`, `float`, or anything else.
- Bit-identical to `development` for every `theta_ref`, in both precisions. Comparing the two forms side by side, they agree for all physical values, but the epsilon form already diverges from the true reciprocal at `theta_ref = 1e-35` (`9.9999e+29` vs `1e+35`) in float and in double, and at `1e-20` in double. Those are unphysical inputs, but it means the stated invariant "the selected value does not change" holds only over a range someone has to keep in their head; the denominator form has no such range.
- Drops `std::abs`, so there is no `<cmath>` question and no host/device `amrex::Math::abs` consideration.

The comment can then say what it is actually guarding, e.g.

```cpp
// Divide by the selected value, not the raw theta_ref: an optimized build
//   if-converts the select, and 1/theta_ref with the default theta_ref = 0
//   trips fpe_trap_zero
```

Everything else in the PR — the reproduction, the `amrex_fcompare` verification, and the note that the k-eqn deck is only a vehicle for reaching the RANS sites — holds as written, and the bit-identity results carry over unchanged to this form.

hgopalan:
@asalmgren thanks for the review. I pushed 2af4a37ab, which removes the absolute epsilon. It does not use the denominator-select form, because that form still traps with clang. Point by point:

## 1. Single precision

The guard no longer has an absolute constant. The divisor is clamped to the smallest normal `Real` before the select:

```cpp
const bool use_ref_theta = (turbChoice.theta_ref > 0);
const Real inv_theta_ref = one / amrex::max(turbChoice.theta_ref, std::numeric_limits<Real>::min());
const Real l_inv_theta0  = (use_ref_theta) ? inv_theta_ref : one;
```

The divisor is never zero, and `1/numeric_limits<Real>::min()` is finite for both `float` (8.5e37) and `double` (4.5e307). So nothing needs re-checking for a different precision. I also built `ERF_PRECISION=SINGLE` and ran every case; results are in the table below.

## 2. Selecting the denominator traps with clang

```cpp
const Real l_inv_theta0 = one / (use_ref_theta ? turbChoice.theta_ref : one);
```

AppleClang 21 at `-O3` folds `1/1` to a constant and moves the division back into the select arm. In a standalone `-O3` test, both `float` and `double` raised FE_DIVBYZERO at `theta_ref = 0`. The `double` code is the same sequence as the original `use_ref_theta ? one/theta_ref : one`:

```
fmov  d1, #1.0
fdiv  d2, d1, d0      <- 1/theta_ref, unconditional
fcmp  d0, #0.0
fcsel d1, d2, d1, gt
```

You're right about GCC. In the same standalone test, GCC 14 kept a real branch for this form, but it did the same for the original code. GCC never showed the bug, so it can't tell the two forms apart. I built ERF with the suggested form at all four sites (Release, AppleClang arm64, `amrex.fpe_trap_zero=1`):

| Case | With `one / (use_ref_theta ? theta_ref : one)` |
|---|---|
| `ABL/mrf_unstable` (theta_ref unset) | SIGILL at step 1 in `DiffusionSrcForState_S` |
| `ABL/inputs_deardorff`, `theta_ref = 0` | SIGILL at step 1 in `ComputeTurbulentViscosity` |
| `mrf_unstable` + `pbl_type=None rans_type=kEqn dirichlet_k=true` | SIGILL at init in `SurfaceLayer::update_fluxes` |

The clamp does not have this problem. `theta_ref > 0` does not imply `theta_ref >= numeric_limits<Real>::min()` (denormals), so the compiler cannot drop the `max` inside the arm, and no division by zero reaches the hardware.

## 3. `std::abs` and the `<cmath>` dependency in `ERF_SetupDiff.H`

`std::abs` is gone. The header now uses only `amrex::max` and `std::numeric_limits`, and `AMReX_Algorithm.H` already includes `<limits>`. All four including kernels (`_S`, `_N`, `_T`, `_EB`) compile with `ERF_ENABLE_ALL_WARNINGS=ON` in both precisions, with no new warnings.

## 4. Values unchanged

The clamp is the identity for every normal `theta_ref > 0`: at least 2.2e-308 in double and 1.2e-38 in float. Only a denormal `theta_ref` would differ. The epsilon form, by contrast, departed from `1/theta_ref` at 1e-20 in double.

## 5. Names and comment

- The existing names are kept: `inv_theta0` at the RANS site in `ERF_ComputeTurbulentViscosity.cpp` and the member `theta_ref` in `ERF_SurfaceLayer.cpp`.
- The comment in `ERF_SetupDiff.H` follows your suggestion. It also notes that selecting the divisor is if-converted too, and that the clamp is exact and finite in both precisions.
- The three `.cpp` sites carry a two-line comment that points to the header.

## Verification

Release, AppleClang arm64, MPI, `ERF_ENABLE_ALL_WARNINGS=ON`, branch merged with current `development`. All runs use `amrex.fpe_trap_zero=1`. "Bit-identical" means `amrex_fcompare --abs_tol 0 --rel_tol 0` on all plotted fields against the previous commit on this branch, which was itself bit-identical to `development`.

| Case | Steps | `PRECISION=DOUBLE` | `PRECISION=SINGLE` |
|---|---|---|---|
| `ABL/mrf_unstable` (theta_ref unset) | 300 | clean, bit-identical | clean, bit-identical |
| `ABL/inputs_deardorff`, `n_cell = 32 32 32`, `theta_ref = 300` | 40 | clean, bit-identical | clean, bit-identical |
| `ABL/inputs_deardorff`, `n_cell = 32 32 32`, `theta_ref = 0` | 10 | clean, bit-identical | clean, bit-identical |
| k-eqn variant, `theta_ref = 300` | 2 | clean, bit-identical | clean, bit-identical |
| k-eqn variant, theta_ref unset | 2 | clean, bit-identical | clean, bit-identical |

`codespell` and `git diff --check` are clean. I've updated the PR description to match.

PR #3974, opened 2026-09-10
[EB] Construct implicit function based on STL

skang67:
This PR enables the construction of an implicit function (signed-distance function) from an input STL file. 

## Motivation
Currently, terrain is represented using a vertical (not the full 3D) signed-distance function derived from height data, which does not accurately represent vertical surfaces, such as building walls. Using STL geometry enables a more accurate representation. 

## Additional Details
- When both terrain and buildings are modeled, an implicit function is constructed for each geometry and combined using `EB2::makeUnion`.
- The `mask_fc_from_cc` mask is applied to FC EB datasets to ensure consistency between the CC and FC EB datasets. Once AMReX is updated to construct all CC and FC EB datasets from a single `IndexSpace`, this additional masking step can be removed.


asalmgren:
@skang67 -- I'm wondering if you could take the EB-related code out of ERF_Constructors and into a new file that would live in Source/EB ... then simply call that routine from ERF_Constructors ... it would keep things cleaner ... thanks!

skang67:
> @skang67 -- I'm wondering if you could take the EB-related code out of ERF_Constructors and into a new file that would live in Source/EB ... then simply call that routine from ERF_Constructors ... it would keep things cleaner ... thanks!

I will refactor the constructor for this.

skang67:
@adamwise95 In this PR, I encapsulated all the EB and ImmersedForcing related initialization in a separate function `InitializeEB` within the ERF constructor.

PR #3975, opened 2026-09-10
Fix development CI warnings and Windows SHOC startup tests

pressel:
## Summary

Fix the CI failures present on `development`:

- Restore the safe unstable-branch MRF expression using `HOL_abs`, eliminating the GCC/SYCL/HIP unused-variable warning.
- Resolve multi-configuration Windows executable paths in the native SHOC removed-transport startup tests.

## Testing

Using the provided OpenMPI environment and a parallel build with `--parallel 8`:

- Unit tests: 552/552 passed
- Parallel tests: 48/48 passed
- Regression tests: 74/74 passed
- Removed transport-mode tests: 2/2 passed

No existing branches were modified.

### DIFF
--- a/Source/PBL/ERF_ComputeDiffusivityMRF.cpp
+++ b/Source/PBL/ERF_ComputeDiffusivityMRF.cpp
@@ -601,7 +601,7 @@ pblh_mf.setVal(0.0);
                             ? (enable_qnse_d > Real(0.5)
                                ? (1 + qnse_am_d * HOL_bounded) / (1 + qnse_bm_d * HOL_bounded)
                                : (1 + 5 * HOL_bounded))
-                            : std::pow(amrex::max(1 - 16 * HOL_bounded, Real(0.01)), -one_quarter);
+                            : std::pow(1 + 16 * HOL_abs, -one_quarter);
             const Real phiM_safe = amrex::max(phiM, Real(0.01));
 
             // Absolute bounds [0.01, 5.0] m/s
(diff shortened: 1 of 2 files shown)


Source/PBL/ERF_ComputeDiffusivityMRF.cpp
```diff
@@ -601,7 +601,7 @@ pblh_mf.setVal(0.0);
                             ? (enable_qnse_d > Real(0.5)
                                ? (1 + qnse_am_d * HOL_bounded) / (1 + qnse_bm_d * HOL_bounded)
                                : (1 + 5 * HOL_bounded))
-                            : std::pow(amrex::max(1 - 16 * HOL_bounded, Real(0.01)), -one_quarter);
+                            : std::pow(1 + 16 * HOL_abs, -one_quarter);
```
AMLattanzi, line 604: Should the other computations in this file also use `1 + 16 * HOL_abs`? The comment to use `|HOL|` appears other places but the computation still seems to use `1 - 16 HOL_bounded`?

pressel: Yes — I think you're right. I traced this back to #3961 (5c5ea7b), which intentionally changed all of these unstable MRF phi expressions to 1 + 16*abs(HOL). The reason wasn't just algebraic cleanup: in optimized builds the compiler could hoist the fractional pow() out of the selected branch and evaluate an invalid base, tripping amrex.fpe_trap_invalid.

The later PBLH/QNSE changes appear to have reintroduced the older guarded 1 - 16*HOL_bounded form in Pass 4 and in the K-profile phiM/phit expressions. The unused-variable warning exposed the Pass-4 occurrence, but the two K-profile expressions are still in the older form.

I'll restore the safe form there as well:

phiM: pow(1 + 16*HOL_abs, -1/4)

phit: pow(1 + 16*HOL_abs, -1/2)

For the selected unstable branch HOL <= 0, these are algebraically identical to 1 - 16*HOL, so this preserves the intended calculation while restoring the optimizer/FPE-safety fix from #3961. The cloud-unstable expressions already use HOL_abs.

I'll make this consistent.

PR #3976, opened 2026-09-10
BTF fine level: make the terrain-fitted mesh independent of the z split

hgopalan:
## Summary

With `erf.terrain_smoothing = 0` (BTF), the terrain-fitted mesh on a refined level depended on how that level's BoxArray is split in z.

- **The defect.** The BTF branch of `init_which_terrain_grid` maps every column of a box between the box's lowest node and the domain top, taking the height at that lowest node as the surface. On a fine level, a box that sits on another box of the level holds only the mesh interpolated from the coarse level at its lowest node. The mesh above a z split therefore followed the coarse terrain rather than the fine terrain. Level 0 never showed it, because `make_terrain_fitted_coords` already builds level 0 on full-height `ChopGrids2D` columns when a box does not reach the bottom.
- **The fix.** `make_terrain_fitted_coords` now builds a fine BTF level on boxes that hold whole columns, and copies the result back:
  - boxes stacked in z are joined (boxes stacked with different lateral extents are cut at each other's lateral edges first);
  - the mesh is copied in with its ghost nodes and built with the unchanged BTF branch;
  - it is copied back, with valid nodes taking precedence over ghost nodes.
- **Unchanged cases.** A layout with no boxes stacked in z takes the old call and is bitwise unchanged. The bottom box of a column, such as a refined patch aloft, still starts from its own lowest node. STF and Sullivan TF fine levels (`terrain_smoothing` 1 or 2) are not affected.

## Numbers

32 x 32 x 40 radial Witch of Agnesi hill (h = 100 m, L = 100 m, 640 x 640 x 400 m, periodic in x and y, Isentropic init), level 1 with `amr.ref_ratio_vect = 2 2 1` over the hill top (160..480 m), step 0. Split: `amr.max_grid_size_z = 1024 16` (fine grids split at k = 16 and 32). Whole columns: `amr.max_grid_size_z = 1024 1024`.

| comparison | `development` | this PR |
|---|---|---|
| split vs whole columns, level-1 `z_phys` | 0.54 m at k = 16, decaying to 0.01 m at k = 39 | bitwise |
| split vs whole columns, `detJ` (both levels) | 2.3e-3 | bitwise |
| whole columns, `development` vs this PR (all plotted fields) | | bitwise |
| unsplit patch aloft (fine grids k = 10..39), `development` vs this PR | | bitwise |
| split vs unsplit patch aloft, `z_phys` | 1.7e-13 | bitwise |

Together with #3970 (this PR's `ERF_TerrainMetrics.cpp` on top of its head 28f4f4813), the whole plotfile is bitwise identical between split and whole-column fine grids on both levels at step 0: `z_phys`, `detJ`, `dens_hse`, `pres_hse`, `density`, `theta` and `pressure`. The whole-column run is bitwise the same with and without #3970.

## Tests

- **`TerrainFineColumns` (gtest, serial and parallel executables).** Builds a fine-level mesh from a synthetic interpolated mesh and fine terrain, and compares every node, valid and ghost, bit for bit against whole columns. Cases: split in z, uneven split, a stack whose lower boxes are cut in x and upper boxes in y, a split patch aloft, each with and without periodic x and y. A fifth test shows that the per-box build does depend on the split.
- **`Terrain2Lev_BTF_ZSplit` (regression).** New `add_test_terrain_zsplit_parity` / `Tests/RunTerrainZSplitParity.cmake`. It runs the deck above with whole-column and split fine grids and compares `z_phys` and `detJ` with `fcompare -a` at zero tolerance. It takes about 1.5 s on 2 ranks and needs no gold file. `development` fails it.

All 455 serial and 21 parallel unit tests pass.

## Notes

- Independent of #3970: different files, either merge order. With this PR alone the fine-level `dens_hse` and `pres_hse` still depend on the split at step 0 (by 1.6e-4 and 19 Pa on the deck above), because the fine-level hydrostatic initialisation integrates box by box; item 5 of #3970 fixes that. The CTest therefore compares only the mesh.
- Seen in passing, not changed: the comment in the BTF branch saying the top node of a box is not overwritten is stale (the loop box is grown by `nGrow - 1` at the top), and the BTF loop reads the surface node in the same `ParallelFor` that rewrites it.

### DIFF
--- a/Source/Utils/ERF_TerrainMetrics.cpp
+++ b/Source/Utils/ERF_TerrainMetrics.cpp
@@ -3,7 +3,13 @@
 #include <AMReX_ParmParse.H>
 #include <ERF_Constants.H>
 #include <ERF_Interpolation_1D.H>
+#include <algorithm>
+#include <array>
 #include <cmath>
+#include <map>
+#include <numeric>
+#include <utility>
+#include <vector>
 
 using namespace amrex;
 
@@ -41,6 +47,119 @@ init_default_zphys (int /*lev*/,
         });
     }
 }
+
+namespace {
+
+/**
+ * The cells of a (cell-centered) BoxArray as boxes that are never stacked on each other.
+ *
+ * Wherever a box of ba sits on the top face of another box of ba, the run of cells in z
+ * that the two belong to becomes a single box.  Boxes stacked with different lateral
+ * extents are first cut at each other's lateral edges; the joined pieces are then merged
+ * in x and y and chopped to no more than the largest lateral size of the boxes they came
+ * from.  Returns ba itself when no two boxes of ba are stacked.
+ */
+BoxArray
+join_boxes_stacked_in_z (const BoxArray& ba)
+{
+    AMREX_ALWAYS_ASSERT(ba.ixType().cellCentered());
+
+    const int nboxes = static_cast<int>(ba.size());
+
+    // Gather the boxes into stacks: link every box to the boxes that sit on its top face
+    std::vector<int> stack_of(nboxes);
+    std::iota(stack_of.begin(), stack_of.end(), 0);
+    auto find_stack = [&stack_of] (int i) {
+        while (stack_of[i] != i) {
+            stack_of[i] = stack_of[stack_of[i]];
+            i = stack_of[i];
+        }
+        return i;
+    };
+
+    bool any_stacked = false;
+    for (int i = 0; i < nboxes; ++i) {
+        Box above(ba[i]);
+        above.setRange(2, ba[i].bigEnd(2)+1);
+        for (const auto& isect : ba.intersections(above)) {
+            stack_of[find_stack(isect.first)] = find_stack(i);
+            any_stacked = true;
+        }
+    }
+    if (!any_stacked) { return ba; }
+
+    // std::map keeps the order, and so the joined boxes, the same on every rank
+    std::map<int, std::vector<Box>> stacks;
+    for (int i = 0; i < nboxes; ++i) {
+        stacks[find_stack(i)].push_back(ba[i]);
+    }
+
+    BoxList bl_joined;
+    for (const auto& stack : stacks)
+    {
+        const std::vector<Box>& boxes = stack.second;
+        if (boxes.size() == 1) {
+            bl_joined.push_back(boxes[0]);
+            continue;
+        }
+
+        // Cut the stack at every lateral edge of its boxes
+        std::array<std::vector<int>,2> cuts;
+        IntVect max_size(1);
+        for (const auto& b : boxes) {
+            for (int idim = 0; idim < 2; ++idim) {
+                cuts[idim].push_back(b.smallEnd(idim));
+                cuts[idim].push_back(b.bigEnd(idim)+1);
+                max_size[idim] = std::max(max_size[idim], b.length(idim));
+            }
+        }
+        for (auto& c : cuts) {
+            std::sort(c.begin(), c.end());
+            c.erase(std::unique(c.begin(), c.end()), c.end());
+        }
+        auto cut_index = [&cuts] (int idim, int coord) {
+            return static_cast<int>(std::lower_bound(cuts[idim].begin(), cuts[idim].end(), coord)
+                                    - cuts[idim].begin());
+        };
+
+        // The z extents of the boxes over each piece, keyed by the piece's cut indices
+        std::map<std::pair<int,int>, std::vector<std::pair<int,int>>> zruns_of_piece;
+        for (const auto& b : boxes) {
+            for (int ix = cut_index(0, b.smallEnd(0)); ix < cut_index(0, b.bigEnd(0)+1); ++ix) {
+                for (int iy = cut_index(1, b.smallEnd(1)); iy < cut_index(1, b.bigEnd(1)+1); ++iy) {
+                    zruns_of_piece[std::make_pair(ix,iy)].emplace_back(b.smallEnd(2), b.bigEnd(2));
+                }
+            }
+        }
+
+        // Over each piece, join the extents that touch into one box per run of cells
+        BoxList bl_stack;
+        for (auto& piece : zruns_of_piece) {
+            const int ix = piece.first.first;
+            const int iy = piece.first.second;
+            auto& zruns = piece.second;
+            std::sort(zruns.begin(), zruns.end());
+            std::size_t n_first = 0;
+            for (std::size_t n = 1; n <= zruns.size(); ++n) {
+                if (n == zruns.size() || zruns[n].first != zruns[n-1].second+1) {
+                    bl_stack.push_back(Box(IntVect(cuts[0][ix],     cuts[1][iy],     zruns[n_first].first),
+                                           IntVect(cuts[0][ix+1]-1, cuts[1][iy+1]-1, zruns[n-1].second)));
+                    n_first = n;
+                }
+            }
+        }
+
+        // Merging in x and y joins pieces with the same z extent, so it cannot stack them again
+        bl_stack.simplify();
+        max_size[2] = bl_stack.minimalBox().length(2);
+        bl_stack.maxSize(max_size);
+        bl_joined.join(bl_stack);
+    }
+
+    return BoxArray(std::move(bl_joined));
+}
+
+} // namespace
 
 /**
  * Computation of the terrain grid from BTF, STF, or Sullivan TF model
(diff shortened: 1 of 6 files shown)


Tests/CTestList.cmake
```diff
@@ -686,6 +711,7 @@ add_test_r(MovingTerrain_nosub               ""  "erf_exec" "plt00020" RUNTIME_O
 add_test_r(MovingTerrain_sub                 ""  "erf_exec" "plt00010" RUNTIME_OPTIONS "erf.vert_implicit=false ")
 add_test_r(Terrain2Lev_STF_interp            ""  "erf_exec" "plt00010" RUNTIME_OPTIONS "erf.vert_implicit=false ")
 add_test_r(Terrain2Lev_STF_transform         ""  "erf_exec" "plt00010" RUNTIME_OPTIONS "erf.vert_implicit=false ")
+add_test_terrain_zsplit_parity(Terrain2Lev_BTF_ZSplit "plt00000")
```
asalmgren: **High.** This registration sits in the unconditional section of `CTestList.cmake`, but every other `execute_process`-driven test of this shape (cloud chamber, anelastic wall diffusion, SHOC mutation) is inside the `if(ERF_ENABLE_MPI)` block at lines 480-506 / 697. Two things break in a non-MPI build:

1. `find_package(MPI)` runs only under `if(ERF_ENABLE_MPI)` (`CMakeLists.txt:448`), so with `-DERF_ENABLE_MPI:BOOL=OFF` both `MPIEXEC_EXECUTABLE` and `MPIEXEC_NUMPROC_FLAG` are undefined. Passing `-DMPIEXEC=` still *defines* `MPIEXEC` (as the empty string), so the guard at `RunTerrainZSplitParity.cmake:3` passes, and `execute_process(COMMAND ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ${NRANKS} ...)` collapses to `COMMAND 1 <erf_exec> ...` -- it tries to exec a program literally named `1`.
2. On WIN32, `resolve_test_exe` returns a glob (`${CMAKE_BINARY_DIR}/Exec/*/erf_exec.exe`, line 11) and `FCOMPARE_EXE` is likewise a glob (`CMakeLists.txt:238`). Those work only because `add_test_r` goes through `sh -c`; `execute_process` does no globbing, so the paths never resolve.

Concrete failure: `.github/workflows/windows.yml` configures with `ERF_ENABLE_MPI:BOOL=OFF` and then runs `ctest -C Debug -L regression -VV` (line 81). This test carries `LABELS "regression"`, so it will be selected there and fail.

Suggest `if(ERF_ENABLE_MPI AND NOT WIN32)` around the registration, matching the existing precedent.

Source/Utils/ERF_TerrainMetrics.cpp
```diff
@@ -93,7 +212,40 @@ make_terrain_fitted_coords (int lev,
...
+        //
+        // BTF maps each column of a box between the box's lowest node and the domain top.
+        // A box that sits on another box of this level must continue the columns of the box
+        // below it: its own lowest node holds only the mesh interpolated from the coarse
+        // level, so starting from there would make the mesh above it follow the coarse
+        // terrain and depend on where the grids are split in z.  As on level 0 we therefore
+        // build the mesh on boxes that hold whole columns and copy it back.  The bottom box
+        // of a column, such as a refined patch aloft, still starts from its own lowest node.
+        //
+        ParmParse pp("erf");
```
asalmgren, line 286: **Low/medium (performance).** The column `BoxArray` and its `DistributionMapping` are recomputed from scratch on every call, and `make_terrain_fitted_coords` is not called once per run at lev > 0: `Source/TimeIntegration/ERF_TI_utils.H` calls it three times per RK-stage geometry update (lines 160, 166, 173) and twice per acoustic substep (lines 227, 230).

So for a `MovingFittedMesh` run with a z-split fine level, each of those five calls per substep now re-runs the union-find over all boxes, `BoxList::simplify()` (quadratic in the piece count), builds a fresh `DistributionMapping`, allocates a full extra nodal `MultiFab`, and does three `ParallelCopy`s. Because `ba_col`/`dm_col` are new objects each time, their `BDKey` differs from the previous call's, so every one of those `ParallelCopy`s misses the `FabArrayBase` CPC cache and rebuilds its communication metadata from scratch.

The joined BoxArray only changes on regrid, so caching it (and its DM) alongside the level's grids would remove the whole per-substep cost.

Source/Utils/ERF_TerrainMetrics.cpp
```diff
@@ -93,7 +212,40 @@ make_terrain_fitted_coords (int lev,
...
+        int terrain_smoothing = 0;
+        pp.query("terrain_smoothing", terrain_smoothing);
+
+        const BoxArray ba_cc = amrex::convert(z_phys_nd.boxArray(), IntVect::TheCellVector());
+        BoxArray ba_col = (terrain_smoothing == 0) ? join_boxes_stacked_in_z(ba_cc) : ba_cc;
+
+        if (ba_col == ba_cc) {
+            init_which_terrain_grid(lev, geom, z_phys_nd, z_levels_h, fine_terrain, z_phys_interp);
+        } else {
+            DistributionMapping dm_col(ba_col);
```
asalmgren: **Medium.** This builds a `DistributionMapping` over `ba_col` with no lateral re-chop, so the joined level can end up with far fewer boxes than there are ranks.

Joining divides the box count by the number of z-splits, and `maxSize(max_size)` cannot increase it again -- `max_size[0..1]` is the *largest* lateral length seen among the stack's boxes, so nothing gets chopped laterally. AMReX's default `max_grid_size` is 32 in z on CPU, so any fine level taller than 32 cells already takes this path by default.

Concrete scenario: a 512x512x256 fine level with `amr.max_grid_size = 512 512 32` on 64 ranks. The 8 boxes stacked per lateral tile collapse to 1; with a single lateral tile, the entire fine-level nodal `z_phys_nd` plus `ng` ghosts is allocated and computed on one rank while the other 63 idle.

The lev == 0 branch guards against exactly this: at line 196 it calls `ChopGrids2D(ba_new, domain, ParallelDescriptor::NProcs())` before building its `DistributionMapping`. The new lev > 0 branch has no equivalent. Chopping `bl_stack` laterally down to at least `NProcs()` boxes (rather than `maxSize(max_size)` alone) would restore the guarantee.

Tests/RunTerrainZSplitParity.cmake
```diff
@@ -0,0 +1,41 @@
...
+   NOT DEFINED WORKING_DIRECTORY OR NOT DEFINED FCOMPARE OR NOT DEFINED PLTFILE)
+    message(FATAL_ERROR "RunTerrainZSplitParity.cmake missing required argument")
+endif()
+
+set(FULL_DIR "${WORKING_DIRECTORY}/full_columns")
+set(SPLIT_DIR "${WORKING_DIRECTORY}/split_in_z")
+file(REMOVE_RECURSE "${FULL_DIR}" "${SPLIT_DIR}")
+file(MAKE_DIRECTORY "${FULL_DIR}" "${SPLIT_DIR}")
+
+foreach(RUN IN ITEMS "full_columns|1024 1024" "split_in_z|1024 16")
```
asalmgren, line 14: **Low (test robustness).** Nothing in this script checks that the two runs actually produced *different* level-1 BoxArrays.

If the `amr.max_grid_size_z=1024 16` override ever stops taking effect -- a raised `amr.blocking_factor`, a change to the deck's fine-level extent, or a quoting change that makes ParmParse see only one value -- then both runs generate identical grids, `fcompare` takes its `grids_match == true` path, and it compares two byte-identical plotfiles. The test passes while protecting nothing, and the regression it guards silently goes uncovered.

Since the entire value of this test is that the two grid layouts differ, that should be asserted: e.g. fail if the two `plt00000/Level_1` Headers describe the same BoxArray, or grep the two `simulation.log`s for differing level-1 grid counts.

asalmgren (commented):
Automated correctness review of the diff.

I checked the core of `join_boxes_stacked_in_z` fairly carefully and it looks sound: the joined BoxArray covers exactly the same cells as the input and stays disjoint; per-piece z-runs are maximal, so `BoxList::simplify()` can only merge laterally and cannot re-create a stack; `max_size[2] = minimalBox().length(2)` guarantees `maxSize` never chops in z; and `grow(C_nd, ng)` is contained in the union of `grow(Bi_nd, ng)`, so the `ParallelCopy` into `z_phys_nd_col` leaves no uninitialized node. The union-find (path halving, and the C++17 sequencing of `stack_of[find(a)] = find(b)`) is correct, and `std::map` ordering keeps the result rank-independent. I also confirmed that `-a` really is `--allow_diff_grids` in the vendored AMReX `fcompare` (not `--abs_tol`), that AMReX concatenates argv before parsing so `"amr.max_grid_size_z=1024 16"` does parse as two values, and that no existing gold-file test uses multilevel BTF, so no baselines are invalidated.

Four findings below: one CI break, one scaling concern, one performance concern, one test-robustness gap. No correctness bug found in the mesh construction itself.

Tests/CTestList.cmake
hgopalan: Agreed, fixed in 8d2105fa5. The registration moved into the `if(ERF_ENABLE_MPI)` block next to the other `execute_process` tests, wrapped in `if(NOT WIN32)`, so neither the non-MPI Windows job nor a Windows MPI build selects it.

Source/Utils/ERF_TerrainMetrics.cpp
hgopalan: Fixed in 8d2105fa5. `column_grids()` keeps the column BoxArray and its DistributionMapping per level, keyed on the level's domain and its cell-centred grids, and rebuilds them only when those change. The union-find, `simplify` and `DistributionMapping` now run once per regrid.

`z_phys_nd`, `z_phys_nd_new` and `z_phys_nd_src` share the level's BoxArray reference, so the key comparison is a pointer check on every call. Since the column BoxArray and DistributionMapping are now the same objects on every call, the three `ParallelCopy` calls hit the `FabArrayBase` CPC cache after the first one.

I left the temporary nodal MultiFab allocated per call rather than cached. Keeping it would hold a second copy of the fine `z_phys_nd` for the whole run, and the allocation comes from the arena pool. The cache is cleared through `ExecOnFinalize`.

Source/Utils/ERF_TerrainMetrics.cpp
hgopalan: Agreed, fixed in 8d2105fa5. After joining, the columns are chopped laterally with `ChopGrids2D(ba_col, domain, ParallelDescriptor::NProcs())`, the same guard the level-0 branch uses, before the DistributionMapping is built. `ChopGrids2D` only halves the x and y chunk sizes, so it never splits a column in z again.

Checked that the chop does not change the mesh:
- The gtests pass on 2, 3 and 4 ranks. On those rank counts the misaligned-stack and split-patch-aloft cases join to a single box, which then gets chopped.
- The PR deck on 8 ranks joins its 4 columns and chops them to 8. Both the split and the whole-column layout give `z_phys` and `detJ` bitwise identical to the 1-rank run.

Tests/RunTerrainZSplitParity.cmake
hgopalan: Good point, fixed in 8d2105fa5. After both runs the script reads the level-1 BoxArray size from each `plt00000/Level_1/Cell_H`. It fails unless the split run has more level-1 boxes than the whole-column run (12 against 4 on 2 ranks).

Checked it by running the script through a wrapper that drops the `amr.max_grid_size_z` override. Both runs then make 4 boxes, and the test stops with "The split run has 4 level-1 boxes and the whole-column run 4: the fine grids were not split in z" before it reaches fcompare.

PR #3977, opened 2026-09-11
[ASPIRE] Adding weather diagnostics

nataraj2:
This PR adds the capability to export line plots for variables of interest at a given station (ie. a given (latitude, longitude)) for comparison with [ARM data](https://adc.arm.gov/armuserreg/new). 

Images show the instantaneous rain accumulation contours for the Joplin Tornado case and the rain accumulation over time at a given location (lat, lon).

The inputs options (for eg.) are 
```
erf.io_station_tracker = true
erf.station_latitude = 37.347 # deg N
erf.station_longitude = -94.857 # deg W
erf.station_plot_freq = 100
```

Screen Shot 2026-09-11 at 4 44 00 PM


Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1547,6 +1547,23 @@ struct SolverChoice {
             }
         }
 
+        pp.queryAdd("io_station_tracker", io_station_tracker);
+        pp.queryAdd("station_latitude", station_latitude);
+        pp.queryAdd("station_longitude", station_longitude);
+        pp.queryAdd("station_plot_freq", station_plot_freq);
+
+        if(io_station_tracker)
+            if(station_latitude != -1.0 || station_longitude == -1.0){
```
asalmgren, line 1556: **blocking** — this condition is inverted on the latitude, so it aborts whenever the latitude is *valid*.

With the example from the PR description (`erf.station_latitude = 37.347`), `37.347 != -1.0` is true and the run stops here at start-up. The only way past the guard is to leave `station_latitude` at its sentinel, which is the opposite of what it is checking for. As it stands the feature cannot be switched on.

The hurricane equivalent a few lines up (`:1542`) has the form this wants:

```cpp
if(station_latitude == -1.0 || station_longitude == -1.0){
```

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1547,6 +1547,23 @@ struct SolverChoice {
...
+                amrex::Abort("You have turned on IO for station tracker by using erf.station_tracker=true."
```
asalmgren, line 1557: **low** — the message names `erf.station_tracker=true`, but the input is `erf.io_station_tracker` (queried at `:1550`). Someone who greps their inputs file for the name in the abort will not find it. Same wording in the `station_plot_freq` message at `:1562`.

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -2302,6 +2319,10 @@ struct SolverChoice {
     bool io_hurricane_eye_tracker = false;
     amrex::Real hurricane_eye_latitude = -1e10, hurricane_eye_longitude = -1e10;
 
+    bool io_station_tracker = false;
+    amrex::Real station_latitude= -1.0, station_longitude = -1.0;
```
asalmgren, line 2323: **low** — `-1.0` is a legal latitude and a legal longitude, so using it as the "unset" sentinel makes a station at 1&deg;S or 1&deg;W impossible to request: it would be rejected as a missing input once the check above is fixed.

`hurricane_eye_latitude`/`longitude` three lines up use `-1e10` for exactly this reason; worth matching so the two trackers behave the same way.

Source/Utils/ERF_WeatherDiagnostics.cpp
```diff
@@ -0,0 +1,350 @@
...
+    //       minimum. The first pass reduces the distance only; the second
+    //       records the location of the cells that attain it, through one
+    //       atomic on a packed index so value and location stay consistent.
+    const Dim3 dlo = lbound(lev_geom.Domain());
+    const int nx = lev_geom.Domain().length(0);
+
+    Gpu::DeviceScalar<Long> d_idx_min(std::numeric_limits<Long>::max());
+    Long* d_idx_min_ptr = d_idx_min.dataPtr();
+
+    if(sc.init_type == InitType::WRFInput){
```
asalmgren, line 159: **blocking** — there is a branch for `InitType::WRFInput` here and one for `InitType::HindCast` at `:201`, and no fallback for anything else.

On a uniform, `input_sounding`, metgrid or ideal run neither pass executes, `d_val_min` keeps `1e10`, `d_idx_min` keeps `Long::max`, and the unpack below yields an `(i,j)` that names no cell. `TrackerAtStation_RainAccumulation` then finds no box containing it, so `h_rain_accum_global` stays at `-bogus_large_value` and the line-plot file fills with `-1e18` at every sample.

Failing silently into a plausible-looking output file is the worst of the options here. Either support those init types or abort up front naming the ones that are supported.

Source/Utils/ERF_WeatherDiagnostics.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+/**
+ * Wrapper to track the hurricane eye position over time.
+ *
+ * @param[in] sc Solver choices
+ */
+void
+ERF::WeatherDiagnosticsTracker (const SolverChoice& sc)
+{
+    static bool is_start = true;
```
asalmgren, line 329: **blocking** — `is_start` is never set to `false`, so the "locate the station once" this is reaching for does not happen: the two full passes over the level plus the MPI reduce and broadcast run again on every output step.

The trap is what happens if that is fixed in the obvious way. `station_loc` at `:335` is default-constructed, so it is `{0, 0}`; adding `is_start = false` at the end of the `if` would make every later call sample cell (0,0) instead of the station, silently and with no sentinel to catch it.

Better to hold the location in a member (initialised to something invalid) than in a local guarded by a static. Note also that the indices are tied to `finest_level`, so whatever caches them has to be invalidated on a regrid.

Source/Utils/ERF_WeatherDiagnostics.cpp
```diff
@@ -0,0 +1,350 @@
...
+void
+ERF::TrackerAtStation_RainAccumulation(const SolverChoice& sc,
+                                       const int levc,
+                                       const std::pair<int,int>& station_loc,
+                                       const Real& time)
+{
+    const MoistureComponentIndices& mi = sc.moisture_indices;
+    const int idx = mi.qmoist_index_for_var("rain_accum");
+
+    AMREX_ALWAYS_ASSERT(idx >= 0);
```
asalmgren, line 275: **medium** — `qmoist_index_for_var("rain_accum")` returns `absent` for `MoistureType::None`, `SatAdj`, `Kessler_NoRain`, `MoistNoCondensation` and `SAM_NoPrecip_NoIce`, so this bare assertion fires at the first `post_timestep` of any such run with nothing to tell the user why.

This is knowable when the inputs are parsed. An abort next to the other `io_station_tracker` checks, saying the station tracker needs a microphysics scheme that carries `rain_accum`, would cost the user a start-up instead of a run.

Source/Utils/ERF_WeatherDiagnostics.cpp
```diff
@@ -0,0 +1,350 @@
...
+        i = -1;
+        j = -1;
+    } else {
+        i = static_cast<int>(idx % static_cast<Long>(nx)) + dlo.x;
+        j = static_cast<int>(idx / static_cast<Long>(nx)) + dlo.y;
+    }
+}
+
+} // anonymous namespace
+
```
asalmgren, line 60: **low (reuse)** — `ComputeLocation` duplicates `ERF::ComputeGlobalMinLocation` (`Source/Utils/ERF_HurricaneDiagnostics.cpp:77`), and `ComputeStationLocationIJ` below, together with the `pack_ij`/`unpack_ij` helpers, is a near-verbatim copy of `ERF::HurricaneEyeTrackerInitial` &mdash; including the comment explaining why the arg-min cannot be done in one pass.

That comment is a good sign the logic is subtle enough that two copies will drift, and the inverted check in `ERF_DataStruct.H:1556` is already damage from this copy. One shared "index of the cell nearest (lat, lon)" helper, taking the lat/lon arrays as arguments, would serve both trackers.

Source/ERF.cpp
```diff
@@ -601,6 +601,16 @@ ERF::post_timestep (int nstep, double time, double dt_lev0)
         }
     }
 
+    if(solverChoice.io_station_tracker &&
+       (nstep == 0 || nstep%solverChoice.station_plot_freq == 0)) {
```
asalmgren, line 605: **medium** — `nstep % solverChoice.station_plot_freq` is evaluated for every step once the tracker is on, but the input validation only rejects `station_plot_freq == -1`.

`erf.station_plot_freq = 0` is a natural way for a user to write "every step" and gives integer division by zero, i.e. SIGFPE at step 1 (`nstep == 0` short-circuits only the first one). Validating `station_plot_freq > 0` where it is read closes it.

Docs/sphinx_doc/Inputs.rst
```diff
@@ -2467,6 +2467,27 @@ hurricane eye through the run.
...
+
++-----------------------------------+-----------------------------------------------------------------+--------------------+------------+
+| Parameter                         | Definition                                                      | Acceptable Values  | Default    |
++===================================+=================================================================+====================+============+
+| **erf.io_station_tracker**        | required if the weather tracker output needs to be written out  | Boolean            | false      |
++-----------------------------------+-----------------------------------------------------------------+--------------------+------------+
+| **erf.station_latitude**          | latitude of the weather station in deg N                        | Real               | must be set|
++-----------------------------------+-----------------------------------------------------------------+--------------------+------------+
+| **erf.station_longitude**         | longitude of the weather station in deg W                       | Real               | must be set|
```
asalmgren, line 2486: **medium** — the convention documented here disagrees with the code. This says "deg W", but `ComputeStationLocationIJ` differences `station_longitude` directly against `lon_m` and against component 1 of `forecast_state_interp[lev][4]`, both of which are degrees **East** &mdash; and the example in the PR description uses `-94.857`.

A user who follows the table and enters `94.857` gets a station on the other side of the globe with no warning. Suggest "longitude of the weather station in deg E".

The Default column for the three station parameters says "must be set"; giving the actual default (the sentinel) would match how the rest of the table reads.

asalmgren (commented):
Review of the ASPIRE weather-station diagnostics. I read the new code against the hurricane tracker it is derived from and checked each finding at the PR head rather than by inspection alone.

The feature does not work as shipped: the input guard in `ERF_DataStruct.H:1556` has an inverted comparison, so `erf.io_station_tracker = true` with any valid latitude &mdash; including the `37.347` in the PR description &mdash; aborts at start-up. Two more findings are in the same class: there is no fallback for init types other than `WRFInput` and `HindCast`, which writes `-1e18` into the output file instead of failing, and the `is_start` guard never fires, so the station is relocated every output step (and would sample cell (0,0) forever if it were fixed in the obvious way).

Most of these trace back to the same root: `ERF_WeatherDiagnostics.cpp` is largely a copy of `ERF_HurricaneDiagnostics.cpp`, and the copy picked up an inverted operator, stale Doxygen, an include guard in a `.cpp`, and a second copy of a two-pass arg-min whose own comment says it is easy to get wrong. Sharing that helper would be worth doing before this lands.

Eleven findings inline: three blocking, four worth fixing before merge, four cleanup.

*Reviewed with Claude Code.*

Source/IO/ERF_TrackerOutput.cpp
```diff
@@ -139,3 +139,21 @@ ERF::WriteVTKPolyline(const std::string& filename,
...
+    }
+}
+
+std::string
+ERF::MakeStationTrackerFilename(const std::string& basename,
+                                const int nstep)
+{
+    std::ostringstream filename;
+    filename << "StationTracker/" << basename << "_"
+             << std::setw(6) << std::setfill('0') << nstep << ".txt";;
```
asalmgren, line 156: **nit** — stray double semicolon after `".txt"`.

Source/Utils/ERF_WeatherDiagnostics.cpp
```diff
@@ -0,0 +1,350 @@
+#ifndef ERF_WEATHER_DIAGNOSTICS_H_
```
asalmgren, line 1: **low** — this is a `.cpp`, but it is wrapped in an include guard (`ERF_WEATHER_DIAGNOSTICS_H_` &hellip; `#endif` at the end of the file), carried over from the file it was copied from.

The Doxygen blocks came along too and still describe the hurricane tracker: "Compute and track the maximum wind velocity near the hurricane eye" sits above `TrackerAtStation_RainAccumulation`, "Wrapper to track the hurricane eye position over time" above `WeatherDiagnosticsTracker`, and `@param station_latitude` is documented as "Target latitude for the eye".

asalmgren:
Going to close this and we will open a new PR with more complete capability.

PR #3979, opened 2026-09-11
Add hydrostatic reconstruction options for the lateral pressure gradient

asalmgren:
Over terrain-fitted coordinates the two cell centers straddling a lateral face sit at different physical heights, so each must be carried to the height of the face before differencing.  gradp_type 0 and 1 do this with a difference of p' -- centered, and one-sided (Klemp 2011) respectively.

Both leave a residue.  "Nearly hydrostatic" means dp/dz ~ -rho g at all times, and since the base state is itself hydrostatic that transfers to the deviation: dp'/dz ~ -rho' g.  The metric term is therefore not small compared with the answer; relative to the true horizontal PGF it scales as the terrain slope times the aspect ratio of the flow feature.  Over steep terrain one is differencing two terms individually larger than their difference, even after the base state has been subtracted.  Subtracting a fixed p0 removes the balance of p' against rho0; it does nothing about the balance of p' against rho', which is a property of the instantaneous solution and so cannot be removed by any stored base state.

Add two options that remove that second balance where and when the gradient is taken, by carrying p' to the face along dp'/dz = -g rho' instead:

  gradp_type = 2   rho' held constant over the extrapolation segment
  gradp_type = 3   rho' reconstructed linearly over the segment

gradp_type = 3 is exact in the sense that if rho' is linear in z and p' is its exact hydrostatic integral, the same profile in both columns, the computed horizontal gradient vanishes identically whatever the terrain slope.  This is the finite-difference analogue of the local hydrostatic reconstruction of Botta, Klein, Langenberg & Luetzenkirchen (2004), with rho' playing the role of their locally assumed entropy profile.

rho' is formed as rho(1+q_t) - rho0(1+qv0), the same expression used by buoyancy_rhopert, which is what keeps the reconstruction consistent with the buoyancy term.  Only the lateral gradients change: compute_gradp_z is called unchanged for gradp_type 0, 2 and 3, so gpz remains an exact algebraic complement of the buoyancy term.

The new stencils need no special casing at the top or bottom of the domain and no branch on the sign of the offset, since rho' is defined in every cell and the formula is symmetric under a sign flip of the offset.

make_gradp_pert now takes the full base state and q_t rather than p0 alone. Neither new option is implemented for EB; that is enforced with an always-assert because with EB the boxes are grown by more ghost cells than q_t carries.

Verified: bit-identical to gradp_type 0 with no terrain and in the vanishing-terrain limit of a terrain-fitted mesh; max|w| identically zero for all four options for a hydrostatic atmosphere at rest over a hill; within 0.07% of gradp_type 0 and 1 for a gentle, nearly hydrostatic mountain wave.  Over a steep hill in a strongly non-hydrostatic regime the new options differ more substantially, as expected, since they discard the non-hydrostatic part of dp'/dz.


AMLattanzi:
Should this approach be contingent on using buoyancy type 1?

PR #3982, opened 2026-09-11
Fix the MOST reference height on flat stretched meshes

hgopalan:
## Summary

When `erf.most.zref` is unset, the MOST surface layer uses a wrong reference
height on a flat mesh with a stretched vertical grid. Depending on the terrain
type, the run either aborts at start-up or silently takes the wind from a
different height than the one it uses in the log law. This was found on a flat
periodic precursor column: 8x8x40, `erf.initial_dz = 10`,
`erf.grid_stretching_ratio = 1.1`, `zlo.type = "surface_layer"`,
`erf.most.z0 = 0.1`, MRF PBL, anelastic.

### 1. A query height on a cell face is never found

`set_k_indices_T`, `set_norm_indices_T` and `trilinear_interp_T` accepted a
height only if `z_lo < z < z_hi`. The default query distance is 10 m, which is
exactly the top face of a 10 m first cell, so no cell matched.

With stretching alone, the mesh becomes StaticFittedMesh + StretchedDz
(plane average, `set_k_indices_T`):

```
most.zref not specified, query distance default is 10
0::Assertion `found' failed, file "Source/BoundaryConditions/ERF_MOSTAverage.cpp", line 768, Msg: zref not found with terrain! !!!
SIGABRT
```

With `erf.terrain_type = StaticFittedMesh`, the mesh is VariableDz (local
average with interpolation, `trilinear_interp_T`):

```
Note: Switching to local averaging for MOST with defaults
most.zref not specified, query distance default is 10
0::Assertion `found' failed, file "Source/BoundaryConditions/ERF_MOSTAverage.H", line 371, Msg: MOSTAverage: Height above terrain not found, try increasing z_ref! !!!
SIGABRT
```

`erf.most.zref = 10.01` avoided both. The searches now use `[z_lo, z_hi)`
through a shared `MOSTAverage::in_cell_z`, so a height on a face belongs to the
cell above it.

### 2. The no-terrain default assumes a uniform dz

`set_k_indices_N` took cell heights from `CellSize(2) = (prob_hi - prob_lo)/nz`
even on a StretchedDz mesh. A stretched grid with terrain type None is switched
to StaticFittedMesh, so this path is reached with
`erf.terrain_type = ImmersedForcing`. On the column above, the uniform spacing is
110.6 m, so the log printed `Reference height for MOST set to 55.32406946`, while
the k index still pointed at the first cell (5 m). u* was evaluated with the
wind at 5 m and the height 55.3 m, with no warning.

The nominal `zlevels_stag` are now passed from `ERF` through `SurfaceLayer` to
`MOSTAverage`. When the mesh is not ConstantDz they set the default height
(first cell center), the cell for a given `zref`, and the height for
`k_arr_in`. ConstantDz keeps the existing uniform-spacing arithmetic.

### 3. The interpolation weight assumes equal cell heights (found in review)

`trilinear_interp_T` weighted the two cells around the query by the fraction of
the containing cell, `lk + (z - z_lo)/(z_hi - z_lo) + 1/2`. That is exact only
when the cells have equal heights. At the 10 m face of the stretched column it
returned the plain mean of the 5 m and 15.5 m cells, which is the wind at
10.25 m. With a sheared wind, u* came out 2.3e-3 high. The weight now uses the
physical heights of the two bracketing cell centers; a center beyond the
searched nodes is mirrored about the shared face. On equal cell heights this is
the same weight as before, up to round-off.

The start-up log now prints the height actually used, for example
`MOST reference height at level 0: 5`. `Inputs.rst` and `SurfaceLayer.rst`
describe the defaults.

## Before and after

The test sounding ramps the wind linearly from 10 m/s at the ground to 15 m/s
at 50 m, constant above. ERF interpolates the sounding linearly to cell centers,
so every cell holds the exact ramp value at its height. u*(0) must therefore
equal `0.41 * U(zref) / ln(zref / 0.1)` with the wind taken at zref. The last
column inverts the ramp: it is the height whose wind reproduces u*(0) with the
reported zref.

| setup (dz0 = 10 m, ratio 1.1, zref unset) | development 48f0fe0cd | this PR: zref | u*(0) | log law | wind taken at |
|---|---|---|---|---|---|
| stretching only | abort, `ERF_MOSTAverage.cpp:768` | 15.5 m | 0.938945 | 0.938945 | 15.50 m |
| `erf.terrain_type = StaticFittedMesh` | abort, `ERF_MOSTAverage.H:371` | 10 m | 0.979334 | 0.979334 | 10.00 m |
| `erf.terrain_type = ImmersedForcing` | zref 55.32 m, u* 0.681625, wind taken at 5.00 m | 5 m | 1.10045 | 1.10045 | 5.00 m |
| uniform 10 m column (ConstantDz) | zref 5 m, u* 1.10045 | 5 m | 1.10045 | 1.10045 | 5.00 m |

Before item 3, the `StaticFittedMesh` row gave u*(0) 0.98156, which is the wind
at 10.25 m.

MOST reference height per setup

Over 10 steps, the stretched ImmersedForcing column follows the uniform 10 m
column: the largest u* difference is 3.8e-4.

u* on the stretched ImmersedForcing column

## Tests

- **New regression test `MOST_Zref_Stretched`** (1 rank, about 2 s). It runs
  `Tests/test_files/MOST_Zref_Stretched/MOST_Zref_Stretched.i` in the four
  setups above. It checks each reported zref and u*(0) against the log law
  with the wind at zref (2e-5), and the stretched no-terrain column against the
  uniform column (1e-3). It aborts if the sounding is ever made too uniform to
  tell a cell from its upper neighbour.
  - A deliberately broken build, whose k lookups return the cell above, fails
    every index case by about 10%: the wind is taken at 27.05, 15.50 and
    15.00 m.
  - `run_most_zref.py --old-exe <binary>` tabulates a build without the fix
    next to the new one; the table above was produced that way.
- **New unit tests `MOSTAverageZref`** (5 tests):
  - the half-open face test;
  - stretched cell centers;
  - index agreement with `floor(zref/dz - 1/2)` on a uniform column;
  - `InterpolationOnFace`: a field equal to the cell-center height interpolates
    back to the query height, faces and the region below the first center
    included;
  - `InterpolationUniformUnchanged`.
- **Existing tests:** `ctest -L regression` passes 76/76 against the tracked
  gold files (macOS, AppleClang, Release, MPI, 2 ranks). All 182 regression
  plotfiles are bitwise identical to a build of 48f0fe0cd (`amrex_fcompare` with
  zero tolerance). `ABL_MOST_IMP_DIFF_WOA` is the only CTest deck on the
  interpolated path; it prescribes u*, t* and q* (`flux_type = "custom"`), and
  its output is unchanged. Unit tests pass 471/471.
- The build has no warnings. codespell and the trailing-whitespace check are clean.

The same change is in the ERF-Hazard fork (hgopalan/ERF#407).


AMLattanzi:
# Summary

1. The regression test cannot detect the bug it is written for

sounding_most_zref is 15.0 m/s at every height. With a spatially uniform wind, u*(0) = κU/ln(zref/z0) holds regardless of which cell the wind is sampled from, so the log-law check validates the reported zref and nothing about the pairing. A build that printed zref = 5 and sampled k = 20 would pass every assertion in run_most_zref.py — yet that pairing is exactly what was broken in the ImmersedForcing case (55.3 m height against the 5 m wind).

- Give the sounding shear — a log profile with the deck's z0 = 0.1, or a linear ramp — so u*(0) is sensitive to the sampled cell.

hgopalan:
Thanks, agreed. With a uniform wind the check only validated the reported zref, not where the wind came from. This is addressed in 55e7fdd6d.

- **Sheared sounding.** `sounding_most_zref` now ramps the wind linearly from 10 m/s at the ground to 15 m/s at 50 m, constant above. ERF interpolates the sounding linearly to the cell centers, so every cell holds the exact ramp value at its height, and so does any interpolation that is linear in z. u*(0) must therefore equal `0.41 * U(zref) / ln(zref / 0.1)`, with the wind taken at zref.
  - `run_most_zref.py` reads the profile from the sounding.
  - It inverts u*(0) for the height the wind actually came from.
  - It fails if the sounding is ever too uniform to tell a cell from its upper neighbour.
- **Checking the checker.**
  - I built a binary whose k lookups deliberately return the cell above the right one. It fails every index case by about 10%: the wind is taken at 27.05, 15.50 and 15.00 m instead of 15.5, 5 and 5 m.
  - The unfixed development binary is caught too: in the ImmersedForcing case it reports zref 55.32 m but takes the wind at 5.00 m.
- **A second pairing error the sheared check found.** `trilinear_interp_T` weighted the two cells around the query by the fraction of the containing cell, which is exact only for equal cell heights. At the 10 m face of the stretched column it returned the wind at 10.25 m, so u* came out 2.3e-3 high.
  - The weight now uses the physical heights of the two bracketing cell centers, with a center beyond the searched nodes mirrored about the shared face.
  - On equal cell heights this is the old weight, up to round-off.
  - New gtests: `InterpolationOnFace` (a field equal to the cell-center height must interpolate back to the query height, faces included) and `InterpolationUniformUnchanged`.

| setup | zref | u*(0) | log law | wind taken at |
|---|---|---|---|---|
| stretching only | 15.5 m | 0.938945 | 0.938945 | 15.50 m |
| `terrain_type = StaticFittedMesh` | 10 m | 0.979334 | 0.979334 | 10.00 m (10.25 m before the weight fix) |
| `terrain_type = ImmersedForcing` | 5 m | 1.10045 | 1.10045 | 5.00 m (development: zref 55.32 m, wind at 5.00 m) |
| uniform 10 m column | 5 m | 1.10045 | 1.10045 | 5.00 m |

**Regression.**
- `ctest -L regression` passes 76/76 against the tracked gold files.
- All 182 plotfiles are bitwise identical to a build of 48f0fe0cd.
- `ABL_MOST_IMP_DIFF_WOA` is the only CTest deck on the interpolated path. It prescribes u*, t* and q* (`flux_type = "custom"`), and its output is unchanged.
- Unit tests pass 471/471.

I have updated the PR description and figures to match.

AMLattanzi (approved):

PR #4004, opened 2026-09-14
Two-stream radiation: share RRTMGP's sun, site and surface-temperature inputs, one solar-geometry code, rad_fluxes filled, one Radiation directory, Sphinx warnings cleared

hgopalan:
Follow-up to #3950, the list of things that would confuse a user of the two radiation models. The two-stream model no longer has its own sun, site or surface-temperature inputs: it reads the keys the RRTMGP interface reads, with the same meaning, so a deck switches models by changing `erf.radiation_model` alone.

## What changed

**Inputs shared with RRTMGP** (`Source/DataStructs/ERF_RadStruct.H`)

| Before (two-stream only) | Now (shared with RRTMGP) |
|---|---|
| `erf.radiation.solar_zenith` (degrees) | `erf.fixed_solar_zenith_angle`, the **cosine** of the angle, as RRTMGP takes it (60° is 0.5); a value above 1 aborts with that explanation |
| `erf.radiation.S0` | `erf.fixed_total_solar_irradiance`; unset means 1360.9 W/m² times the Earth-Sun distance factor of the date, as RRTMGP |
| `erf.radiation.surface_temp_k` (default 300) | `erf.rad_t_sfc`, required, as for RRTMGP |
| `erf.radiation.latitude_deg`, `longitude_deg` | `erf.rad_cons_lat`, `erf.rad_cons_lon`, plus the grid's own `lat_m`/`lon_m` fields per column when a WRF or metgrid initialisation filled them (what RRTMGP does) |
| `erf.radiation.day_of_year`, `time_zone_offset_hours`, `solar_geometry_dynamic_enable`, `earth_sun_distance_enable`, simulation time modulo one day | `start_datetime` (UTC) and `erf.rad_orbital_year/eccentricity/obliquity/mvelp`; the sun follows the calendar whenever no fixed zenith angle is given |

ParmParse ignores a key nobody reads, so a deck still carrying one of the nine former keys would run with silently different settings. `RadChoice::init_params` now stops at start-up on each of them and names the replacement. A calendar sun without a start date (no `start_datetime`, no WRF/metgrid date) aborts at the first sweep and says what to set. `erf.rad_freq_in_steps` on a two-stream deck prints a warning that the model runs every step.

**One solar-geometry implementation.** `Source/Radiation/ERF_SolarGeometry.H` (Spencer 1971 declination, equation of time, azimuth) is gone. `ERF_OrbCosZenith.H` gains `orbital_calday(year, mon, day, sec)` (the leap-aware calendar day RRTMGP's interface formed inline; the interface now calls it too) and `orbital_cos_zenith_instant(jday, lat, lon, declin)`, a device-callable form of the formula `orbital_cos_zenith` uses without an averaging interval, which the host wrapper now calls. The two-stream driver takes the date from `start_time + t`, the RRTMGP time base, and runs `orbital_params` and `orbital_decl` per call for the declination and the Earth-Sun factor; the column kernel evaluates the position over each column.

**Surface temperature in RRTMGP's order.** The longwave boundary temperature is the land-surface model's `t_sfc`, else the surface layer's temperature (`SurfaceLayer::get_t_surf`, which RRTMGP already uses), else `erf.rad_t_sfc`; the prognostic surface energy balance supplies its own state ahead of the surface layer when it is on. The surface layer works in potential temperature, so its value is converted with the Exner function of the lowest cell before the σT⁴ emission (gtest). A fixed zenith angle without a fixed irradiance takes the date-scaled 1360.9 W/m² when `start_datetime` is known and the unscaled value (with a one-time note) when it is not; the orbital elements are computed once per year.

**`rad_fluxes` filled, with the top of the atmosphere.** `HasRadiation()` is true for the two-stream model, so REMORA coupling and the 2D plotfile read `rad_fluxes`; they held zeros. The array keeps RRTMGP's level layout (index k = the lower interface of layer k) and now carries one ghost cell in z so that index `khi + 1` holds the top-of-atmosphere interface: the two-stream sweep writes all nz + 1 levels (blended over the cloud fraction like the heating rates), the RRTMGP fill writes its level `nlay` there too, and the `OLR` plot variable reads that level, so the plotted OLR equals the `LW_up_TOA` of the diagnostics (verified: 424.4368 W/m² in both on the ColumnHeating deck). REMORA's surface reads at `k = 0` are unchanged. The `SW_TOA` diagnostic is a domain mean formed by the sweep, since with a calendar sun it varies across columns.

**One Radiation directory** (layout as Ann asked: the interface moves under `Source/Radiation`, not the other way). `Source/PhysicsInterfaces/Radiation/` is gone. `Source/Radiation/` holds what every model shares (`ERF_RadiationInterface.H`, `ERF_RadiationModels.H`, `ERF_OrbCosZenith.H`), `Source/Radiation/RRTMGP/` the RRTMGP interface (`ERF_Radiation.*`, `ERF_RRTMGP_*`, its memory note), `Source/Radiation/Simple/` and `Source/Radiation/TwoStream/` the two models. CMake, `Exec/Make.ERF` (a `Make.package` per directory; the RRTMGP one keeps its `USE_RRTMGP` guard), three source comments and the case READMEs follow. Pure move, no content change.

**Submodules.** The branch was started on a development that had the six `Submodules/*` gitlinks turned into symlinks by bee7f9422; that is fixed separately in #4003 (merged), and this branch is rebased on top of it.

**Docs.** `Inputs.rst`: the nine rows are gone, the RRTMGP rows say which models read them, and a paragraph before the two-stream table explains the shared inputs and the start-up abort; `theory/Radiation.rst` follows. Every other Sphinx warning of the development build is fixed in a separate commit (WindFarmModels not in a toctree, PBLschemes list indentation, four missing download files, short underlines, the deprecated theme call): `sphinx-build` is now silent.

## Verification

- Warning-free CMake build (Release, MPI, `ERF_ENABLE_ALL_WARNINGS`), `sphinx-build` with no warnings.
- 555 gtests pass. New: `ERF_GTestOrbitalGeometry.cpp` (calendar day incl. leap rules, `orbital_cos_zenith_instant` == host routine on 144 inputs, noon/midnight sanity, solstice declination and perihelion/aphelion factor) and, in the column tests, the irradiance scaling, the calendar sun over a column (a 60 N latitude field beats the constants and gives cos(60° − δ)), and the interface fluxes against the surface and top diagnostics.
- CTests `TwoStream_ColumnHeating`, `TwoStream_ColumnHeating_Terrain` (1-rank vs 2-rank CSV parity) and `Plotfile3D_TwoStreamHeatingSelection` pass.
- All 17 canonical cases with their Python checkers and the 5-case benchmark suite, on the rewritten decks (30 decks, the master deck and the three CTest decks now use the shared keys); the timing-comparison fairness check requires the sun and the surface temperature in the shared block.
- Negative tests: each of the nine retired keys, a cosine above 1, a latitude out of range, a missing `erf.rad_t_sfc` and a calendar sun without `start_datetime` stop at start-up with the intended message; with `start_datetime = "2021-06-21 18:00:00"` the run proceeds and the incident TOA flux over the default site is 1361 × 0.9518 W/m², the hand-computed cosine for that time.

**Shared-input validation on both sides.** The RRTMGP interface applies the same two checks the two-stream model does on the keys they share: `erf.fixed_solar_zenith_angle` above 1 (it is a cosine) and a non-positive `erf.rad_t_sfc` abort at start-up. RRTMGP's own use of the surface layer's potential temperature as a temperature is unchanged here and tracked in a follow-up issue.

## Left as is, deliberately

`erf.radiation.v` mirrors `erf.v`; the `erf.radiation.diag_*` keys have no RRTMGP counterpart (RRTMGP's data log is `erf.profile_rad_int`); surface albedo and emissivity fallbacks stay under `erf.radiation.` because RRTMGP has no inputs for them (it uses the LSM fields or constants).


hgopalan:
Layout changed as asked: the interface moved under `Source/Radiation`, not the two-stream files under `PhysicsInterfaces`. `Source/PhysicsInterfaces/Radiation/` is gone; the tree is now

```
Source/Radiation/
  ERF_RadiationInterface.H, ERF_RadiationModels.H, ERF_OrbCosZenith.H   (shared)
  RRTMGP/    ERF_Radiation.H/.cpp, ERF_RRTMGP_Interface.H/.cpp, ERF_RRTMGP_Utils.H, RRTMGP_Memory_Reduction.md
  Simple/    ERF_RadiationSimple.H/.cpp
  TwoStream/ the two-stream model
```

CMake (`BuildERFExe.cmake` source lists and include directories), `Exec/Make.ERF` (one `Make.package` per directory; the RRTMGP one keeps its `USE_RRTMGP` guard, `ERF_OrbCosZenith.H` moves out of that guard since the two-stream model uses it), three source comments and the case READMEs follow. Pure move, no content change; the PR description is updated. The earlier "move two-stream under PhysicsInterfaces" commit is replaced by this one (force-pushed).

Also audited, at Harish's request, for the device-code error class the Balbi model hit under CUDA (a namespace-scope constant bound to the const-reference parameter of `amrex::min`/`max` in device code): every `min`/`max` in the two-stream kernels and helpers takes locals, closure members or `amrex::Real(...)` prvalues, the namespace constants are read by value only, and the `min`/`max` calls in `ERF_OrbCosZenith.H` are in host-only routines. No instance.

Re-verified on the new layout: warning-free build, 555 gtests, the three CTests, all 17 canonical cases with their checkers and the benchmark suite, and `Docs/BuildDocs.sh` with no Sphinx warning.

asalmgren:
## Code review

Most of this is a clean directory move plus input-deck/doc rewrites; the substantive new code is `set_solar_state`, the `rad_fluxes` fill in the two-stream driver, the `orbital_calday` / `orbital_cos_zenith_instant` helpers, and the input-key migration in `RadChoice::init_params`. Four findings, one of them a blocker.

### 1. Build breaks for the `Exec/` gmake + RRTMGP path — `Exec/Make.ERF.general:343`

This PR deletes `Source/PhysicsInterfaces/Radiation/` (only `Shoc` remains under `PhysicsInterfaces` on the branch), but `Make.ERF.general` still has:

```make
ERF_RAD_DIR = $(ERF_SOURCE_DIR)/PhysicsInterfaces/Radiation
include $(ERF_RAD_DIR)/Make.package
```

`Exec/GNUmakefile:48-52` selects `Make.ERF.general` whenever `USE_RRTMGP = TRUE`, so the include fails and make stops immediately. `Exec/Make.ERF` got the corresponding rewrite (`Source/Radiation` at line 353 plus the new `Source/Radiation/RRTMGP` block at 358); `Make.ERF.general` needs the same two entries. This is the only stale `PhysicsInterfaces/Radiation` reference left in the tree.

### 2. `rad_fluxes` never carries the top-of-atmosphere interface — `Source/Radiation/TwoStream/ERF_TwoStreamColumn.H:1135`

The sweep writes each layer's *lower* interface (`(*rad_flux_out)(i,j,k,2) = F_lw_up_iface(m)`, with `m = k - kmin`), so `k = kmax` holds `F_lw_up_iface(nlev-1)`, not `F_lw_up_iface(nlev)`. `Source/IO/ERF_Plotfile2D.cpp:408` reads `OLR` from component 2 at `khi = Domain().bigEnd(2)`.

Concretely: a two-stream run that previously plotted `OLR = 0` will now plot the upward LW at the interface *below* the top layer, which differs from the `LW_up_TOA` printed in the radiation diagnostics by the top layer's emission/absorption. The same one-interface gap applies to SW up/down at `k = khi`.

RRTMGP's fill (`ERF_Radiation.cpp:818-821`, `ilay = k`) uses the identical convention, so this is inherited rather than introduced by this PR — but this PR is what makes it user-visible for the two-stream model, and the two diagnostics now openly contradict each other.

### 3. `seb_prognostic_enable` without `seb_enable` freezes the surface temperature — `Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp:590`

```cpp
} else if (rad_choice.seb_prognostic_enable && m_t_sfc[lev]) {
    t_sfc_arr = m_t_sfc[lev]->const_array(mfi); has_t_sfc_field = true;
} else if (t_surf != nullptr) { ...
```

The prognostic branch is gated only on `seb_prognostic_enable`, but the force-restore update that actually evolves `m_t_sfc` additionally requires `seb_enable` (~line 897: `if (rad_choice.seb_prognostic_enable && rad_choice.seb_enable && call_site == "post_dycore")`).

With `seb_prognostic_enable = true, seb_enable = false` — a combination `init_params` only prints a NOTE about (`ERF_RadStruct.H:1043`) — `m_t_sfc` stays at the constant `erf.rad_t_sfc` forever, yet it still wins over the surface layer's `t_surf`. The LW boundary is then a static 300 K while MOST is evolving a real surface temperature. Adding `&& rad_choice.seb_enable` to the branch would make it match the update's own gate.

### 4. Stale paths after the move (cosmetic)

- `Exec/CanonicalTests/Radiation/inputs_twostream_master:24` still points at `Source/Radiation/ERF_TwoStreamRadiation.cpp`, now `Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp`.
- `Exec/CanonicalTests/Radiation/SW_Scattering_Cloud/check_flux_accuracy.py:74` points at `Source/Radiation/ERF_TwoStreamSW.H`, now under `TwoStream/`.
- `inputs_twostream_master:118`: the `# ====` rule above the `SECTION 8` banner was dropped, leaving it without its top border.

### Checked and correct

- Every deck selecting `radiation_model = TwoStream` supplies `erf.rad_t_sfc` (including `inputs_twostream` via `FILE = shared_settings`), and each one either fixes both `fixed_solar_zenith_angle`/`fixed_total_solar_irradiance`, sets `start_datetime`, or has `sw_enabled = false` — so the new start-up and first-sweep aborts do not fire on any in-tree case.
- `set_solar_state` mirrors RRTMGP's semantics exactly (`eccf` override via fixed TSI, orbital year taken from the timestamp when `rad_orbital_year < 0`, `mu0 > 0` meaning "fixed"), and `orbital_calday` is algebraically identical to the inlined RRTMGP expression it replaces — all four arguments are `int` there, so no truncation change.
- All three `vertical_two_stream_sweep` call sites were updated for the positional `time_utc_seconds` -> `sw_down_toa` change.
- `t_surf`, `lat_m`, `lon_m` and `rad_fluxes` all share `cons`'s `BoxArray`/`DistributionMapping` (`ba_flux` is `ba` with `setRange(2,0)`), so the `const_array(mfi)` indexing in the untiled `MFIter` is valid; the new scratch `FArrayBox` follows the existing `Gpu::synchronize()`-before-destruction pattern.
- `ERF_OrbCosZenith.H` moved out of the `USE_RRTMGP` guard in the new `Source/Radiation/Make.package`, and both CMake and `Exec/Make.ERF` add `Source/Radiation` to the include path, so the two-stream include resolves in non-RRTMGP builds.
- No dangling references to `ERF_SolarGeometry.H`, `ERF_GTestSolarGeometry.cpp`, or any of the nine retired input keys outside the abort table and the docs.

hgopalan:
Thanks, all four addressed in the new commit.

1. **`Exec/Make.ERF.general`**: same rewrite as `Make.ERF`, so the gmake + RRTMGP path includes `Source/Radiation`, `Source/Radiation/RRTMGP`, `Source/Radiation/Simple` and `Source/Radiation/TwoStream`. Note that the general file had only the interface block before this PR (no `Simple`, no two-stream directory), so that path was already missing `ERF_TwoStreamRadiation.cpp` since #3950; it now lists everything `Make.ERF` lists.

2. **Top-of-atmosphere interface**: `rad_fluxes` keeps RRTMGP's level layout (index k = lower interface of layer k) and gets one ghost cell in z, so `khi + 1` holds the top interface. The two-stream sweep writes all nz + 1 levels (the cloudy scratch and the blend cover the extra level), the RRTMGP fill writes its level `nlay` into the same cell (`ilay + 1` at `k == ktop`; the flux tables are `(ncol, nlay+1)`), and `ERF_Plotfile2D.cpp` reads OLR at `khi + 1`. REMORA's `ParallelCopy` and its `k = 0` reads are untouched. The column gtest now checks `flux_lw_up[nz] == lw_up_toa`, `flux_sw_dn[nz] == sw_down_toa`, `flux_sw_up[nz] == sw_up_toa`, `flux_lw_dn[nz] == 0`, and that an isothermal column has equal surface and top LW while a cold column does not; on the ColumnHeating deck the plotted OLR (424.4368 W/m²) now equals the CSV `LW_up_TOA` of the last sweep. Comments in `ERF.H` and `ERF_MakeNewArrays.cpp` describe the layout.

3. **Prognostic branch gate**: now `seb_prognostic_enable && seb_enable`, the gate the force-restore update uses. `init_params` already turns `seb_enable` (and `seb_diagnostic_enable`) on whenever `seb_prognostic_enable` is set, so no run reached the frozen state; the two gates simply agree now.

4. **Stale paths**: the master deck's file pointer and its SECTION 7/8 banner rules, and the `SW_Scattering_Cloud` checker docstring.

Also in the commit: the driver's `gmtime` is now `gmtime_s`/`gmtime_r` (the pattern `ERF_Provenance.cpp` uses), so MSVC does not warn and the call is thread-safe.

GPU builds: there is no CUDA/HIP/SYCL toolchain here, so besides reading the device code I ran a static check over the radiation sources for the pitfalls those CI jobs catch (a namespace constant bound to a const-reference parameter of `min`/`max` in device code, host-only constructs inside device lambdas or host-device functions, reference captures). It flags the three Balbi sites on the pre-fix ERF-Hazard file and nothing in this PR's files.

Re-verified: warning-free build, 555 gtests, the three CTests, all 17 canonical cases with their checkers and the benchmark suite, Sphinx with no warnings.

hgopalan:
Merged development (#4005 and #4006) into the branch. The one conflict was the column gtest: #4005's namespace-scope kernel-launch helper and its indentation are kept, together with this branch's flux, sun and latitude/longitude harness; `ERF_TwoStreamColumn.H` merged cleanly with #4005's signature spacing.

On the GPU side: the static check I mentioned above now also flags a device lambda launched inside a `TEST()` body, the class #4005 fixed; run on this branch's files it reports nothing.

Re-verified after the merge: warning-free build, 555 gtests, the three CTests, all 17 canonical cases with their checkers and the benchmark suite.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp
```diff
@@ -539,12 +594,46 @@ TwoStreamRadiation::advance (int lev,
                         t_sfc_arr = lsm_ptr->const_array(mfi);
                         has_t_sfc_field = true;
                     }
+                } else if (rad_choice.seb_prognostic_enable && rad_choice.seb_enable && m_t_sfc[lev]) {
+                    t_sfc_arr = m_t_sfc[lev]->const_array(mfi);
+                    has_t_sfc_field = true;
+                } else if (t_surf != nullptr) {
```
asalmgren: **SurfaceLayer `t_surf` now overrides `erf.rad_t_sfc`, and in MOST flux mode it is a *potential* temperature.**

This branch puts `SurfaceLayer::get_t_surf` ahead of the model's own `m_t_sfc`. With `theta_type == HEAT_FLUX` — which is what `erf.most.surf_temp_flux = 0.0` selects, the setting used by essentially every two-stream canonical deck — `ERF_MOSTStress.H:612` sets

```
t_surf = surf_temp_flux*(...)/(u* kappa) + tm_arr
```

where `tm_arr` is the plane-averaged **potential** temperature at `zref`. So `resolve_surface_temp_k` hands theta to `compute_thermal_intensity` (sigma*T^4), and the newly *required* `erf.rad_t_sfc` becomes a no-op in that configuration.

Concrete case: a deck at ~900 hPa surface pressure with `erf.rad_t_sfc = 280` gets a longwave boundary of theta ~ T+8 K taken from the flow rather than 280 K — roughly 11% error in surface LW emission, silently.

RRTMGP has the same precedence, so the ordering may well be deliberate; the theta-vs-T substitution is the part that looks like a genuine unit mismatch. Worth a guard, or at minimum a note in the docs that `erf.rad_t_sfc` is ignored whenever the surface layer supplies a temperature.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.H
```diff
@@ -86,7 +102,13 @@ public:
                   const amrex::MultiFab* z_phys_nd,
                   const amrex::Geometry& geom,
                   LandSurface& lsm,
-                  amrex::MultiFab* qheating);
+                  amrex::MultiFab* qheating,
+                  amrex::MultiFab* rad_fluxes,
+                  const amrex::MultiFab* t_surf,
+                  const amrex::MultiFab* lat_m,
+                  const amrex::MultiFab* lon_m,
+                  amrex::Real epoch_time,
```
asalmgren: **`epoch_time` should be `double`, not `amrex::Real`.**

The value passed is `t_old[lev] + start_time` (`ERF_AdvanceRadiation.cpp:127`), i.e. Unix epoch seconds, ~1.7e9. ERF supports `PRECISION=FLOAT` / `AMReX_PRECISION=SINGLE` (`Exec/Make.ERF:15`, `CMake/SetAmrexOptions.cmake:7`). In a single-precision build a float holds that magnitude with ~128 s resolution, so the `time_t(epoch_time)` conversion in `set_solar_state` quantizes the time of day into ~2-minute jumps and can land on the wrong minute or hour entirely; the calendar sun then steps discontinuously.

The RRTMGP path deliberately uses `double` for exactly this quantity (`ERF_RadiationInterface.H:28`, and `double time_for_rad` at `ERF_AdvanceRadiation.cpp:92`). Suggest changing the parameter and the two call sites to `double` to match.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp
```diff
@@ -8,18 +8,85 @@
...
+                      amrex::Real epoch_time, bool have_datetime, int lev, int nstep)
+{
+    const bool fixed_sun = (rc.fixed_solar_zenith_angle > 0.0);
+    const bool fixed_tsi = (rc.fixed_total_solar_irradiance >= 0.0);
+    p.solar_dynamic    = !fixed_sun;
+    p.cos_zenith_fixed = rc.fixed_solar_zenith_angle;
+    p.lat_cons_rad     = rc.rad_cons_lat * PI / 180.0;
+    p.lon_cons_rad     = rc.rad_cons_lon * PI / 180.0;
+    p.S0               = fixed_tsi ? rc.fixed_total_solar_irradiance : tsi_reference;
+    if (!rc.sw_enabled || (fixed_sun && fixed_tsi)) { return; }
```
asalmgren, line 53: **A fixed zenith alone still forces `start_datetime`, which the new docs do not say.**

The early return fires only when `fixed_sun && fixed_tsi`. With `erf.fixed_solar_zenith_angle = 0.5` and no `erf.fixed_total_solar_irradiance`, execution falls through to the `have_datetime` abort just below, and the run dies at step 0.

That is a defensible design, but the new paragraph at `Docs/sphinx_doc/Inputs.rst:3365` presents the two keys independently ("leave it unset for 1360.9 W/m^2 scaled by the Earth-Sun distance factor"), so the most natural idealized deck — fix the sun angle, take the default irradiance — will not start. Either one sentence in Inputs.rst, or default the irradiance to the unscaled reference when the zenith is fixed.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp
```diff
@@ -8,18 +8,85 @@
...
+    const int day = timeinfo.tm_mday;
+    const int sec = timeinfo.tm_hour*3600 + timeinfo.tm_min*60 + timeinfo.tm_sec;
+
+    // Orbital parameters of the year (Berger 1978) unless overridden, then
+    // the declination and Earth-Sun distance factor of the day.
+    real eccen = rc.rad_orbital_eccentricity;
+    real obliq = rc.rad_orbital_obliquity;
+    real mvelp = rc.rad_orbital_mvelp;
+    real obliqr = 0.0, lambm0 = 0.0, mvelpp = 0.0;
+    orbital_params(year, eccen, obliq, mvelp, obliqr, lambm0, mvelpp);
```
asalmgren: **(perf, minor) `orbital_params` runs on every slow step.**

`orbital_params` (`Source/Radiation/ERF_OrbCosZenith.H:83`) heap-allocates six `std::vector<real>` totalling ~290 elements and evaluates three trig series on every call. The two-stream model calls `set_solar_state` once per `advance` — every step, since it has no `rad_freq_in_steps` — yet the result depends only on `year` and the `erf.rad_orbital_*` overrides.

Caching it per year (or computing it once in `define_level`) and keeping only `orbital_decl` on the per-call path would drop the repeated allocation.

asalmgren (commented):
Automated review pass (Claude Code). Most of this is a clean rename/move plus a well-guarded input migration, and the parts I checked hardest came back clean — see the bottom of this comment. Four inline findings, plus one that lands outside the diff:

**`Docs/sphinx_doc/theory/Radiation.rst:375` — stale section (medium).** The "Solar Geometry and Diurnal Cycle" section still documents `solar_geometry_dynamic_enable = true`, along with the day-of-year declination formula, the hour angle and the time-zone offset. That key is now in `retired_keys` and aborts the run at start-up (`ERF_RadStruct.H:762`), and this same file's Limitations section — updated in this PR — now says the opposite. A user following the section writes a deck that cannot start. It wants rewriting for `orbital_cos_zenith_instant` / `start_datetime`, or deleting. (The section is outside the diff hunks, hence no inline comment.)

<details><summary>Verified clean (no finding)</summary>

- The retired-key aborts, the `fixed_solar_zenith_angle > 1` guard, and the new `rad_t_sfc` requirement are all inside `if (enabled)`, so RRTMGP and Simple decks are unaffected.
- `rad_fluxes` ghost cell: `IntVect(0,0,1)` is written at `khi+1` by both models; the two-stream asserts full-column boxes, RRTMGP's `ilay = k` already implies them, and `sw_flux_up` is dimensioned `(ncol, nlay+1)` so `ilay+1` is in range. REMORA's `ParallelCopy` / `k=0` reads are unaffected. OLR for the Simple model was already zero (it never writes component 2), so `khi -> khi+1` is not a regression there.
- `&rad_flux_clear_arr` / `&rad_flux_cloudy_arr` inside the `[=]` device lambda take the address of the closure copy, not a host stack address — correct; and `rad_flux_cloudy_fab` outlives the kernel thanks to the `Gpu::synchronize()` still inside the MFIter body.
- Longitude and declination conventions in `orbital_cos_zenith_instant` match the host routine (checked noon, midnight and the +/-90 degree cases by hand); `lat_m` / `lon_m` are degrees and null outside WRF/metgrid, as the comment claims.
- Every TwoStream deck now sets `erf.rad_t_sfc` (the timing deck via `FILE = shared_settings`); no deck or Python checker still parses a retired key; the four deleted `_static/Generic_Layer3_prompt_*.md` download rows really do point at missing files.
- `Make.ERF.general` additions do not double-include with `Make.ERF` (only `Exec/GNUmakefile` uses the former).
</details>

hgopalan:
All five addressed in the new commit.

1. **Theory page, stale "Solar Geometry and Diurnal Cycle" section**: rewritten for the calendar sun (`start_datetime`, the Berger orbital elements and their `erf.rad_orbital_*` overrides, the instantaneous cos-zenith formula with the UTC day fraction and east longitude, per-column `lat_m`/`lon_m` or the constants, the irradiance rules) and the fixed-zenith alternative. The Spencer formulas, the hour angle and the time-zone offset are gone.

2. **Surface layer `t_surf` is a potential temperature**: the sweep now takes a `t_sfc_is_theta` flag, which the driver sets only for the surface layer's field, and converts that value with the Exner function of the lowest cell before it enters the σT⁴ emission; the LSM field, the prognostic state and `erf.rad_t_sfc` are temperatures and pass through unchanged. New gtest `SurfaceLayerPotentialTemperatureIsConvertedBeforeEmission`: a θ field of 300 K over a 290 K column gives exactly σ(300·Π)⁴ at the surface, more than 1 % below σ·300⁴; without a field the `erf.rad_t_sfc` value is used as is. The precedence itself stays RRTMGP's (the same θ-as-T substitution is in `ERF_Radiation.cpp:721`, which I have not touched here), and the docs now say that with a surface layer present `erf.rad_t_sfc` is only the initial prognostic value.

3. **Fixed zenith without a fixed irradiance**: no longer forces `start_datetime`. With a date the irradiance is 1360.9 W/m² times the distance factor; without one it is the unscaled reference, printed once as a NOTE. Checked: fixed cosine 0.5 alone gives `SW_TOA` = 680.45 W/m² and, with `start_datetime` on the June solstice, 658.94 W/m²; a calendar sun without a date still aborts with the input to set. `Inputs.rst` and the theory page state the rule.

4. **`orbital_params` every step**: the orbital elements are cached per year in a small `OrbitalCache` member of `TwoStreamRadiation`, so only `orbital_calday` and `orbital_decl` remain on the per-call path.

5. **`epoch_time` is `double`** in `advance()` and the helper; the two call sites already formed a `double` (`start_time` is one).

Re-verified: warning-free build, 556 gtests, the three CTests, all 17 canonical cases with their checkers and the benchmark suite, the retired-key and no-date aborts, Sphinx with no warnings, and the device-code check clean on the changed files.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp
hgopalan: Fixed: the sweep now takes t_sfc_is_theta, set only for the surface layer's field, and converts with the Exner function of the lowest cell before the emission; gtest SurfaceLayerPotentialTemperatureIsConvertedBeforeEmission. Docs say erf.rad_t_sfc is then only the initial prognostic value. See the summary comment.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.H
hgopalan: Fixed: epoch_time is double in advance() and the helper; both call sites already formed a double.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp
hgopalan: Fixed: a fixed zenith without a fixed irradiance no longer needs start_datetime (date-scaled 1360.9 W/m^2 with a date, the unscaled reference with a one-time NOTE without); Inputs.rst and the theory page state the rule. Checked: 680.45 and 658.94 W/m^2.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp
hgopalan: Fixed: the orbital elements are cached per year in an OrbitalCache member; only orbital_calday and orbital_decl run per call.

asalmgren:
Third review pass (Claude Code), against head `524d82e1`. Emphasis on whether the second pass's five findings are actually fixed in the tree rather than only claimed fixed. Nothing blocking; five observations at the bottom, none of them a blocker.

## Prior findings, checked against the current head

**Pass 1 (4 findings) — all resolved.**

- `Exec/Make.ERF.general` now has all four blocks (`Source/Radiation`, `/RRTMGP`, `/Simple`, `/TwoStream`), and `Make.ERF` has no duplicate include of any of them. No `PhysicsInterfaces/Radiation` reference survives anywhere in the tree (including `.github/`); `Source/PhysicsInterfaces/` now holds only `Shoc`.
- Top-of-atmosphere level: `rad_fluxes` is allocated with `IntVect(0,0,1)`, the sweep writes `kmax + 1` in both bands, the RRTMGP fill writes level `ilay + 1` at `k == ktop` (all four flux tables are `(ncol, nlay+1)`, so the index is in range), and `ERF_Plotfile2D.cpp:411` reads `khi + 1`. `FabArray::setVal(val)` uses `n_grow`, so the ghost level is zero-initialised — the Simple model, which never writes component 2, still plots `OLR = 0` rather than uninitialised memory. REMORA's `ParallelCopy` defaults to `src_nghost = 0`, so the extra level does not leak into the ocean lanes.
- The prognostic branch is now gated on `seb_prognostic_enable && seb_enable`, the gate the force-restore update itself uses.
- Stale paths fixed (master deck file pointer, `SW_Scattering_Cloud` checker docstring, SECTION 7/8 banner rules).

**Pass 2 (5 findings) — all resolved; one leaves a residue worth tracking.**

1. **Theory page.** The "Solar Geometry and Diurnal Cycle" section is rewritten for the calendar sun; the Spencer formulas, the hour angle, the time-zone offset and the Spencer reference are gone. The documented formula `cos(theta_z) = sin(phi) sin(delta) - cos(phi) cos(delta) cos(2 pi f + lambda)` matches `orbital_cos_zenith_instant` exactly, including the sign convention that puts solar noon at `2 pi f + lambda = pi`.
2. **Surface-layer potential temperature.** The sweep takes `t_sfc_is_theta`, the driver sets it only on the `t_surf` branch, and the kernel does `t_surface /= get_inverse_exner(i, j, kmin, state_arr)` — which returns `1/Pi` and falls back to `1.0` on unphysical input, so this is `theta * Pi` with no division hazard. Confirmed that MOST really is in theta: `ERF_MOSTStress.H` forms `tm_arr - t_surf_arr` against the plane-averaged potential temperature. The new gtest covers it.
3. **Fixed zenith without a fixed irradiance.** `fixed_sun && !have_datetime` now returns the unscaled 1360.9 W/m^2 with a one-time NOTE; with a date it falls through and picks up `eccf`. `Inputs.rst` and the theory page state the rule.
4. **Per-year orbital elements.** `orbital_params` is cached in `OrbitalCache`; only `orbital_calday` and `orbital_decl` remain on the per-call path. The refactor passes a copy `iyear` rather than the original, which is harmless — `orbital_params` never writes `iyear_AD`.
5. **`epoch_time` is `double`** through `advance()` and `set_solar_state`. `t_old`, `t_new` and `start_time` are all `double` in `ERF.H`, so both call sites were already exact.

## Input and documentation audit, both directions

I extracted every `pp.query` / `pp.get` / `pp.queryAdd` / `pp.contains` key in `ERF_RadStruct.H` and compared it with `Inputs.rst`. **Zero** keys are read by the code and missing from the docs. The nine keys that appear in the docs but not in the code are exactly the retired ones, and they appear only in the new "this aborts at start-up" paragraph, which is the right way to document them. Grid-table alignment checks clean on every `Inputs.rst` row (pipe columns and row end-columns both match their borders), so no table will be silently dropped by Sphinx. The `diag_enable` "default true -> false" edit is a documentation correction: the member was already `false` on `development`.

I also audited all 30 in-tree TwoStream decks, resolving `FILE =` includes. Every one sets `erf.rad_t_sfc`, none carries a retired key or `erf.rad_freq_in_steps`, and each either fixes the zenith angle, sets `start_datetime`, or has `sw_enabled = false` — so neither the start-up nor the first-sweep abort fires on any in-tree case. No TwoStream deck exists outside `Exec/CanonicalTests/Radiation` and `Tests/test_files`.

**Gold files: none are affected.** The three regression decks translate to numerically identical values (`cos 60 deg = 0.5` exactly, same `S0`, same 300 K), they use `zlo.type = SlipWall` so the new surface-layer precedence never engages, and there is no `Tests/ERFGoldFiles` directory for any radiation test.

## New observations (none blocking)

1. **The parity claim now has an exception.** `Source/Radiation/RRTMGP/ERF_Radiation.cpp:721` still fills RRTMGP's `t_sfc` from the surface layer's potential temperature and feeds it to `sigma T^4` unconverted. After this PR the two models therefore disagree on the longwave surface boundary by the Exner factor in exactly the MOST configuration the PR sets out to unify. Deliberately out of scope here, since it changes RRTMGP answers. Filed as #4012 — the write-up there notes that the value also reaches `t_lev(icol, 0)` and, through the `lsm_in_arr` write-back on the same line, the land-surface model's own `t_sfc` field; the error is `(1/Pi)^4 - 1`, about 13% at a surface pressure of 900 hPa and zero at 1000 hPa. Worth tracking because `Inputs.rst` now tells users the two models read these inputs "with the same meaning".
2. **Validation is one-sided.** The `fixed_solar_zenith_angle > 1` abort and the `rad_t_sfc` positive-and-finite check live in `RadChoice::init_params`, i.e. TwoStream only. A deck that switches to RRTMGP with `erf.fixed_solar_zenith_angle = 45` gets no error. Cheap to mirror into the `Radiation` constructor.
3. **Documentation nit.** The new `Inputs.rst` paragraph says that with a land-surface model or a surface layer present, `erf.rad_t_sfc` "is only the initial prognostic value". That holds when the prognostic SEB is on; otherwise the value is simply unused.
4. **Nit.** If `lsm.Get_DataIdx("t_sfc") >= 0` but `Get_Data_Ptr` returns null, the precedence chain short-circuits on the first branch and falls back to `erf.rad_t_sfc` instead of continuing on to the surface layer. The shape is inherited from the albedo and emissivity blocks above it, but the chain is longer now.
5. **Nit.** `ERF_OrbCosZenith.H` puts `typedef double real;` at global scope. It used to be reachable only under `USE_RRTMGP`; it now reaches every translation unit that includes `ERF_TwoStreamColumn.H`.

This pass is a static review against the branch; I did not build or run anything. The author reports a warning-free build, 556 gtests, the three CTests, all 17 canonical cases with their checkers, the benchmark suite, and a silent `sphinx-build`.

hgopalan:
Thanks for the third pass. Rebased onto your merge (524d82e1) and the five observations are handled in the new commit:

1. **RRTMGP's θ-as-T substitution**: out of scope here, as you say; opened #4013 with the location (`mf_to_kokkos_buffers`, the `tsurf_arr` branch), the magnitude and the suggested fix.
2. **One-sided validation**: the RRTMGP interface now applies the same two checks after it reads the shared keys: `erf.fixed_solar_zenith_angle > 1` and a non-positive or non-finite `erf.rad_t_sfc` abort at start-up with the same messages. (Not compiled here, no RRTMGP toolchain; the two ifs use only `Real`, `std::isfinite` and `std::to_string`.)
3. **Doc wording**: `erf.rad_t_sfc` is the initial prognostic value when the surface energy balance evolves one, and unused otherwise; both `Inputs.rst` and the theory page say so.
4. **Precedence chain**: each source is tried in turn, so an LSM that lists `t_sfc` but returns no pointer falls through to the surface layer and then the input value.
5. **Global `typedef double real`**: gone from `ERF_OrbCosZenith.H`; the orbital routines and the two-stream driver spell `double`.

Re-verified on the rebased head: warning-free build, 556 gtests, the three CTests, all 17 canonical cases with their checkers and the benchmark suite, the start-up aborts, Sphinx with no warnings, and the device-code check clean on the changed files.

PR #4008, opened 2026-09-14
Multilevel Restart Fix

AMLattanzi:
# Summary
Many reads from the chk file were hardcoded to `LEVEL_0` and are thus not valid for multilevel simulations. This PR corrects those reads.

## Changes
1. Define a path to the level data
2. Append path with data name
3. Check if data exists and if so read it
4. Vars impacted: MapFac, SST, TSK, Qrad, Lmask, Lat, SinPhi
5. Fix the restart of the LSM by checking if the restart file exists at a given level and reading if it does


Source/IO/ERF_Checkpoint.cpp
```diff
@@ -1168,8 +1168,11 @@ ERF::ReadCheckpointFile ()
             two_stream_rad.read_checkpoint(lev, restart_chkfile);
         }
 
+        // Set path to level directory
+        std::string PathToChkfileData = restart_chkfile + "/Level_" + std::to_string(lev) + "/";
```
asalmgren: Nit, nothing is broken today: this hand-rolls AMReX's checkpoint path layout, while the rest of this file probes with `MultiFabFileFullPrefix(lev, restart_chkfile, "Level_", name) + "_H"` (see `read_if_present` around line 1017, `read_most_var` around line 1450, and `TwoStreamRadiation::read_checkpoint`). The two agree, so this is correct as written — but building the probe path from the same helper that the `VisMF::Read` calls use would keep the probe and the read from ever drifting apart again, which is precisely the failure mode this PR exists to fix.

asalmgren (commented):
The change looks correct to me, and it fixes two distinct bugs.

**Probe/read mismatch.** The existence probes were hardcoded to `Level_0/...` while the subsequent `VisMF::Read` opened `Level_<lev>/...` via `MultiFabFileFullPrefix`. For `lev > 0`, a level that legitimately has no SST/TSK/LMASK/etc. would see level 0's header, enter the block, and then abort on the nonexistent file. `PathToChkfileData` reproduces exactly what AMReX builds (`LevelPath` is unpadded `Level_<lev>`), so probe and read now agree for Qrad, MapFactor_{mx,ux,vx}, SST, TSK, LMASK, LAT and SinPhi.

**The `ngv` line is a real second fix, not cosmetic.** On the write side `ngv` is derived from `ng`, which the *unconditional* LMASK block has just set to `vars_new[lev][Vars::cons].nGrowVect()` with `[2]=0`. On read, `ng` previously carried over from whichever optional block ran last, so with no LMASK file present `ngv` fell back to the `mapfac[v_x]` ghost count — which need not match what was written, and disagrees with how `lat_m`/`sinPhi_m`/`cosPhi_m` are allocated at init (`ERF_InitFromWRFInput.cpp:476`, `ERF_InitFromMetgrid.cpp:170`, both `cons.nGrowVect()` with `[2]=0`). Setting it unconditionally makes read, write and init agree.

I didn't find a regression path: every touched block's probe is now strictly more accurate, `sst_lev[lev]`/`tsk_lev[lev]`/`lmask_lev[lev]` are sized and allocated for all levels by `MakeNewLevelFromScratch` before the data loop in `ReadCheckpointFile`, and `ng` is not read after the new `ngv` assignment.

One thing worth folding into this PR, since it's the same class of bug in the same function but sits outside the diff so I can't comment on it inline:

**`Source/IO/ERF_Checkpoint.cpp:1391` — the same hardcoded level-0 gate survives.**

```cpp
if (amrex::FileExists(LsmRestartDir + "/Level_0.nc")) {
```

This still controls `lsm.Read_Lsm_Restart(lev, ...)` for *every* level in the loop below it — the identical "probe Level_0, read Level_N" pattern this PR removes everywhere else, about 30 lines past the last hunk. `NOAHMP::Read_Lsm_Restart` routes through `with_land_comm(...)` → `noahmpio.ReadRestart(dir)`, so a level whose `Level_<lev>.nc` was never produced (no land-owning ranks at that level when the checkpoint was written) will still be attempted because level 0's file happens to exist.

Source/IO/ERF_Checkpoint.cpp
AMLattanzi: I agree. The latest commit utilizes this approach.

AMLattanzi:
Confirmed on Perlmutter that this PR enables restart with dynamic AMR.

Source/IO/ERF_Checkpoint.cpp
```diff
@@ -1384,21 +1385,23 @@ ERF::ReadCheckpointFile ()
...
-        if (amrex::FileExists(LsmRestartDir + "/Level_0.nc")) {
-            for (int lev = 0; lev <= finest_level; ++lev) {
+        for (int lev = 0; lev <= finest_level; ++lev) {
+            std::string LsmRestartFile = LsmRestartDir + "/Level_" + std::to_string(lev) + ".nc";
+            if (amrex::FileExists(LsmRestartFile)) {
                 lsm.Read_Lsm_Restart(lev, LsmRestartDir);
+                Print() << "Restored full NoahMP prognostic state from "
+                        << LsmRestartFile << " at level " << lev << std::endl;
+            } else {
+                Print() << "NoahMP restart file " << LsmRestartFile << " is not present at level " << lev << std::endl;
```
asalmgren: These three new messages are NoahMP-specific, but they sit inside `if (solverChoice.lsm_type != LandSurfaceType::None)`, which also covers `LandSurfaceType::SLM`. SLM's `Write_Lsm_Restart` is the `NullSurf` no-op, so `noahmp_restart/` is always created and always empty for SLM runs. Every SLM restart will therefore print

```
NoahMP restart file .../noahmp_restart/Level_0.nc is not present at level 0
WARNING: NoahMP will cold-initialize from wrfinput at level 0
```

for a model that is not NoahMP and does not read wrfinput for its state. Suggest gating these prints (or the whole probe) on `solverChoice.lsm_type == LandSurfaceType::NOAHMP`.

asalmgren (commented):
Reviewed the switch to `MultiFabFileFullPrefix` and the NoahMP restart loop restructuring. The core change looks correct: at `lev == 0` the generated paths are byte-identical to the old hardcoded strings, every converted variable (Qrad, MapFactor_{mx,ux,vx}, SST, TSK, LMASK, LAT, SinPhi) is written per-level by `WriteCheckpointFile`, and the new `ngv` derived from `vars_new[lev][Vars::cons].nGrowVect()` matches both the write side and how `lat_m`/`sinPhi_m` are sized at init. The NoahMP loop change is a real fix too -- the old form called `Read_Lsm_Restart` on *all* levels whenever `Level_0.nc` existed, which hits `NoahmpIO_abort()` in `NoahmpReadRestartMod.F90:41` for a missing fine-level file. The guarding `FileExists` is evaluated identically on all ranks, so the MPI-collective `Read_Lsm_Restart` stays safe.

Two low-severity comments inline, plus one observation:

**Observation (not a defect in this diff):** on restart, `InitData_post` calls `Interp2DArrays(lev, ...)` for `lev >= 1` at `Source/ERF.cpp:1213`, *after* `ReadCheckpointFile`. For a fine level with no `nc_init_file[lev]`, that runs `interp_mapfac_from_coarse(lev)` and `interp_land_masks_from_coarse(lev)`, overwriting the fine-level MapFactor and land-mask values this PR just fixed reading. The values are static so results should be unchanged, but the fine-level MapFactor/LMASK reads are effectively dead on the tagging-created-level path -- worth confirming that is intended.

Source/IO/ERF_Checkpoint.cpp
```diff
@@ -1384,21 +1385,23 @@ ERF::ReadCheckpointFile ()
         // pulls it into the physics state on the first Advance (issue #3255).
         // Legacy checkpoints without this directory fall back to cold-init.
         std::string LsmRestartDir(restart_chkfile + "/noahmp_restart");
-        if (amrex::FileExists(LsmRestartDir + "/Level_0.nc")) {
-            for (int lev = 0; lev <= finest_level; ++lev) {
+        for (int lev = 0; lev <= finest_level; ++lev) {
+            std::string LsmRestartFile = LsmRestartDir + "/Level_" + std::to_string(lev) + ".nc";
```
asalmgren: This re-implements a filename convention that actually lives in Fortran (`NoahmpWriteRestartMod.F90` / `NoahmpReadRestartMod.F90:33-35`), where the level is formatted with `'(I1.1)'` into a `character(len=1)`. For `lev >= 10` the Fortran writer emits `Level_*.nc` while this check looks for `Level_10.nc`, so the probe fails and the level silently cold-initializes instead of restoring.

Practically unreachable today (it needs `amr.max_level >= 10`), but the duplicated convention means the C++ and Fortran sides can now drift independently. Worth at least a comment pointing at the Fortran format, or widening the Fortran field.

AMLattanzi: This is a good catch. I think moving to a Noah PR that uses a scratch buffer with `I0` and trims would be a better approach than in this PR.

Source/IO/ERF_Checkpoint.cpp
AMLattanzi: The block is now gated with an if noahmp check.

Source/IO/ERF_Checkpoint.cpp
AMLattanzi: Addressed in NOAH [PR 12](https://github.com/erf-model/noahmp/pull/12)

asalmgren:
Reviewed this — the core fix looks right. The eight `restart_chkfile + "/Level_0/..."` probes were checking level 0 while the matching `VisMF::Read` used `MultiFabFileFullPrefix(lev, ...)`, so a multilevel restart would try to read e.g. `Level_1/SST_0` whenever level 0 had written one. Routing the probes through `MultiFabFileFullPrefix(lev, chk, "Level_", "<name>_H")` makes them symmetric with the write-side guards, and level-0 behavior is unchanged.

Also worth noting the `ngv` change is a genuine fix rather than churn: on the write side the LMASK block is unconditional, so `ngv` is always `vars_new[lev][Vars::cons].nGrowVect()` with `ngv[2]=0`; on the read side that block is guarded by `FileExists`, so a checkpoint without LMASK left `ng` at `mapfac[lev][v_x]->nGrowVect()`. Those only agree by luck when `ngrow_state == 3`.

One small thing:

**`Source/IO/ERF_Checkpoint.cpp:1361`** — the `lsm_step` restore is now gated on `lsm_type == LandSurfaceType::NOAHMP`, but `WriteCheckpointFile` still *writes* `chk/lsm_step` for every `lsm_type != LandSurfaceType::None` (around line 159). The write and read guards no longer agree. It's benign today only because `SLM` inherits the no-op `Set_LSM_Step` from `NullSurf` (`Source/LandSurfaceModel/Null/ERF_NullSurf.H:133`) — but if SLM or a future model ever overrides it, an SLM restart will silently drop a counter that was faithfully checkpointed, with no warning. The "legacy checkpoint without lsm_step" message also becomes unreachable for SLM.

Suggestion: keep the `lsm_step` read under the existing `!= None` guard and gate only the `noahmp_restart` loop on NOAHMP. Or drop the NOAHMP gate entirely, since `Read_Lsm_Restart` is already a no-op for non-NoahMP models — that would also remove the now-redundant nesting of `if (lsm_type == NOAHMP)` inside `if (lsm_type != None)`.

PR #4011, opened 2026-09-14
Multilevel refinement enhancements for WRF input pathway: terrain smoothing, vertical refinement, radiation support

adamwise95:
### Summary
This PR enables multilevel AMR with WRF input initialization and terrain smoothing, adding three major features:

1. **STF/Sullivan terrain smoothing with multilevel WRF input** (`terrain_smoothing = 1` or `2`)
2. **Optional atmospheric state interpolation from coarse level** (`erf.interp_atmos_from_coarse`)
3. **Automatic radiation memory pool expansion for vertical refinement**

### Problem Being Solved

**Issue 1: Terrain smoothing incompatible with multilevel WRF input**
Previously, using `terrain_smoothing = 1` (STF) or `terrain_smoothing = 2` (Sullivan) with multilevel AMR initialized from WRF input files would abort. The STF and Sullivan algorithms are only defined for full-column domains, but fine levels needed a way to construct terrain-fitted coordinates that maintain consistency with the smoothed coarse level.

**Issue 2: Time-mismatched WRF input files**
When nested WRF domains have different initialization times (e.g., `wrfinput_d01` at t=0h, `wrfinput_d02` at t=6h), directly reading atmospheric state from the fine-level file can introduce:
- Stale/inconsistent velocity fields
- Discontinuities at the coarse-fine interface
- Initialization instabilities

**Issue 3: Radiation crashes with vertical refinement**
When fine levels have more vertical cells than coarse levels (vertical refinement), the RRTMGP memory pool allocated for the coarse level was too small, causing crashes. We also cannot currently run radiation when fine levels do not extend the full column.

#### Feature 1: STF/Sullivan Terrain Smoothing with Multilevel WRF Input

**Required runtime option:**
```bash
erf.amr_terrain_refinement = transform  # REQUIRED when terrain_smoothing = 1 or 2 with multilevel WRF input
```

**Problem solved:**
STF (Smoothed Terrain Following) and Sullivan terrain-following transformations iteratively smooth terrain to prevent grid layer crossing and excessive vertical compression. These algorithms are only defined for domains spanning the full atmospheric column. Fine levels in multilevel AMR typically don't span the full column, requiring a different approach.

**Solution - Transform Mode:**
The code **requires** Transform mode when using terrain smoothing with multilevel WRF input (Interpolate mode is not supported and will abort).

**How Transform mode works:**
- Reads fine-resolution terrain (PH+PHB) from wrfinput file (e.g., wrfinput_d02)
- Computes difference between fine surface and interpolated surface: `delta = z_fine_surface - z_interp_surface`
- Applies correction with linear decay: `z_final(k) = z_interp(k) + decay(k) * delta`
- Decay: 1.0 at surface → 0.0 at top of fine grid
- Result: Fine terrain details near surface, C0 continuity at coarse-fine interface
- This is similar to the previous implementation for multilevel ideal simulations with terrain.

**Key behavior:**
- Level 0: Full STF/Sullivan smoothing algorithm applied
- Fine levels (lev > 0): No additional smoothing; Transform mode blends fine terrain onto smoothed coarse mesh
- Ensures smooth coarse-fine interface while preserving fine-scale terrain features

**Error handling:**
If `terrain_smoothing = 1` or `2` is used with multilevel WRF input and Transform mode is not set, the code will abort with:
```
terrain_smoothing = N with wrfinput initialization on level > 0 requires
erf.amr_terrain_refinement = transform (not interpolate)
```

#### Feature 2: Interpolate Atmospheric State from Coarse Level

**New runtime option:**
```bash
erf.interp_atmos_from_coarse = true
```

**Behavior when enabled:**
- **Terrain and surface fields** read from fine-level wrfinput file (PH+PHB, SST, TSK, land masks, LSM variables)
- **Atmospheric state** (U, V, W, THM, density, moisture) interpolated from running coarse level via `FillCoarsePatch`
- Results in physically consistent atmospheric state with high-resolution surface properties

**Implementation approach:**
- New function `init_from_wrfinput_surface_only()` reads only terrain/surface fields from wrfinput
- LSM initialization deferred until after `FillCoarsePatch` provides valid atmospheric state
- Modified `make_lsm_at_level()` to conditionally defer LSM setup (only for lev > 0)

**Clarification on "surface only":**
Despite the function name, `init_from_wrfinput_surface_only()` reads:
- **Full vertical grid structure** (PH+PHB for all levels)
- Surface properties (SST, TSK, surface pressure, land masks)
- LSM variables (soil temperature/moisture, vegetation)
- Map factors and WRF vertical coordinates

It does NOT read atmospheric state variables (U, V, W, temperature, moisture).

**Interaction with Feature 1:**
The `interp_atmos_from_coarse` option works seamlessly with terrain smoothing:
- Terrain grid (z_phys_nd) is always constructed from PH+PHB in wrfinput (even in surface-only mode)
- Transform mode blends fine terrain onto interpolated coarse mesh
- Only atmospheric state is interpolated from coarse level

#### Feature 3: Automatic Radiation Memory Pool Expansion

**Problem:**
RRTMGP radiation uses a Kokkos memory pool sized for the initial vertical grid. When a fine level with vertical refinement (more vertical layers) is created, the pool was too small.

**Solution:**
- Track maximum `nlay` seen across all levels (`pool_max_nlay`)
- Detect when a new level has `nlay > pool_max_nlay`
- Automatically finalize old pool and reinitialize with larger size
- Pool expansion happens once per simulation when fine level is first created

**Nested patch optimization:**
Fine levels that don't reach the model top (nested patches) skip radiation entirely and use interpolated heating rates from coarse level. Memory pool expansion only occurs for full-column refined levels.

### Modified Files

**New files:**
- `Source/Initialization/ERF_InitFromWRFInput_SurfaceOnly.cpp` - Surface-only initialization for `interp_atmos_from_coarse`

**Modified files:**
- `Source/ERF_MakeNewLevel.cpp` - Deferred LSM initialization logic, fixed level check for surface-only mode
- `Source/ERF_MakeNewArrays.cpp` - Added Transform/Interpolate mode selection for fine terrain
- `Source/DataStructs/ERF_DataStruct.H` - Added `interp_atmos_from_coarse` runtime parameter
- `Source/ERF.H` - Added `init_from_wrfinput_surface_only()` declaration
- `Source/Utils/ERF_TerrainMetrics.cpp` - Transform mode implementation, fine terrain grid construction
- `Source/Initialization/ERF_InitFromWRFInput.cpp` - Terrain interpolation and Transform mode logic for WRF input
- `Source/PhysicsInterfaces/Radiation/ERF_RRTMGP_Interface.cpp` - Pool expansion logic with max nlay tracking
- `Source/PhysicsInterfaces/Radiation/ERF_Radiation.H` - Nested patch detection and debug output
- `Source/PhysicsInterfaces/Radiation/ERF_Radiation.cpp` - Added buffer allocation logging
- `Source/TimeIntegration/ERF_AdvanceRadiation.cpp` - Boundary filling for nested patches
- `CMake/BuildERFExe.cmake` - Added new source file to build
- `Source/Initialization/Make.package` - Added new source file to build

### Testing

**Test 1: STF terrain smoothing with multilevel WRF input**
- 2-level AMR with WRF nested domains (wrfinput_d01, wrfinput_d02)
- `terrain_smoothing = 1` (STF)
- `erf.amr_terrain_refinement = transform`
- `erf.interp_atmos_from_coarse = true`
- LSM enabled (Noah-MP)
- Radiation enabled (RRTMGP)
- Result: Level 0 vertical grid smoothed, fine level blends high-res terrain with linear decay, stable initialization, fine terrain detail preserved, consistent atmospheric state, LSM properly initialized, velocities look realistic and good multiple hours in.

**Test 2: Radiation with vertical refinement**
- 2-level AMR with WRF nested domains (wrfinput_d01, wrfinput_d02)
- `terrain_smoothing = 1` (STF)
- `erf.amr_terrain_refinement = transform`
- `erf.interp_atmos_from_coarse = true`
- LSM enabled (Noah-MP)
- Radiation enabled (RRTMGP)
- Level 0: 100 vertical layers
- Level 1: 200 vertical layers (2x refinement)
- Memory pool automatically expanded 2x
- No crashes, radiation runs successfully on fine level

**Test 3: Radiation with vertical refinement and finer level does not reach model top**
- 2-level AMR with WRF nested domains (wrfinput_d01, wrfinput_d02)
- `terrain_smoothing = 1` (STF)
- `erf.amr_terrain_refinement = transform`
- `erf.interp_atmos_from_coarse = true`
- LSM enabled (Noah-MP)
- Radiation enabled (RRTMGP)
- Level 0: 100 vertical layers
- Level 1: 49 vertical layers (but with 2x refinement):
- `amr.ref_ratio_vect = 7 7 2`
- `erf.refinement_indicators = box1`
- `erf.box1.max_level = 1`
- `erf.box1.in_box_lo_indices = 650  650   0`
- `erf.box1.in_box_hi_indices = 749  749  49`
- Memory pool does not expand
- Does not reach model top → radiation skipped, uses interpolated rates

### Backward Compatibility

- `erf.interp_atmos_from_coarse` defaults to `false` - existing behavior unchanged
- Radiation pool expansion is automatic and transparent
- **Breaking change for terrain smoothing + multilevel WRF input**: Previously would abort; now requires `erf.amr_terrain_refinement = transform`
- Single-level simulations and non-WRF initialization methods are unaffected

### Known Limitations

- **Transform mode requirement**: `terrain_smoothing = 1` or `2` with multilevel WRF input requires `erf.amr_terrain_refinement = transform` (Interpolate mode not supported)
- `interp_atmos_from_coarse` requires WRF input initialization (`init_type = WRFInput`)
- Only applies to fine levels (lev > 0); level 0 always reads full atmospheric state
- Radiation pool expansion may show Kokkos warnings about type conversions (cosmetic, not fatal)
- Transform mode requires every box on fine level to reach the surface (increase `amr.max_grid_size_z` if needed)

### Future Work

- Extend nested patch radiation to use 1D column solver instead of full skip

### DIFF
--- a/Source/ERF_MakeNewLevel.cpp
+++ b/Source/ERF_MakeNewLevel.cpp
@@ -486,16 +486,42 @@ ERF::MakeNewLevelFromCoarse (int lev, Real time, const BoxArray& ba,
         vars_new[lev][Vars::zvel].setVal(0.0); vars_old[lev][Vars::zvel].setVal(0.0);
 
         AMREX_ALWAYS_ASSERT(solverChoice.terrain_type == TerrainType::StaticFittedMesh);
+
+        //
+        // CHOOSE INITIALIZATION PATH:
+        // If interp_atmos_from_coarse is enabled for a finer level created at later time,
+        // read only surface fields from wrfinput and interpolate atmospheric state from coarse.
+        // Otherwise use the standard path of reading all fields from wrfinput.
+        //
+        bool use_surface_only = solverChoice.interp_atmos_from_coarse && (lev > 0) && (time > 0.0);
+
         if (solverChoice.init_type == InitType::Metgrid) {
             init_from_metgrid(lev);
         } else if (solverChoice.init_type == InitType::WRFInput) {
-            init_from_wrfinput(lev, *mf_PSFC[lev]);
+            if (use_surface_only) {
+                amrex::Print() << "Using interp_atmos_from_coarse mode at level " << lev << ":\n";
+                amrex::Print() << "  - Reading surface fields from wrfinput\n";
+                amrex::Print() << "  - Atmospheric state will be interpolated from level " << lev-1 << "\n";
+                init_from_wrfinput_surface_only(lev, *mf_PSFC[lev]);
+            } else {
+                init_from_wrfinput(lev, *mf_PSFC[lev]);
+            }
         }
         init_zphys(lev, time);
         update_terrain_arrays(lev);
         make_physbcs(lev);
 
         dz_min[lev] = (*detJ_cc[lev]).min(0) * geom[lev].CellSize(2);
+
+        //
+        // If we used surface-only init, we need to rebuild the base state and
+        // interpolate the atmospheric state from coarse (just like a level with no init file)
+        //
+        if (use_surface_only) {
+            rebuild_base_state_from_wrfinput(lev, base_state[lev]);
+            (*physbcs_base[lev])(base_state[lev],0,base_state[lev].nComp(),base_state[lev].nGrowVect());
+            FillCoarsePatch(lev, time);
+        }
 
     } else {
 #endif
--- a/Source/ERF_MakeNewLevel.cpp
+++ b/Source/ERF_MakeNewLevel.cpp
@@ -489,11 +489,11 @@ ERF::MakeNewLevelFromCoarse (int lev, Real time, const BoxArray& ba,
 
         //
         // CHOOSE INITIALIZATION PATH:
-        // If interp_atmos_from_coarse is enabled for a finer level created at later time,
+        // If interp_atmos_from_coarse is enabled for a finer level,
         // read only surface fields from wrfinput and interpolate atmospheric state from coarse.
         // Otherwise use the standard path of reading all fields from wrfinput.
         //
-        bool use_surface_only = solverChoice.interp_atmos_from_coarse && (lev > 0) && (time > 0.0);
+        bool use_surface_only = solverChoice.interp_atmos_from_coarse && (lev > 0);
 
         if (solverChoice.init_type == InitType::Metgrid) {
             init_from_metgrid(lev);
@@ -521,6 +521,15 @@ ERF::MakeNewLevelFromCoarse (int lev, Real time, const BoxArray& ba,
             rebuild_base_state_from_wrfinput(lev, base_state[lev]);
             (*physbcs_base[lev])(base_state[lev],0,base_state[lev].nComp(),base_state[lev].nGrowVect());
             FillCoarsePatch(lev, time);
+
+            // Now initialize LSM with the properly filled atmospheric state
+            if (solverChoice.lsm_type != LandSurfaceType::None) {
+                amrex::Print() << "Initializing LSM at level " << lev << " after FillCoarsePatch\n";
+                IntVect RefRatio(1);
+                for (int l = 0; l < lev; ++l) { RefRatio *= refRatio(l); }
+                lsm.Init(lev, vars_new[lev][Vars::cons], Geom(lev), Geom(0),
+                         domain_bcs_type, RefRatio, zero, nc_init_file);
+            }
         }
 
     } else {
@@ -607,7 +616,9 @@ ERF::MakeNewLevelFromCoarse (int lev, Real time, const BoxArray& ba,
     }
 
     //********************************************************************************************
-    // Land Surface Model
+    // Land Surface Model - setup data structures
+    // NOTE: Actual LSM initialization (lsm.Init) is deferred if using interp_atmos_from_coarse
+    //       because it needs valid atmospheric state from FillCoarsePatch
     // *******************************************************************************************
     int lsm_data_size  = lsm.Get_Data_Size();
     int lsm_flux_size  = lsm.Get_Flux_Size();
@@ -616,7 +627,19 @@ ERF::MakeNewLevelFromCoarse (int lev, Real time, const BoxArray& ba,
     lsm_flux[lev].resize(lsm_flux_size);
     lsm_flux_name.resize(lsm_flux_size);
     lsm.Define(lev, solverChoice);
-    if (solverChoice.lsm_type != LandSurfaceType::None) {
+
+    // Check if we'll be using surface-only init (atmospheric state comes later from FillCoarsePatch)
+    bool will_use_surface_only = false;
+#ifdef ERF_USE_NETCDF
+    if (!nc_init_file[lev].empty() &&
+        (solverChoice.init_type == InitType::WRFInput || solverChoice.init_type == InitType::Metgrid)) {
+        will_use_surface_only = solverChoice.interp_atmos_from_coarse;
+    }
+#endif
+
+    // Only initialize LSM now if we're NOT using surface-only init
+    // (for surface-only, LSM init must wait until after FillCoarsePatch provides atmospheric state)
+    if (solverChoice.lsm_type != LandSurfaceType::None && !will_use_surface_only) {
         //
         // A level with no land file of its own takes its LSM state from level 0 rather
         // than from its parent (see NOAHMP::interp_from_lev0, which is handed Geom(0)
--- a/Source/TimeIntegration/ERF_AdvanceRadiation.cpp
+++ b/Source/TimeIntegration/ERF_AdvanceRadiation.cpp
@@ -45,5 +45,22 @@ void ERF::advance_radiation (int lev,
                       qheating_rates[lev].get(), rad_fluxes[lev].get(),
                       z_phys_nd[lev].get()     , lat_ptr, lon_ptr,
                       lsm_updated);
+
+        // Fill ghost cells after radiation computes (needed for interpolation to finer levels)
+        // This should be fast since it only fills this level's own ghost cells
+        if (solverChoice.rad_type != RadiationType::None && !rad[lev]->is_nested_patch()) {
+            qheating_rates[lev]->FillBoundary(geom[lev].periodicity());
+        }
+
+        // For nested patches (fine levels that don't reach model top), radiation
+        // was skipped. Interpolate heating rates from parent level.
+        if (lev > 0 && rad[lev]->is_nested_patch()) {
+            InterpFromCoarseLevel(*qheating_rates[lev], qheating_rates[lev]->nGrowVect(),
+                                  IntVect(0,0,0),
+                                  *qheating_rates[lev-1], 0, 0, 2,
+                                  geom[lev-1], geom[lev],
+                                  refRatio(lev-1), &cell_cons_interp,
+                                  domain_bcs_type, BCVars::cons_bc);
+        }
     }
 }
--- a/Source/TimeIntegration/ERF_AdvanceRadiation.cpp
+++ b/Source/TimeIntegration/ERF_AdvanceRadiation.cpp
@@ -55,12 +55,20 @@ void ERF::advance_radiation (int lev,
         // For nested patches (fine levels that don't reach model top), radiation
         // was skipped. Interpolate heating rates from parent level.
         if (lev > 0 && rad[lev]->is_nested_patch()) {
+            // Ensure parent level's ghost cells are filled before interpolation
+            // This is needed even when radiation doesn't run this step, especially
+            // with two-way coupling where the grid structure may have changed
+            if (!rad[lev-1]->is_nested_patch()) {
+                qheating_rates[lev-1]->FillBoundary(geom[lev-1].periodicity());
+            }
+
             InterpFromCoarseLevel(*qheating_rates[lev], qheating_rates[lev]->nGrowVect(),
                                   IntVect(0,0,0),
                                   *qheating_rates[lev-1], 0, 0, 2,
                                   geom[lev-1], geom[lev],
                                   refRatio(lev-1), &cell_cons_interp,
                                   domain_bcs_type, BCVars::cons_bc);
+
         }
     }
 }
--- a/Source/TimeIntegration/ERF_AdvanceRadiation.cpp
+++ b/Source/TimeIntegration/ERF_AdvanceRadiation.cpp
@@ -69,6 +69,36 @@ void ERF::advance_radiation (int lev,
                                   refRatio(lev-1), &cell_cons_interp,
                                   domain_bcs_type, BCVars::cons_bc);
 
+            // Also interpolate LSM radiation output fields (surface fluxes needed by NoahMP)
+            // These are 2D surface fields (k=0 only) that may have no ghost cells
+            if (solverChoice.lsm_type != LandSurfaceType::None) {
+                Vector<std::string> lsm_output_names = rad[lev]->get_lsm_output_varnames();
+
+                for (int i = 0; i < lsm_output_names.size(); ++i) {
+                    int varIdx_fine = lsm.Get_DataIdx(lev, lsm_output_names[i]);
+                    int varIdx_coarse = lsm.Get_DataIdx(lev-1, lsm_output_names[i]);
+                    if (varIdx_fine >= 0 && varIdx_coarse >= 0) {
+                        MultiFab* lsm_fine = lsm.Get_Data_Ptr(lev, varIdx_fine);
+                        MultiFab* lsm_coarse = lsm.Get_Data_Ptr(lev-1, varIdx_coarse);
+                        if (lsm_fine && lsm_coarse && lsm_coarse->nComp() > 0) {
+                            // Create temporary coarse MultiFab with 1 ghost cell for safe interpolation
+                            MultiFab tmp_coarse(lsm_coarse->boxArray(), lsm_coarse->DistributionMap(),
+                                                lsm_coarse->nComp(), 1);
+                            MultiFab::Copy(tmp_coarse, *lsm_coarse, 0, 0, lsm_coarse->nComp(), 0);
+                            tmp_coarse.FillBoundary(geom[lev-1].periodicity());
+
+                            // Now interpolate from tmp_coarse (with ghost) to lsm_fine (no ghost needed)
+                            InterpFromCoarseLevel(*lsm_fine, IntVect(0,0,0),
+                                                  IntVect(0,0,0),
+                                                  tmp_coarse, 0, 0, lsm_coarse->nComp(),
+                                                  geom[lev-1], geom[lev],
+                                                  refRatio(lev-1), &cell_cons_interp,
+                                                  domain_bcs_type, BCVars::cons_bc);
+                        }
+                    }
+                }
+            }
+
         }
     }
 }
--- a/CMake/BuildERFExe.cmake
+++ b/CMake/BuildERFExe.cmake
@@ -209,6 +209,7 @@ function(build_erf_lib erf_lib_name)
   if(ERF_ENABLE_NETCDF)
     target_sources(${erf_lib_name} PRIVATE
                    ${SRC_DIR}/Initialization/ERF_InitFromWRFInput.cpp
+                   ${SRC_DIR}/Initialization/ERF_InitFromWRFInput_SurfaceOnly.cpp
                    ${SRC_DIR}/Initialization/ERF_InitFromMetgrid.cpp
                    ${SRC_DIR}/Initialization/ERF_InitFromNCFile.cpp
                    ${SRC_DIR}/IO/ERF_NCInterface.cpp
--- a/Source/DataStructs/ERF_DataStruct.H
+++ b/Source/DataStructs/ERF_DataStruct.H
@@ -862,6 +862,10 @@ struct SolverChoice {
                          "onto the nodes rather than reconstructing them).");
         }
 
+        // Interpolate atmospheric state from coarse level instead of reading from wrfinput
+        // when creating a finer level from coarse at a later time?
+        pp.queryAdd("interp_atmos_from_coarse", interp_atmos_from_coarse);
+
         // Rebalance wrf state?
         pp.queryAdd("rebalance_wrf_input", rebalance_wrf_input);
 
@@ -2169,6 +2173,9 @@ struct SolverChoice {
 
     // Average the input z-face heights onto the nodes rather than reconstructing them?
     bool avg_grid_faces_to_nodes {false};
+
+    // Interpolate atmospheric state from coarse when creating finer level at later time?
+    bool interp_atmos_from_coarse {false};
 
     // Use forest canopy model?
     bool do_forest_drag {false};
--- a/Source/ERF.H
+++ b/Source/ERF.H
@@ -799,6 +799,7 @@ public:
 
 #ifdef ERF_USE_NETCDF
     void init_from_wrfinput (int lev, amrex::MultiFab& mf_PSFC);
+    void init_from_wrfinput_surface_only (int lev, amrex::MultiFab& mf_PSFC);
     void init_from_metgrid  (int lev);
     void init_from_ncfile   (int lev);
 
--- a/Source/ERF_MakeNewArrays.cpp
+++ b/Source/ERF_MakeNewArrays.cpp
@@ -499,11 +499,9 @@ ERF::init_stuff (int lev, const BoxArray& ba, const DistributionMapping& dm,
     //*********************************************************
     if (solverChoice.rad_type != RadiationType::None)
     {
-        // Allocate with ghost cells to enable interpolation from coarse level
-        // (needed for nested patches where radiation is skipped)
-        // Use same pattern as z_phys: ComputeGhostCells + extra for interpolation
-        int ngrow_rad = ComputeGhostCells(solverChoice) + 2;
-        qheating_rates[lev] = std::make_unique<MultiFab>(ba, dm, 2, ngrow_rad);
+        // Allocate with 1 ghost cell for interpolation stencil (cell_cons_interp)
+        // and FillBoundary operations (needed for nested patches)
+        qheating_rates[lev] = std::make_unique<MultiFab>(ba, dm, 2, 1);
         rad_fluxes[lev]     = std::make_unique<MultiFab>(ba, dm, 4, 0);
         qheating_rates[lev]->setVal(zero);
         rad_fluxes[lev]->setVal(zero);
--- a/Source/ERF_MakeNewArrays.cpp
+++ b/Source/ERF_MakeNewArrays.cpp
@@ -499,7 +499,11 @@ ERF::init_stuff (int lev, const BoxArray& ba, const DistributionMapping& dm,
     //*********************************************************
     if (solverChoice.rad_type != RadiationType::None)
     {
-        qheating_rates[lev] = std::make_unique<MultiFab>(ba, dm, 2, 0);
+        // Allocate with ghost cells to enable interpolation from coarse level
+        // (needed for nested patches where radiation is skipped)
+        // Use same pattern as z_phys: ComputeGhostCells + extra for interpolation
+        int ngrow_rad = ComputeGhostCells(solverChoice) + 2;
+        qheating_rates[lev] = std::make_unique<MultiFab>(ba, dm, 2, ngrow_rad);
         rad_fluxes[lev]     = std::make_unique<MultiFab>(ba, dm, 4, 0);
         qheating_rates[lev]->setVal(zero);
         rad_fluxes[lev]->setVal(zero);
--- a/Source/Initialization/ERF_InitFromWRFInput.cpp
+++ b/Source/Initialization/ERF_InitFromWRFInput.cpp
@@ -8,6 +8,7 @@
 #include "ERF_Utils.H"
 #include "ERF_ProbCommon.H"
 #include "ERF_DataStruct.H"
+#include "ERF_TerrainMetrics.H"
 
 #include "ERF_ReadFromWRFInput.H"
 #include "ERF_ReadFromWRFBdy.H"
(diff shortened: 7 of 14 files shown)


adamwise95:
@AMLattanzi see changes to the radiation code for how I expanded the memory pool when there is vertical refinement and that level expands the entire column. I'm not sure this is the best way, open to alternatives.

asalmgren:
Reviewed this. There's a build blocker, and then a set of real issues in the deferred-LSM and radiation logic.

## Blocker: the new source file isn't in the branch

`Source/Initialization/ERF_InitFromWRFInput_SurfaceOnly.cpp` is added to `CMake/BuildERFExe.cmake:212` and `Source/Initialization/Make.package:21`, declared at `Source/ERF.H:806`, and called at `Source/ERF_MakeNewLevel.cpp:486` — but the file does not exist on the PR head (`git ls-tree` finds nothing). CMake fails at configure time ("Cannot find source file"), GNU make fails on the missing `CEXE_sources` entry, and `ERF::init_from_wrfinput_surface_only` is an undefined symbol. Every `ERF_ENABLE_NETCDF` build is broken; it looks like the file was lost in a rebase. Everything below is review of code that hasn't been compiled as pushed.

## LSM deferral is broken on all three level-creation paths

**`Source/ERF_MakeNewLevel.cpp:514` — fine-level LSM data/flux pointers end up all null.** The deferred block assigns `lsm_data[lev][mvar]`/`lsm_flux[lev][mvar]`, but those inner vectors are only *sized* later, by `make_lsm_at_level(lev)` at line 599 — `ERF_Constructors.cpp:107` resizes only the outer vector. On first creation of level 1 both inner vectors are empty, so both loops iterate zero times. `make_lsm_at_level` then resizes them to `lsm_data_size` (filling with `nullptr`) and, because `will_use_surface_only` is true, skips the pointer assignment. `ERF_Checkpoint.cpp:374` (`lsm_data[lev][ivar]->boxArray()`), `advance_radiation`'s `lsm_input_ptrs`, and `make_SurfaceLayer_at_level` then dereference nulls. `lsm.Init` is also called here before `lsm.Define(lev, solverChoice)` at line 1208 — harmless for NOAHMP today since `Define` is a no-op, but the ordering invariant is inverted.

**`Source/ERF_MakeNewLevel.cpp:1222` — `RemakeLevel` loses LSM entirely.** `will_use_surface_only` is computed inside `make_lsm_at_level` with no knowledge of which caller is running, and `RemakeLevel` (line 1059) has no compensating deferred init. The comment immediately above that call already spells out the consequence: stale pointers surviving the regrid on the old `(ba,dm)` and then indexed with an MFIter over the new grids — an out-of-bounds device read on the level's first step after the regrid.

**`Source/ERF_MakeNewLevel.cpp:105` — the same gate fires in `MakeNewLevelFromScratch`.** A run started from scratch with `wrfinput_d01` + `wrfinput_d02` and `interp_atmos_from_coarse = true` skips both `lsm.Init` and the pointer assignment at level 1, and there is no deferred init on that path. On that path `interp_atmos_from_coarse` is also silently ignored (the fine atmospheric state is read from file), which contradicts the new `Inputs.rst` entry.

**`Source/ERF_MakeNewLevel.cpp:477` — `use_surface_only` ignores `init_type`.** With Metgrid, `init_from_metgrid` reads the full atmospheric state and then the `if (use_surface_only)` block runs `rebuild_base_state_from_wrfinput` + `FillCoarsePatch` on top of it. The docs say the option applies only to `WRFInput`; the gate should say so too.

## Radiation

**`Source/TimeIntegration/ERF_AdvanceRadiation.cpp:103` — rank-dependent branch around MPI collectives; this will hang.** `is_nested_patch` is computed by iterating `MFIter mfi(*cons_in)`, which visits only *rank-local* boxes. A rank owning no boxes on the fine level never enters the loop and keeps `m_is_nested_patch == false` — `ERF_Radiation.H:110` explicitly anticipates ranks with `m_ncol == 0`. That rank then takes the `!is_nested_patch()` branch and calls `qheating_rates[lev]->FillBoundary(...)`, while every other rank takes the nested branch and calls `qheating_rates[lev-1]->FillBoundary(...)` plus `InterpFromCoarseLevel(...)`. Both are collective in AMReX, so the ranks issue mismatched communication. Reachable any time the fine level has fewer boxes than ranks. Needs a `ParallelDescriptor::ReduceBoolOr` over the local result.

**`Source/PhysicsInterfaces/Radiation/ERF_RRTMGP_Interface.cpp:414` — the pool re-init can shrink the pool and then never grow it back.** `rrtmgp_initialize` deliberately sizes the pool from `m_ncol_chunk_requested`, and the comment at `ERF_Radiation.cpp:1134` says why: "so that the pool remains valid even if m_ncol grows after regridding". The new expansion path instead uses the `ncol` argument of `rrtmgp_main`, which is `ncol_c` — the current chunk, bounded by `m_ncol_chunk = min(m_ncol, requested)`. Concrete case: level 0 with the default `rad_ncol_chunk = 1024` and `nlay = 100` gives `nvar*1024*100*ngpt`; a fine level with `nlay = 200` but only 100 local columns re-inits at `nvar*100*200*ngpt`, five times smaller. Because `pool_max_nlay` is now 200, the coarse level's next 1024-column chunk never re-triggers expansion and `alloc_raw` overruns `s_mem`. Size from `pool_nvar * m_ncol_chunk_requested * nlay * ngpt`, or track `max(ncol*nlay)` rather than `nlay` alone.

**`Source/PhysicsInterfaces/Radiation/ERF_Radiation.H:86` — `m_is_nested_patch` is never cleared.** `Init` re-runs on every `MakeNewLevelFromCoarse`/`RemakeLevel` for the same object. The early return sets it true, but the non-nested path never sets it back to false, so a level that stops being a nested patch after a regrid keeps skipping radiation forever. Add `m_is_nested_patch = false;` before the detection loop.

**`Source/TimeIntegration/ERF_AdvanceRadiation.cpp:143` — LSM surface slabs interpolated with the atmospheric geometry and the 3D refinement ratio.** `lsm_fine`/`lsm_coarse` are single-cell-thick k=0 slabs on `m_lsm_geom`, not on `geom[lev]`. The call passes `geom[lev-1]`, `geom[lev]` and `refRatio(lev-1)` — `7 7 2` in this PR's own Test 3. `cell_cons_interp` will compute a z slope for a slab whose z-neighbours do not exist: `tmp_coarse` has 1 ghost, `MultiFab::Copy(..., 0)` copies none, and `FillBoundary` cannot fill k=±1 because no box owns them, so the slope comes from uninitialized memory and contaminates the fine surface fluxes. Either force the z ratio to 1 (or use `piecewise_constant_interp`), or interpolate on the LSM geometry.

## Terrain

**`Source/Initialization/ERF_InitFromWRFInput.cpp:1337` — the Transform blend is silently skipped when `avg_grid_faces_to_nodes` is true.** The new `lev > 0 && terrain_smoothing != 0` block — the mandatory-`transform` abort plus the interp and save of `z_phys_interp` — sits *outside* the `if (!solverChoice.avg_grid_faces_to_nodes)` guard that contains its only consumer, `make_terrain_fitted_coords`. So with `erf.avg_grid_faces_to_nodes = true` the user is forced by the new abort into transform mode and the transform is then never applied; `z_phys_nd[lev]` keeps the raw wrfinput heights with no warning. Either abort on the combination or move the check inside the same guard.

## Minor

- `ERF_RRTMGP_Interface.cpp:417` — `pool_t::finalize()` defaults to `verbose=true` and calls `print_state()` to `std::cout`, so this prints a "MemPoolSingleton used ..." line from every MPI rank. Pass `finalize(false)`.
- `ERF_Radiation.H:120` — the new `alloc_buffers()` in `Init` duplicates the call `set_grids` already makes (`ERF_Radiation.cpp:288`), which fully reassigns every Kokkos View. It allocates the full `ncol x nlay` buffer set at every level creation and regrid, immediately discarded, and moves the O3-VMR-length assert from the first radiation step to level creation.
- `ERF_InitFromWRFInput.cpp:1329` re-parses `terrain_smoothing` with a local `ParmParse` instead of reading it from `solverChoice`.
- `ERF_TerrainMetrics.cpp:589` — the improved abort message says "'interpolate' only works with analytic/ideal terrain", but this branch fires on `FineTerrain::None`; an actual `interpolate` setting aborts at the new site in `init_from_wrfinput` instead, so the message conflates two conditions.

One thing I checked that is *not* a bug: `z_phys_interp` is defined with `nGrowVect()` ghosts but copied with zero ghosts. The Transform kernel reads only `mfi.tilebox()` and does its own `FillBoundary` afterwards, so no ghost of `z_phys_interp` is ever read — the extra ghost allocation is simply unused.

Source/ERF_MakeNewLevel.cpp
```diff
@@ -102,7 +102,10 @@ void ERF::MakeNewLevelFromScratch (int lev, Real time, const BoxArray& ba_in,
     //********************************************************************************************
     // Land Surface Model
     // *******************************************************************************************
-    make_lsm_at_level(lev);
+    // If time > 0, this is a regrid, so pass from_regrid=true to avoid incorrectly
+    // deferring LSM initialization (deferred init only applies during initial level creation)
+    bool is_regrid = (time > 0.0);
```
asalmgren: **HIGH — `is_regrid` is always false here, so LSM is deferred but never initialized for fine levels created at startup.**

`ERF::InitData` calls `InitFromScratch(zero)` (`ERF.cpp:639`), and AmrCore creates *every* initial level — including `lev > 0` — through `MakeNewLevelFromScratch` with that same `time == 0`. So `time > 0.0` is false for the initial creation of level 1, not just for level 0.

With `erf.interp_atmos_from_coarse = true`, `init_type = WRFInput`, and `nc_init_file[1]` non-empty, `make_lsm_at_level(1, /*from_regrid=*/false)` therefore computes `will_use_surface_only = true` (line 1252) and skips both `lsm.Init` and the `lsm_data[lev]` / `lsm_data_name` pointer setup.

But this function does not take the surface-only path — it only prints the new warning and then calls `init_only` -> `init_from_wrfinput`, which at `ERF_InitFromWRFInput.cpp:1137-1139` does:

```cpp
int lsm_idx = lsm.Get_DataIdx(lev, var.second);
AMREX_ALWAYS_ASSERT_WITH_MESSAGE(lsm_idx != -1, "LSM variable mapping invalid!");
AMREX_ALWAYS_ASSERT(lsm_data[lev][lsm_idx]);
```

`NOAHMP::Lsm_DataIndex` loops `idx` over `[0, m_lsm_data_size)` reading `LsmDataName[idx]`, but `LsmDataName` is sized and populated only in `NOAHMP::Init` (`ERF_NOAHMP_Init.cpp:55-64`) — which was just skipped — while `m_lsm_data_size` is already non-zero (set via `Get_Data_Size()` -> `m_ensure_nsoil_resolved()`). So this reads past the end of an empty `Vector`, and the `lsm_data[lev]` entries are still the nullptrs left by `resize`. Best case it aborts on "LSM variable mapping invalid!"; worst case it is an out-of-bounds read.

This is exactly the configuration the PR describes as Test 1 (2-level WRF + Noah-MP + `interp_atmos_from_coarse`). The deferred init only ever runs in `MakeNewLevelFromCoarse`, so nothing recovers it. Since the warning at line 163 already says the flag is ignored in this function, `make_lsm_at_level` should not defer here at all.

Source/Radiation/RRTMGP/ERF_RRTMGP_Interface.cpp
```diff
@@ -403,6 +409,20 @@ rrtmgp_main (const int ncol, const int nlay,
...
+    // This happens with vertical refinement where fine levels have more vertical cells
+    if (static_cast<size_t>(nlay) > pool_max_nlay) {
+
+        // Finalize old pool
+        pool_t::finalize();
+
+        // Reinitialize with new larger size using same nvar from initialization
+        const size_t ngpt = std::max(k_dist_sw_k->get_ngpt(), k_dist_lw_k->get_ngpt());
+        auto my_size_ref = static_cast<unsigned long>(pool_nvar * ncol * nlay * ngpt);
+        pool_t::init(my_size_ref);
```
asalmgren, line 427: **HIGH — this can *shrink* the pool, permanently, leading to a silent heap overwrite.**

The re-init uses the *current chunk's* `ncol`, and there is no `pool_max_ncol` counterpart to `pool_max_nlay`. `initialize_impl` deliberately sizes the pool with `m_ncol_chunk_requested` (default 1024) so it stays valid as `m_ncol` changes (`ERF_Radiation.cpp:1189-1196`), but `rrtmgp_main` is handed `ncol_c = min(m_ncol_chunk, m_ncol - col_s)`, which is `<= m_ncol_chunk_requested`.

Scenario: level 0 has `nlay = 100`, pool sized `nvar*1024*100*ngpt`. Level 1 has `nlay = 200` but only 500 columns on this rank, so `ncol_c = 500`, the branch fires, and the pool is re-inited to `nvar*500*200*ngpt` — **smaller than before**. `pool_max_nlay` is then bumped to 200, so the guard never fires again, and the next level-0 chunk (1024 cols x 100 layers) overruns `s_mem`.

The only bounds check is in `MemPoolSingleton::alloc_raw` (`Submodules/RRTMGP/cpp/rrtmgp_conversion.h`):

```cpp
s_curr_used += num_reals;
assert(s_curr_used <= s_mem.size());
```

a plain `assert`, compiled out under `NDEBUG` — so in an optimized build this is a silent heap overwrite, not an abort.

Suggested fix: track `pool_max_ncol` alongside `pool_max_nlay`, seed both in `rrtmgp_initialize` (with `m_ncol_chunk_requested` for the column count), and size the new pool with `max(pool_max_ncol, ncol) * max(pool_max_nlay, nlay)`.

Minor, same block: `pool_t::finalize()` declares `finalize(bool verbose=true)` and `print_state()` writes to `std::cout`, so every re-init prints `rrtmgp_conversion MemPoolSingleton used ...` from every MPI rank. Pass `false`.

Source/ERF_MakeNewLevel.cpp
```diff
@@ -467,17 +483,80 @@ ERF::MakeNewLevelFromCoarse (int lev, Real time, const BoxArray& ba,
...
+        // If we used surface-only init, we need to rebuild the base state and
+        // interpolate the atmospheric state from coarse (just like a level with no init file)
+        //
+        if (use_surface_only) {
+            // Interpolate atmospheric state from coarse level.
+            // NOTE: This creates thermodynamically inconsistent data because ρ, θ, qv are
+            // interpolated independently. We rely on:
+            // 1. Skipping radiation on the first timestep (see ERF_AdvanceRadiation.cpp)
+            // 2. Letting dynamics equilibrate the state on the first timestep
+            FillCoarsePatch(lev, time);
```
asalmgren, line 550: **MEDIUM — the base state is not rebuilt, contrary to the comment above.**

The comment at line 515 says "we need to rebuild the base state and interpolate the atmospheric state from coarse (just like a level with no init file)", but only `FillCoarsePatch` is called.

The no-init-file path this is modelled on does more: line 579 is preceded by `rebuild_base_state_from_wrfinput(lev, base_state[lev])`, and that ordering is deliberate — `FillCoarsePatch` interpolates *perturbational* quantities relative to the base state. `init_from_wrfinput` does the equivalent internally (`ERF_InitFromWRFInput.cpp:1669`). `init_from_wrfinput_surface_only` does neither.

So `base_state[lev]` still holds the plain coarse cell-centred interpolation done at line 427 — computed *before* the fine terrain and `z_phys_cc` existed. The fine level's base state is therefore inconsistent with its own mesh, and every perturbational quantity `FillCoarsePatch` produces is referenced to the wrong hydrostatic state. A `rebuild_base_state_from_wrfinput(lev, base_state[lev])` call before the `FillCoarsePatch` here would match the sibling path.

Source/Initialization/ERF_InitFromWRFInput_SurfaceOnly.cpp
```diff
@@ -0,0 +1,740 @@
...
+
+            // LSM variables
+            if (use_lsm) {
+                auto &lsm_wrfmap = lsm.Get_WRFInputNames();
+                for (auto &var : lsm_wrfmap) {
+                    if (var_name == var.first) {
+                        bool is_3d = var_fab.box().length(2) > 1;
+                        amrex::Print() << "   Reading " << ((is_3d) ? "3D" : "2D") << " LSM variable '"
+                                      << var.first << "' (" << var.second << ")" << std::endl;
+                        int lsm_idx = lsm.Get_DataIdx(lev, var.second);
```
asalmgren, line 510: **MEDIUM — LSM variables are read into `lsm_data[lev]` before the LSM exists on that level.**

`init_from_wrfinput_surface_only` is only ever called from `MakeNewLevelFromCoarse` (line 505), which does not call `lsm.Define`/`lsm.Init` for this level until line 537, and does not reach `make_lsm_at_level` until line 629.

At this point `lsm_data[lev]` is still empty and `m_lsm_model[lev]->LsmDataName` is still empty (it is populated in `NOAHMP::Init`, `ERF_NOAHMP_Init.cpp:55-64`), while `m_lsm_data_size` is non-zero. So `lsm.Get_DataIdx(lev, ...)` indexes `LsmDataName[0..m_lsm_data_size)` out of bounds, and `lsm_data[lev][lsm_idx]` indexes an empty `Vector` — the same failure mode as in `MakeNewLevelFromScratch`.

Even if it survived, `NOAHMP::Init` at line 537 reallocates every `lsm_fab_data[ivar]` and `setVal(lsm_undefined)`, so everything read here is discarded. Either drop the LSM reads from the surface-only reader (Noah-MP reads the file itself in `Init`), or move the LSM define/init ahead of this call.

Source/Radiation/RRTMGP/ERF_Radiation.H
```diff
@@ -66,10 +66,34 @@ public:
...
         int klo = geom.Domain().smallEnd(2);
         int khi = geom.Domain().bigEnd(2);
 
+        // Detect if any box doesn't reach the domain top (nested patch)
+        // Must check across all ranks since some ranks may have boxes that reach the top
+        // while others don't. If ranks disagree, collective communication below will deadlock.
+        bool is_nested_patch_local = false;
+        for (amrex::MFIter mfi(*cons_in); mfi.isValid(); ++mfi) {
+            const amrex::Box& vbx = mfi.validbox();
+            if (vbx.bigEnd(2) < khi) {
```
asalmgren: **MEDIUM — vertical box decomposition is now silently mistaken for a nested patch.**

The detection here is "any valid box whose `bigEnd(2) < khi`", which is also true for a level split in z by `amr.max_grid_size_z` — including level 0. That case previously hit the `AMREX_ALWAYS_ASSERT_WITH_MESSAGE(..., "Vertical decomposition with radiation is not allowed.")` at line 109; now `Init` returns early and `Run` returns early, so radiation is simply skipped.

For `lev > 0` the heating rates at least come from the coarse level, but at lev 0 there is no fallback — `ERF_AdvanceRadiation.cpp:166` requires `lev > 0` — so `qheating_rates[0]` stays at the `setVal(zero)` from `init_stuff` for the entire run. A user who previously got a hard abort now gets a silently radiation-free simulation.

The check should distinguish "box top < domain top" (a nested patch) from "box bottom > domain bottom" (a z-split): key it on the level's `ba.minimalBox()` not reaching `khi`, and keep aborting on a genuine vertical decomposition.

Source/TimeIntegration/ERF_AdvanceRadiation.cpp
```diff
@@ -59,6 +59,63 @@ void ERF::advance_radiation (int lev,
 
     if (solverChoice.rad_uses_interface()) {
         BL_PROFILE_VAR("ERF::advance_radiation():RRTMGP", rrtmgp_region);
+
+        // On the first timestep after a new level is created from coarse using interp_atmos_from_coarse,
+        // interpolate heating rates and radiation fluxes from coarse instead of computing them.
+        // This avoids NaNs from thermodynamic inconsistencies in the FillCoarsePatch atmospheric
+        // state interpolation, and provides the LSM with the radiation fields it needs.
+        if (solverChoice.interp_atmos_from_coarse && lev > 0 && istep[lev] == 0) {
```
asalmgren: **LOW/MEDIUM — this fires on levels that never used `FillCoarsePatch`.**

The gate is `interp_atmos_from_coarse && lev > 0 && istep[lev] == 0`, with no check that this level actually took the surface-only path. Fine levels created at startup go through `MakeNewLevelFromScratch`, which reads their full atmospheric state from `wrfinput_d02` and explicitly warns (`ERF_MakeNewLevel.cpp:163`) that the flag is ignored there — yet their first radiation call is still replaced by coarse-interpolated heating rates.

The branch also returns without touching `rad_fluxes[lev]`, which therefore stays zero for that step even though the LSM and plotfiles consume it; only the `lsm_output` fields are interpolated.

Consider recording whether the level was initialized surface-only (a per-level flag set in `MakeNewLevelFromCoarse`) and gating on that instead of on `istep[lev] == 0`.

Source/Initialization/ERF_InitFromWRFInput_SurfaceOnly.cpp
```diff
@@ -0,0 +1,740 @@
...
+
+        // Initialize terrain
+        ParmParse pp("erf");
+        int terrain_smoothing = 0;
+        pp.query("terrain_smoothing", terrain_smoothing);
+
+        FineTerrain fine_terrain = FineTerrain::None;
+        MultiFab z_phys_interp;
+
+        if (lev > 0 && terrain_smoothing != 0) {
```
asalmgren: **LOW — guard differs from the sibling path.**

The equivalent block in `ERF_InitFromWRFInput.cpp:1337` is guarded by `&& !solverChoice.avg_grid_faces_to_nodes`; here it is not.

With `erf.avg_grid_faces_to_nodes = true` and `terrain_smoothing != 0`, the surface-only path aborts demanding `amr_terrain_refinement = transform` — or, if `transform` is set, performs an `InterpFromCoarseLevel` plus a full-size `MultiFab` copy whose result is then unused, because `make_terrain_fitted_coords` is skipped at line 684. The regular path runs fine in the same configuration. The two should use the same condition.

asalmgren (commented):
Automated code review of the `interp_atmos_from_coarse` path. Seven findings, all inline below — two high (a guaranteed failure in the configuration the PR advertises as Test 1, and a silent heap overwrite in the RRTMGP pool), three medium, two low.

<details>
<summary>Checked and cleared</summary>

- The `std::string` argument to `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` compiles (AMReX has a `const std::string&` overload of `Assert`).
- The `m_ncol_chunk` -> `m_ncol_chunk_requested` change in `alloc_buffers` only over-allocates; the full-view use in `finalize_impl` is on the `datalog_int > 0` branch, which still uses `m_ncol`.
- The pool `finalize`/`init` is not a use-after-free: `gas_concs_c` uses plain Kokkos allocation, not the pool.
- The forward declarations in the new `ERF_InitFromWRFInput_SurfaceOnly.cpp` match the definitions in `ERF_InitFromWRFInput.cpp`.
- `istep[lev]` really is 0 for a level created mid-run.
- `Docs/sphinx_doc/Inputs.rst` does document the new `erf.interp_atmos_from_coarse` parameter.

</details>

asalmgren:
I reviewed this and pushed one commit (0a1b3532b) directly to the branch, since the mechanical fixes were easier to show than to describe. Please pull before your next local commit. Net effect is 927 lines removed, 251 added; all 117 regression tests pass unchanged, so no gold files needed regenerating.

Thanks for this — the nested-patch detection and the base-state-parameter consistency work are nice, and the `m_ncol_chunk_requested` sizing fix is a real bug caught. What follows is all in the surrounding plumbing.

### What I pushed

**Folded `ERF_InitFromWRFInput_SurfaceOnly.cpp` back into `init_from_wrfinput` behind a `read_atmos_state` flag.**

About 600 of that file's 740 lines were a verbatim copy of `init_from_wrfinput`'s read loop, subdomain/`refine_in_z` logic, `XLAT_V`/`XLONG_U` staggering fix-ups, map-factor handling, and the terrain/Newton-stretching block (lines 652-733 duplicated 1327-1417). The five helper prototypes were also hand-re-declared rather than shared through a header, so a signature change would have compiled into a silent ODR mismatch.

The copy had already drifted in a way that mattered. It replaced the `"failed to read required variable"` abort with `Print() << " not found (skipping)"; continue;`, but left `compute_terrain_here` at `true`. In `init_from_wrfinput` a missing `PH` or `PHB` sets that flag false, which is what guards the only dereference of `mf_PHB`; without it, a wrfinput file missing `PHB` dereferenced an uninitialized `MultiFab*` at lines 646 and 681 rather than aborting with a diagnosable message. Reading through the one function restores the guard.

It also drops four NetCDF reads that could never be used: for `lev > 0` — the only case the function was ever called for — `mf_MUB`, `mf_C1H`, `mf_C2H` and `mf_RDNW` pointed at `MUB_tmp` / `C1H_tmp` / `C2H_tmp` / `RDNW_tmp`, function-local temporaries destroyed at the closing brace without ever being read. `rebuild_base_state_from_wrfinput` uses `wrf_bsp` and `z_phys_cc`, not those.

**Collapsed the two interpolation blocks in `ERF_AdvanceRadiation.cpp` into one helper, and fixed the top-of-atmosphere plane.**

The `istep == 0` block and the nested-patch block were near-identical. Both copied the coarse MultiFabs into ghosted temporaries and called `FillBoundary` on them before interpolating — but `InterpFromCoarseLevel` builds its own coarse patch and fills it with `ParallelCopy(cmf, ..., send_ghost = 0, recv_ghost = 0, ...)` (`AMReX_FillPatchUtil_I.H:1112`), i.e. from the source's *valid* region only. The temporaries therefore contributed nothing to the result while costing an allocation, a copy and a round of communication per field per step — on a nested level, one full coarse MultiFab per LSM output variable (~10) plus `rad_fluxes`, every step. They now interpolate from the coarse MultiFabs directly.

That exposed a real gap. RRTMGP's layout keeps the TOA interface fluxes in the z-ghost cell of `rad_fluxes` at `k = khi+1` (see the comment at `ERF_MakeNewArrays.cpp:521-524`), which is physical data rather than a halo, and `ERF_Plotfile2D.cpp:409-412` reads it for 2D output. It was invisible to the interpolation from both ends: the coarse TOA sits outside the valid region the `ParallelCopy` reads, and on a fine level spanning the full column the destination plane sits outside the fine domain, since `nghost_outside_domain` was `IntVect(0,0,0)`. So the fine TOA kept its stale or zero value. Both planes are now shifted into valid index space at the coarse domain top and interpolated horizontally with `IntVect(rx, ry, 1)`; `CellConservativeLinear` zeroes the slope in any direction whose ratio is one (`AMReX_MFInterp_3D_C.H:113-160`), so nothing reads across the one-cell-thick slab. A nested patch does not reach the model top, so its top plane is an ordinary interior interface the main call already fills, and the new step is skipped there.

**Dropped the `alloc_buffers()` call added to `Radiation::Init`.** This is the one change I am least sure about, so please push back if it was load-bearing. `set_grids()` already calls `alloc_buffers()` at the top of every `Run()` that updates radiation (`ERF_Radiation.cpp:313`) and `finalize_impl()` frees them again (`ERF_Radiation.cpp:1525`), which is what keeps one level's worth of RRTMGP working memory resident at a time. With the Init-time call, every level's buffers were live from the moment that level was created until its *first* `Run`, so a 3-level run held all three sets at startup — the peak `RRTMGP_Memory_Reduction.md` exists to avoid. If you added it chasing a specific GPU failure, I would rather find that cause than pre-allocate; the `m_ncol_chunk_requested` sizing change in the same commit already addresses the shrink-after-regrid problem on its own.

**Padded the `erf.amr_terrain_refinement` row in `Inputs.rst`.** It was 134 characters where the rest of the grid table is 135, which Sphinx rejects with "Malformed table. Text in column margin."

### What I did not touch — see #4060

Three things need a decision from you rather than a patch, so I opened #4060 instead of guessing. The first is worth reading before this merges:

**The soil and vegetation fields are not actually being read.** The description says `wrfinput_d02` supplies TSLB, SMOIS, SH2O, LAI, ZS, DZS, VEGFRA, TMN, SHDMIN and SHDMAX, but the block that reads them is unreachable: `Lsm_WRFInputNames()` returns `wrfinput_map`, and nothing anywhere in `Source/` ever writes to that map. Separately, the pre-read `lsm.Define()` added in `MakeNewLevelFromCoarse` allocates nothing — it is an empty function in both Noah-MP (`ERF_NOAHMP.H:70`) and SLM (`ERF_SLM.H:38`), and the structures the read indexes (`LsmDataMap`, `LsmDataName`, `lsm_fab_data`) are sized in `NOAHMP::Init`, which this path defers until after `FillCoarsePatch`. So if the map were ever populated, `Get_DataIdx()` would index an empty vector and write through a null `lsm_data[lev][lsm_idx]`. Either populate the map and move `lsm.Init` ahead of the read, or drop the block and the LSM names from the variable list.

The other two: `istep[lev] == 0` is not a reliable "first step after creation" test, because `istep` is never reset in `ClearLevel` — a level that is de-refined and re-created carries its old count, so the guard silently lapses and RRTMGP runs on exactly the inconsistent `FillCoarsePatch` state that `ERF_MakeNewLevel.cpp:545-549` warns produces NaNs. And I could not work out how Tests 1-3 reached the new code at all: at startup every level goes through `MakeNewLevelFromScratch`, where `interp_atmos_from_coarse` is explicitly ignored with a warning (`ERF_MakeNewLevel.cpp:164-175`), so a 2-level run from `wrfinput_d01`/`d02` should take the ignore path. Could you say how the surface-only path was exercised? A regression test covering it would be worth adding, since none of the current 117 reach it.

PR #4017, opened 2026-09-15
TKE buoyancy production: average the full face fluxes of the theta diffusion

hgopalan:
## The bug

The buoyancy production of k in the Deardorff and one-equation k closures is `g/theta_0 * hfx_z`, read in `ERF_AddTKESources.H`. Three things were wrong with it:

- **It read a face value as a cell value.** `hfx_z` (`Hfx3`) is z-nodal and holds face fluxes, but the source read `hfx_z(i,j,k)`, the lower face of the cell.
- **With the implicit vertical solve it lost the term.** Above the first cell, the theta diffusion overwrites `hfx_z` with its face flux times the explicit fraction (`DiffusionSrcForState_{N,S,T}`). With the implicit vertical solve (`vert_implicit_fac > 0`, the default for compressible runs) that fraction is zero on the implicit stages, so buoyancy production and destruction of k vanished above the first cell.
- **The outputs could show the same scaled flux.** The hfx3 plot variable, the 1D SGS heat-flux profiles and line sampling all report `Hfx3`. With the default `erf.vert_implicit_fac = 1 1 0` the last stage is explicit, so the output was already the full flux. With a setting whose final stage is implicit (for example `1 1 1`), hfx3 was zero above the surface. The same held for `q1fx3` under `erf.implicit_moisture_diffusion`.

## The fix

This follows the approach proposed in review:

1. **Full face flux.** `DiffusionSrcForState_{N,S,T}` store the full face flux, `hfx_z(i,j,k) = zflux(i,j,k)` and `qfx1_z(i,j,k) = zflux(i,j,k)`, at the twelve write sites of each.
   - Unchanged: `zflux *= explicit_fac` (the RHS and the reflux register). The EB path needs nothing, since its `hfx_z`/`qfx1_z` writes are commented out.
   - Above the surface, `qfx1_z` is read only by diagnostics: the q1fx3 plot variable, the SGS profiles and line sampling. So hfx3 and q1fx3 now agree for every `vert_implicit_fac`.
2. **Face average in the k source.** `ERF_AddTKESources.H` uses `0.5 * (hfx_z(k) + hfx_z(k+1))` in every cell, the first included.
   - With a surface layer, the bottom face holds the surface-layer flux, so the first cell averages it with the diffusion flux at z = dz.
   - The k source runs on tiles of the boxes whose theta diffusion wrote both faces (`TileNoZ`), so the read at `k+1` is always written. The header comment names `erf_slow_rhs_pre` as the required same-stage writer, and an `AMREX_ASSERT` checks that the faces of `bx` lie in `hfx_z`.
   - No new `EddyDiff` component or other array.
3. **Closures store no heat flux.** The Smagorinsky, Deardorff and k-eqn closures no longer write `hfx_x/hfx_y/hfx_z`: they wrote a cell-centred value into a z-face array, and the theta diffusion overwrote every face before anything read it. `ComputeTurbulentViscosityLES` and `ComputeTurbulentViscosityRANS` drop their `Hfx1/Hfx2/Hfx3` arguments.
   - The EB closure keeps its write: `DiffusionSrcForState_EB` does not write `hfx_z`, so on EB terrain it is the only writer of the heat flux output.
4. **Deardorff/k-eqn aborts.**
   - Deardorff with any PBL scheme already aborted. `erf.rans_type = kEqn` now also aborts with a PBL scheme that transports `RhoKE`: `MYJ`, `MYNN25`, `MYNNEDMF` and the SHOC schemes.
   - `erf.terrain_type = EB` now aborts with Deardorff or k-eqn. The EB diffusion path does not include the TKE sources, so such a run integrated `RhoKE` with no buoyancy, shear or dissipation.
   - No deck or CTest uses any of these pairs.
5. **Docs.**
   - The RANS and DNSvsLES theory pages describe the face average and the closure/PBL rule, and state that on a terrain-fitted mesh the stored face flux is the component along ζ only (the slope corrections and the Jacobian are applied where the fluxes enter the RHS), so the source and the hfx3 output omit those terms over sloping terrain. That omission is pre-existing.
   - `Discretizations.rst` gives the discretised buoyancy term as the face average.
   - The plotfile reference describes `hfx3` and `q1fx3` as the full face fluxes.

Other readers of `Hfx3` are unaffected, because they read only the domain bottom, which still holds the surface flux: the implicit solves, SHOC, REMORA and the LSM. The checkpoint stores `Hfx3`, but it is rewritten on every stage.

## Test

`check_implicit_explicit_ke.py` compares two runs of `Canonical_RANS/Convective_ABL_Flat`, compressible for 40 steps: one with the implicit vertical solve and one with explicit vertical diffusion. Both runs now use the full face flux, so they differ only by the time discretisation. The check requires the largest planar-mean KE difference, relative to the largest KE, to stay below a tolerance. The script also checks that both runs are finite, that turbulence is present, and that the two runs do differ.

The CTest entries are `RANS_Convective_ABL_Flat_Buoyancy_kEqn` (tolerance 1e-4) and `RANS_Convective_ABL_Flat_Buoyancy_Deardorff` (3e-4). They are registered by `add_test_rans_pair`, which, like `add_test_rans`, runs with or without MPI. They are skipped on Windows because the command runs through `sh`.

**Box boundaries in z.** `RANS_Convective_Buoyancy_ZSplit_BoxParity_kEqn` and `_Deardorff` (`add_test_box_parity`, `Tests/test_files/RANS_Convective_Buoyancy_ZSplit`) cover the read at `k+1` across z box boundaries.
- **Setup:** the same case, compressible with explicit diffusion, run on one box and on 10 boxes (x split at 4; `max_grid_size_z = 24` gives five boxes of 20 cells, so the z splits are at k = 20, 40, 60, 80, that is z = 400, 800, 1200 and 1600 m, two below the 937 m inversion and two above), with `rel_tol 1e-9`.
- **Results:**
  - The branch gives identical plotfiles.
  - A mutant that skips the `hfx_z` write on each box's top face fails: KE 1.1e-6 / 3.4e-6.
  - A mutant that reads the lower face in each box's top cell fails: KE 2.0e-9 / 6.4e-9.
- **Acoustic substepping is off in this deck:** the implicit `w` solve is closed at every box's lo/hi (since 5bbcef32e, #1762; abort in #4042). The PBL-height length cap is on, since #4044 made the PBL height scan whole columns on z-split grids.

**Implicit against explicit.** Each binary runs both closures, compressible, 40 steps, 2 ranks. The table gives the largest relative KE difference:

| binary | kEqn (tol 1e-4) | Deardorff (tol 3e-4) |
| --- | --- | --- |
| development (42934d887) | 7.10e-4, fails | 2.13e-3, fails |
| this branch, `hfx_z = zflux * explicit_fac` restored | 5.30e-4, fails | 1.88e-3, fails |
| this branch, source reads the lower face only | 1.15e-5, passes | 4.87e-5, passes |
| this branch | 1.12e-5, passes | 5.27e-5, passes |

- **The explicit-fraction mutant fails**, so the test guards the implicit/explicit dependence.
- **Same run twice fails:** giving the checker the same run twice fails it (theta difference 0).
- **Where the flux is taken:** the regression pair cannot see it, because both runs share the location. The unit test below covers it.

**Unit tests** (`Tests/Unit/Diffusion/ERF_GTestTKEBuoyancySource.cpp`) call `DiffusionSrcForState_N` on one column with θ = a + bz + cz², qv linear, and a constant eddy diffusivity. The stencil is exact for these, so every value has a closed form:
- **`FaceFluxesAreFullForExplicitAndImplicitDiffusion`** (`implicit_fac` 0, 0.5, 1): `hfx_z` and `qfx1_z` hold the full face flux, and the k source equals |g|/θ_ref · (−K b). It fails when `hfx_z` or `qfx1_z` keeps the explicit fraction.
- **`CellCentredFluxIsTheAverageOfTheTwoFaces`** (c = 0.1): the source equals |g|/θ_ref · (−K(b + 2c·z_k)). It fails when the source reads only the lower face. The test asserts the lower-face error is at least four tolerances, so it stays resolvable in single precision.
- **`FirstCellAveragesTheSurfaceFluxWithTheFaceAbove`:** a zlo surface layer with a surface flux S on the bottom face; the first cell's source must equal |g|/θ_ref · ½(S − K(b + 2c·dz)). It fails when the first cell uses the surface flux alone.
- All three check every face and cell, the first and last included, and each asserts that the quantity it checks exceeds four tolerances, so none can pass by being smaller than the roundoff floor of the theta differences (eps K a / dz, 7e-3 in single precision).

**Checker command line:** `RANS_Checks_SelfTest` now also calls `check_implicit_explicit_ke.py` with malformed arguments; each call must print the usage and exit with status 2. With the previous parsing, three of the five calls end in a traceback.

## Verification

Rerun in full on f56125ce5, which also merges current `development` (9929d94a9; the `Tests/CTestList.cmake` conflict with #4043's `DATALOG` argument is resolved by keeping both).

| check | result |
| --- | --- |
| Release build (MPI, FFT, all warnings) | no warnings |
| Release `ctest -L regression` | 124 of 127 pass; the three reds (`ABL_MOST_WOA_ZSplit`, `ABL_MOST_Cloudchamber` gold differences, `Closure_BoxParity_MYNNEDMF` FPE in Release) fail identically on development on this machine and do not touch the changed code |
| Release unit / parallel tests | 575/575, 47/47 |
| Debug, MPI, no FFT (CI configuration) | no warnings; 16/16 of the RANS, buoyancy, z-split, closure box-parity and Deardorff entries; unit tests 575/575 with `AMREX_DEBUG` bounds checks |
| MPI off | configures; the pair tests register |
| Mutants | explicit fraction restored, lower-face read and first-cell special case each fail their unit test; the explicit-fraction mutant fails the regression pair |
| Aborts | k-eqn with MYJ/MYNN25/MYNNEDMF/SHOC, and EB with Deardorff or k-eqn, abort; development ran both combinations |
| Single precision (`ERF_PRECISION=SINGLE`) unit tests | the three TKE tests pass; with `hfx_z = zflux * explicit_fac` restored all three fail (before this round the first one passed). 32 unrelated single-precision failures (two-stream radiation, Kessler, WENO) are in files this PR does not touch |
| Single precision, `-Wshadow` | `ERF_ComputeTurbulentViscosity.cpp`, `ERF_DiffusionSrcForState_{N,T}.cpp`, the unit test: no warnings |
| Device code | GPU lint, private-lambda and deep scans: nothing new (one known `KAPPA` false positive on an untouched line); device constants in the unit test are `constexpr`; CUDA build not run here |
| MSVC / GNU Make | no `long`, `M_PI`, `gmtime`; no new source files |
| Docs | Sphinx builds with no warnings |
| Style | codespell, tab and trailing-whitespace scripts, `git diff --check`: clean |


AMLattanzi:
# Review

This current PR identifies a true bug and proposes a viable option. However, I would recommend a different approach that fixes the current issue, and IO bug, and also does not introduce more memory load. The description is given below.

## Notes
1. `hfxz` is nodal in the z-direction `(0,0,1)` and lives at w-faces. So using it as the CC value at `klo` is not formally correct.
2. Lingering error was that we always used `hfxz` for the TKE source and it lives in the wrong place.
3. Lingering error was also that `hfxz` would be output to the datalog files with the implicit factor being multiplied.

## Proposed Alternative
1. Drop explicit_fac from the hfx_z write

Twelve sites, four in each of the three variants (the four l_consA/l_turb branch combinations):

ERF_DiffusionSrcForState_N.cpp: lines 224, 361, 495, 626
ERF_DiffusionSrcForState_S.cpp: lines 228, 363, 495, 624
ERF_DiffusionSrcForState_T.cpp: lines 267, 419, 568, 715

Each becomes hfx_z(i,j,k) = zflux(i,j,k);.

Two things to leave alone. The zflux(i,j,k) *= explicit_fac at N:644, S:652, T:734 is the diffusive flux register used for refluxing and is load-bearing. The adjacent qfx1_z writes have the same diagnostic problem under implicit_moisture_diffusion, but they're a separate flag and a separate argument; keep them out of this PR.

ERF_DiffusionSrcForState_EB.cpp needs nothing, its hfx_z write is commented out at line 229 and the EB path asserts Smagorinsky anyway.

2. Average the hfxz fluxes at the z-faces to CC in ERF_AddTKESources.H

```
    // hfx_z holds the face fluxes of the theta diffusion.  Average the two faces
    // of this cell for the cell-centred vertical heat flux; with a surface layer
    // the first cell keeps the surface flux at its bottom face.
    // NOTE: reads hfx_z at k+1, i.e. one face above bx; Hfx3 is z-nodal with a
    //       ghost in each direction so the top face of the last cell is valid.
    const bool l_surf_cell = l_surf_flux_klo && (k == l_klo_dom);
    const amrex::Real hfx_cc = (l_surf_cell)
                             ? hfx_z(i,j,k)
                             : amrex::Real(0.5) * ( hfx_z(i,j,k) + hfx_z(i,j,k+1) );
    amrex::Real buoyancy_src = l_abs_g * l_inv_theta0 * hfx_cc;
```

SurfLayer_zlo comes from ERF_SetupDiff.H:10 and is already in scope in all three files; drop its [[maybe_unused]] once it's read here. The comment block at the top of the header needs rewriting, since the premise changes from "the closure's value is overwritten" to "the overwrite is now the thing we want."

3. Revert the rest of the PR commit
ERF_IndexDefines.H: remove EddyDiff::Theta_flux_v.
ERF_ComputeTurbulentViscosity.cpp: revert both hunks (LES Deardorff at ~283, RANS kEqn at ~723) back to the single hfx_z write. Optionally add a note that this write is superseded by the theta diffusion above the surface.

No new EddyDiff component means no extra cell-centred field with 2 ghosts allocated for every run, and nothing to think about for ShocDriver.cpp:352, the NumDiffs immersed-cell loop at AdvanceDycore.cpp:661, or the ghost extrapolation.

4. Docs

The two theory hunks (DNSvsLES.rst, RANS.rst) need rewording. The claim is no longer "evaluated at the cell centre with the diffusivity and gradient of the start of the step." It's now the average of the two face fluxes of the theta diffusion, which is metric-correct on stretched and terrain-fitted meshes and no longer depends on the implicit/explicit split.

Also worth a line in the plotfile reference: hfx3 now reports the full SGS heat flux rather than the explicit fraction, so implicit runs stop showing near-zero SGS flux profiles.

5. Tests and verification

check_implicit_explicit_ke.py and the two CTest entries carry over unchanged, and should pass with tighter margins since the flux is now identical between the two runs by construction rather than approximately so.

6. Check for Deardorff and PBL

Both models may transport `TKE` so we need to abort if a user selects Deardorff with a PBL model that does carry `TKE`. Please add this to the PR if it is not already in place.

hgopalan:
Thanks, @AMLattanzi. I've taken the proposed approach in d21c99756, with docs in 83c99b076 and be9b8f2b5. The branch also merges the rewritten `development` (4c538aad7); the four conflicts were in files this PR does not touch, and I kept development's version of each.

1. **Drop `explicit_fac` from the `hfx_z` write.** Done at all twelve sites in `DiffusionSrcForState_{N,S,T}`. The `zflux *= explicit_fac` for the RHS and refluxing, the `qfx1_z` writes, and the EB file are unchanged.
2. **Average the faces in `ERF_AddTKESources.H`.** Done as proposed, and the header comment is rewritten.
   - The first cell uses the surface flux when `SurfLayer_zlo` is set.
   - `[[maybe_unused]]` stays on `SurfLayer_zlo` in `ERF_SetupDiff.H`, because `ERF_DiffusionSrcForState_EB.cpp` includes that header and never reads it.
   - The `k+1` read is safe because the TKE source runs on `TileNoZ` tiles of the boxes whose theta diffusion wrote both faces.
3. **Revert the rest.** `EddyDiff::Theta_flux_v` is removed, and both closure hunks are back to the single `hfx_z` write, with a note that the theta diffusion supersedes it above the surface.
4. **Docs.**
   - The RANS and DNSvsLES theory paragraphs describe the face average, which carries the metric terms and does not depend on the implicit/explicit split.
   - The `hfx3` row in the plotfile reference now describes the full face flux.
   - The discretised TKE equation in `Discretizations.rst` showed the cell-centre gradient; it now shows the face average.
5. **Tests.** The checker and both CTest entries carry over. The margins are only slightly tighter: kEqn 1.30e-5 → 1.12e-5, Deardorff 5.65e-5 → 5.05e-5. The two runs still differ because theta itself is advanced differently, so the face fluxes are not identical.
   - Development and a mutant that restores `* explicit_fac` both fail.
   - A mutant that reads only the lower face (`hfx_z(k)`) **passes**, so this test does not pin the location. The location does matter: it changes KE by about 1 % (kEqn) and 2 % (Deardorff) in 40 steps. If you want that covered, I can add a gold-file test.
6. **Deardorff/k-eqn with PBL.** Deardorff with any PBL scheme already aborts in `TurbChoice::init_params`. The k-eqn RANS closure had no check, so `erf.rans_type = kEqn` now aborts with `MYJ`, `MYNN25`, `MYNNEDMF` or the SHOC schemes, naming both inputs. I confirmed the abort fires for all six. MRF and no PBL still run.

The description is updated with the numbers and the verification: Release and Debug no-FFT builds, regression, unit and parallel tests, MPI-off, single precision, `-Wshadow`, and Sphinx. The two regression failures there, `ABL_MOST_WOA_ZSplit` and `ABL_MOST_Cloudchamber`, reproduce identically with the development binary on this machine.

Source/Diffusion/ERF_DiffusionSrcForState_N.cpp
```diff
@@ -221,7 +221,7 @@ DiffusionSrcForState_N (const Box& bx, const Box& domain,
 
             if (qty_index == RhoTheta_comp) {
                 if (!(SurfLayer_on_zlo || SurfLayer_on_zhi)) {
-                    hfx_z(i,j,k) = zflux(i,j,k) * explicit_fac;
+                    hfx_z(i,j,k) = zflux(i,j,k);
```
asalmgren, line 224: `hfx_z` now stores the full face flux, but `qfx1_z` two lines below still stores `zflux * explicit_fac`. The same asymmetry appears at lines 361, 495 and 626 here, and at the eight matching sites in `ERF_DiffusionSrcForState_S.cpp` / `_T.cpp`.

Concrete case: with `erf.vert_implicit_fac = 1 1 1` (or any setting whose *final* RK stage is implicit) and the default `erf.implicit_moisture_diffusion = true`, a single plotfile reports `hfx3` as the full SGS sensible heat flux while `q1fx3` is identically zero, and `sfs_profiles.dat` shows the same split. Any Bowen-ratio or latent/sensible comparison built off that output is then wrong.

Relatedly: with the default `erf.vert_implicit_fac = {1,1,0}` the last stage is explicit, so both already agreed in the plotfile. The PR description's claim that plotfile `hfx3` was "near zero" with the implicit solve does not hold at default settings — the defect actually being fixed is confined to the intermediate stages that feed the TKE source.

Source/Diffusion/ERF_AddTKESources.H
```diff
@@ -27,7 +30,13 @@
             // **should** be a function of the water vapor and total water
             // mixing ratios, depending on whether conditions are saturated or
             // not (see the WRF model description, Skamarock et al 2019).
-            amrex::Real buoyancy_src = l_abs_g * l_inv_theta0 * hfx_z(i,j,k);
+            // NOTE: reads hfx_z at k+1, one face above bx; Hfx3 is z-nodal and
+            //       the theta diffusion writes every face of its valid box, top included.
+            const bool l_surf_cell = l_surf_flux_klo && (k == l_klo_dom);
+            const amrex::Real hfx_cc = (l_surf_cell)
+                                     ? hfx_z(i,j,k)
+                                     : amrex::Real(0.5) * ( hfx_z(i,j,k) + hfx_z(i,j,k+1) );
```
asalmgren: The `k+1` read rests on an invariant that spans two different functions and two different `MFIter` loops — `erf_slow_rhs_pre` writes the z-faces of `Hfx3`, `erf_slow_rhs_post` reads them — with nothing asserting it.

I checked and the invariant does currently hold: `Hfx3` is `convert(ba,IntVect(0,0,1))` with one ghost, and the `RhoTheta` diffusion loops over `zbx = surroundingNodes(bx,2)`, so every z-face of each valid box (top included) is written in all four `l_consA`/`l_turb` branches of each of `_N`/`_S`/`_T`.

The fragility is that the *closure* (`ComputeTurbulentViscosityLES`/`RANS`) writes `hfx_z` over **cell** tileboxes, so it never writes the top face `khi_valid+1` of any box. If a future path ever runs the `RhoKE` diffusion without the same-stage `RhoTheta` pass, the top cell of every box silently reads a stale or uninitialised face rather than failing loudly. A cheap `AMREX_ASSERT` on the precondition, or a comment in the header naming `erf_slow_rhs_pre` as the required writer, would make the coupling explicit.

Source/DataStructs/ERF_TurbStruct.H
```diff
@@ -175,6 +175,14 @@ public:
         amrex::Error(
           "It is not recommended to use Deardorff LES and a PBL model");
       }
+      // Both the k-eqn RANS closure and these PBL schemes transport RhoKE
+      if ((rans_type == RANSType::kEqn) &&
+          ((pbl_type == PBLType::MYJ) || (pbl_type == PBLType::MYNN25) ||
```
asalmgren, line 180: This abort is correctly placed (inside `if (pbl_type != PBLType::None)`, after the `SHOC -> EAMXX_SHOC` remap) and no deck under `Exec/` currently pairs `kEqn` with a TKE-transporting PBL scheme. It also closes a real hole: `l_apply_surface_layer_fluxes_in_diffusion` is forced `false` in `SlowRhsPost` for EAMXX SHOC but **not** in `SlowRhsPre`, which would have made `SurfLayer_zlo` disagree between the two and halve the surface flux in the first cell.

The parallel assumption for the EB path is not enforced anywhere, though. `ERF_DiffusionSrcForState_EB.cpp` neither writes `hfx_z` (the line is commented out) nor includes `ERF_AddTKESources.H`, and nothing in `init_params` rejects `erf.terrain_type = EB` combined with `erf.les_type = Deardorff` or `erf.rans_type = kEqn`. Such a run silently integrates `RhoKE` with no buoyancy, shear or dissipation source. Pre-existing rather than introduced here, but it is the same class of hole this block just closed for PBL, and this is the natural place for the guard.

Tests/CTestList.cmake
```diff
@@ -1299,6 +1299,79 @@ endfunction(add_test_rans)
...
+        ATTACHED_FILES_ON_FAIL "${log_a};${log_b};${check_log}"
+    )
+endfunction(add_test_rans_pair)
+
+# The buoyancy production of k must not depend on how theta is diffused: the
+# convective case, run compressible with the implicit vertical solve (A) and
+# with explicit vertical diffusion (B), must give the same KE to within the
+# time-discretisation difference. The command runs through sh, so not on Windows.
+if(NOT WIN32)
+    add_test_rans_pair(RANS_Convective_ABL_Flat_Buoyancy_kEqn Convective_ABL_Flat inputs_convective 40 check_implicit_explicit_ke.py
```
asalmgren, line 1516: These two entries run `inputs_convective`, which sets `amr.n_cell = 8 8 100`, `amr.max_grid_size = 100` and `fabarray.mfiter_tile_size = 1024 1024 1024` — a single box with tiling effectively off. With `ERF_TEST_NRANKS = 2` one rank owns no data, so the test never exercises a box or tile boundary in z.

That is precisely the decomposition the new `hfx_z(i,j,k+1)` read in `ERF_AddTKESources.H` depends on: a wrong-side or missing face write across a z-split would pass this test. Adding `amr.max_grid_size = 8 8 32` (or an equivalent `RUNTIME_OPTIONS` override) to at least one of the two entries would close the gap for essentially no extra runtime.

Exec/CanonicalTests/Canonical_RANS/Convective_ABL_Flat/check_implicit_explicit_ke.py
```diff
@@ -0,0 +1,75 @@
...
+FIELDS = ["KE", "theta"]
+
+
+def main(argv):
+    tol = None
+    args = []
+    it = iter(argv)
+    for a in it:
+        if a == "--tol":
+            tol = float(next(it))
```
asalmgren: Minor: `float(next(it))` raises an uncaught `StopIteration` traceback if `--tol` happens to be the last argument, instead of falling through to the usage message and `return 2` below.

asalmgren (commented):
Reviewed the full diff (13 files). The core change checks out: `Hfx3` is z-nodal with one ghost so the `k+1` read is in bounds even at the domain top, `erf_slow_rhs_pre` writes every z-face of each valid box over `zbx = surroundingNodes(bx,2)` before `erf_slow_rhs_post` runs the `RhoKE` call, and the `hfx_z` write precedes the `zflux *= explicit_fac` scaling in all three kernels so it really is the full flux. Other `Hfx3` consumers (the implicit LU solves, SHOC preprocess, the 2D plotfile, the line sampler, REMORA) only read the domain bottom, and `ERF_Plotfile.cpp` already face-averages `hfx3`, matching the new doc text.

Four low-severity comments inline, plus one nit.

Source/Diffusion/ERF_DiffusionSrcForState_N.cpp
hgopalan: Fixed in 392693add: `qfx1_z` now stores the full face flux at all twelve sites too, so `q1fx3` and `hfx3` agree for every `vert_implicit_fac`. Above the surface, `qfx1_z` is read only by diagnostics (plotfile, SGS profiles, line sampling), so no physics changes. The `q1fx3` row in the plotfile reference is updated.

You're right about the description as well. With the default `1 1 0` the last stage is explicit, so the output was already the full flux; only a setting whose final stage is implicit showed zero. The description now says so.

Source/Diffusion/ERF_AddTKESources.H
hgopalan: Added in 392693add:
- **Comment:** the header comment in `ERF_AddTKESources.H` names the `RhoTheta` call in `erf_slow_rhs_pre` as the required same-stage writer of the z-faces. It also notes that the closures loop over cell boxes and never write a box's top face.
- **Assert:** an `AMREX_ASSERT` checks that the z-faces of `bx` lie in `hfx_z`.
- **Test:** the new z-split box-parity tests (next thread) fail if the top face of a box is not written for that box.

Source/DataStructs/ERF_TurbStruct.H
hgopalan: Added in 392693add: `SolverChoice::init_params` now aborts when `erf.terrain_type = EB` is combined with Deardorff or k-eqn RANS, with a message naming both inputs.
- The abort fires for both closures on `ABL/inputs_most_nbl_EB`; Smagorinsky still runs.
- Development ran that deck with Deardorff without complaint.
- The gap goes back to when the EB diffusion path was written without the TKE sources.

Tests/CTestList.cmake
hgopalan: Added as separate box-parity tests, because this deck cannot simply be split in z, for three reasons:

- **`amr.max_grid_size = 8 8 32` is a per-level list, not per direction.** At level 0 it means 8 in every direction. With x and y already ≥ 8, `MakeNewLevelFromScratch` splits each column into one box per rank: a single box on 1 rank, 50/50 on 2.
- **The implicit run cannot be split in z at all:** `define_column_kextent` aborts.
- **Two older bugs make this deck depend on the z decomposition on `development`,** with or without this PR, so a real z-split fails on its own:
  1. **The acoustic substep** closes the implicit `w` solve with Dirichlet rows at every box's `lo.z`/`hi.z+1` (`ERF_MakeFastCoeffs.cpp`, "If at the bottom of the grid"). It dates from 5bbcef32e ("generalizations for multiple grids", #1762, 2024-08-28), and the `ABL_MOST_WOA_ZSplit_NoSub` deck already turns substepping off for this reason. With substepping on, the z-split run of this deck reaches negative rho by step 2, identically with the development binary.
  2. **The PBL-height estimator** in `ERF_PBLHeight.H` sets each tile's z range to the whole domain (`gtbx.setSmall(2,domain.smallEnd(2))`, `setBig(2,domain.bigEnd(2))`). An upper box therefore reads `cons` far outside its own data and computes its own planar `pblh`. That dates from f73c64bbc (#1756, 2024-08-29). The k-eqn length cap `erf.rans_lscale_from_pblh` (105502a67, merged in #3971 on 2026-09-13) is the first TKE-closure consumer of that field, so a z-split k-eqn run is decomposition-dependent: KE differs by 0.28 % after one step, and exactly 0 with the cap off.

**The new tests.** `RANS_Convective_Buoyancy_ZSplit_BoxParity_kEqn` and `_Deardorff` use `add_test_box_parity`:
- **Setup:** the convective RANS case, compressible, explicit diffusion, substepping off, no PBL-height cap. One box on one rank is compared with 10 boxes on 2 ranks (split at k = 24, 48, 72, above the inversion, where the heat flux is non-zero), with `rel_tol 1e-9`.
- **Results:**
  - The branch gives identical plotfiles (KE and theta error 0).
  - A mutant that skips the `hfx_z` write on each box's top face fails: KE error 1.1e-6 (kEqn) and 3.4e-6 (Deardorff).
  - A mutant that reads the lower face in each box's top cell fails: 2.0e-9 and 6.4e-9.
  - Development passes, since it has no cross-box read.
  - Debug no-FFT: 16 s each.

The two decomposition bugs above are outside this PR. Happy to open issues for them.

Exec/CanonicalTests/Canonical_RANS/Convective_ABL_Flat/check_implicit_explicit_ke.py
hgopalan: Fixed in 392693add: `--tol` without a value, or with a non-numeric one, now prints the usage and returns 2. I checked `--tol` alone, `a b --tol` and `--tol abc x y`; the previous version raised `StopIteration`.

Source/DataStructs/ERF_TurbStruct.H
hgopalan: The EB path is from f50f87689 ("EB: Smagorinsky", #2962, 2026-03-12). `ERF_DiffusionSrcForState_EB.cpp` was added for Smagorinsky only and never included `ERF_AddTKESources.H`, and nothing rejected a TKE closure on EB, so the gap has been there since that commit.

hgopalan:
Thanks, @asalmgren. All five comments are addressed in 392693add; the details are in each thread.

1. **`qfx1_z`** now stores the full flux as well, so `q1fx3` and `hfx3` agree. I also corrected the description's claim about the default `1 1 0`.
2. **The `k+1` invariant:** a header comment names `erf_slow_rhs_pre` as the required writer, and an `AMREX_ASSERT` covers the face range.
3. **EB with Deardorff or k-eqn** now aborts. The gap dates from f50f87689 (#2962).
4. **z box boundaries:** new `RANS_Convective_Buoyancy_ZSplit_BoxParity_{kEqn,Deardorff}` tests. They pass on the branch and fail with a top-face-write mutant and a box-top-read mutant. The deck turns off two things that make it depend on the z decomposition on development, whatever this PR does:
   - the acoustic substep `w` solve, closed at every box's lo/hi, from 5bbcef32e (#1762);
   - the PBL-height estimator, which loops each box over the full domain height, from f73c64bbc (#1756).
5. **`--tol`** parsing is fixed.

**Verification:**
- **Release:** regression 112/114. The remaining failures are `ABL_MOST_WOA_ZSplit` and `ABL_MOST_Cloudchamber`, gold differences on this machine that the development binary reproduces digit for digit. Unit tests pass 572/572, parallel tests 41/41.
- **Debug no-FFT:** 12/12, with the assert active.
- **Also run:** single precision with `-Wshadow`, Sphinx, style scripts, and the GPU lint.

Tests/CTestList.cmake
hgopalan: Rather than opening issues, I fixed the three older decomposition bugs found while adding the z box-boundary test, each in its own PR with a z-split test that fails on development:

- #4042: **acoustic substep on grids split in z.** The implicit `w` solve closes every box's column; this goes back to 5bbcef32e (#1762). It is now a start-up/regrid abort next to the implicit-diffusion one (#3998), with `erf.substepping_type = None` named as the way to run such a case. All 428 input decks were started for zero steps and none hits the abort; ERF only splits columns when an input asks for it.
- #4043: **surface history averages multiplied by the number of stacked boxes.** The sums go back to 5079e48ba (#928). They are now summed over the surface copies only, and the existing `ABL_MOST_WOA_ZSplit_NoSub_BoxParity` compares `surf_hist.dat` as well.
- #4044: **MYNN25 PBL height on grids split in z.** The estimator's full-height loop per box goes back to f73c64bbc (#1756). The estimator now runs on full-height column boxes when the grids are split, checked by a new `PBLH_ZSplit_BoxParity`.

With #4044 in, the `RANS_Convective_Buoyancy_ZSplit` deck here could turn the PBL-height cap back on; I'll do that once it merges. This branch also merges current `development` (6788ce03e): the only conflict was `Tests/CTestList.cmake` next to #4033's self-test, and both are kept.

Source/Diffusion/ERF_AddTKESources.H
```diff
@@ -27,7 +39,13 @@
             // **should** be a function of the water vapor and total water
             // mixing ratios, depending on whether conditions are saturated or
             // not (see the WRF model description, Skamarock et al 2019).
-            amrex::Real buoyancy_src = l_abs_g * l_inv_theta0 * hfx_z(i,j,k);
+            // NOTE: reads hfx_z at k+1, one face above bx; Hfx3 is z-nodal and
+            //       the theta diffusion writes every face of its valid box, top included.
+            const bool l_surf_cell = l_surf_flux_klo && (k == l_klo_dom);
+            const amrex::Real hfx_cc = (l_surf_cell)
+                                     ? hfx_z(i,j,k)
```
AMLattanzi: This block specifically guards that when a surface layer is present and we are at the low cell center we do **not** average. This places the MOST heat flux, at the w-face, in the first cell center slot. Why not always average to the cell center for consistency? The only other approach I might consider, is using the proposed MOST profile and then analytically differentiating and evaluating at the z of the first CC (experimenting with that in another PR now).

Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp
```diff
@@ -277,7 +277,9 @@ void ComputeTurbulentViscosityLES (Vector<std::unique_ptr<MultiFab>>& Tau_lev,
 
                 // - heat flux
                 //   (Note: If using SurfaceLayer, the value at k=0 will
-                //    be overwritten)
+                //    be overwritten; above the surface the theta diffusion
+                //    replaces it with its face fluxes, which the TKE buoyancy
+                //    source reads)
                 hfx_x(i,j,k) = zero;
                 hfx_y(i,j,k) = zero;
                 hfx_z(i,j,k) = -mu_turb(i,j,k,EddyDiff::Theta_v) * dtheta_dz; // (rho*w)' theta' [kg m^-2 s^-1 K]
```
AMLattanzi: These operations now appear to have two issues:
1. They are not consumed and are overwritten when diffusion is completed. Since these computations are done in Deardorff, we have diffusion on.
2. They are not consistent with the BA index type of hfx. hfx_z is (0,0,1) and we are writing data computed at CC into it.

AMLattanzi (changes requested):

Source/Diffusion/ERF_AddTKESources.H
hgopalan: Agreed, done in f56125ce5. Every cell, the first included, now averages its two z-faces. With a surface layer the bottom face holds the surface-layer flux, so the first cell uses ½(surface flux + diffusion flux at z = dz). I also rewrote the header comment.

A new unit test, `TKEBuoyancySource.FirstCellAveragesTheSurfaceFluxWithTheFaceAbove`, checks this. It sets a zlo surface layer, puts a surface flux on the bottom face, and requires the first cell's source to equal |g|/θ_ref · ½(S − K(b + 2c·dz)) exactly, for explicit and implicit diffusion. It fails when the special case is restored. The two column tests now cover every face and cell, the first and last included, instead of interior ones only.

The implicit-vs-explicit regression pairs move slightly: kEqn 1.12e-5, Deardorff 5.27e-5 (from 5.05e-5).

A MOST-profile gradient at the first cell centre, as you're trying in the other PR, would slot in here in place of the face average for that cell.

Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp
hgopalan: Agreed, done in f56125ce5. The Smagorinsky, Deardorff and k-eqn closures no longer write `hfx_x`, `hfx_y` or `hfx_z`, and `ComputeTurbulentViscosityLES` and `ComputeTurbulentViscosityRANS` no longer take `Hfx1/Hfx2/Hfx3`.
- **Nothing read the old values.** `AdvanceDycore` is the only caller of the closures, and the θ diffusion then rewrites every z-face on every stage (the surface layer rewrites the bottom face) before anything reads `Hfx3`. The plotfile at step 0 already showed the zero initial value.
- **No output changed.** Regression is unchanged apart from the three reds this Mac shows on development too (`ABL_MOST_WOA_ZSplit`, `ABL_MOST_Cloudchamber`, `Closure_BoxParity_MYNNEDMF`), and every box-parity test passes.

**One write stays, on EB (`ComputeTurbulentViscosityLES_EB`).** `DiffusionSrcForState_EB` does not write `hfx_z` (the line is commented out), so on EB terrain the closure is the only writer of the SGS heat flux output. Removing it would zero `hfx3` for EB runs. A comment there now says so. Moving that output to EB face fluxes would belong with the EB diffusion, not this PR.

hgopalan:
@AMLattanzi, both comments are addressed in f56125ce5; the details are in each thread.

1. **The TKE source averages the two z-faces in every cell**, the first included. With a surface layer, the first cell uses ½(surface flux + diffusion flux at z = dz). A new unit test fails when the special case is restored.
2. **The Smagorinsky, Deardorff and k-eqn closures no longer write the heat fluxes,** and the LES and RANS closure functions drop their `Hfx` arguments. The EB closure keeps its write, since `DiffusionSrcForState_EB` does not write `hfx_z` and would otherwise leave EB's `hfx3` output at zero.

**Other changes in this push:**
- **Conflict resolved:** the merge of `development` keeps both #4043's `DATALOG` argument and this PR's z-split entries.
- **PBL-height cap back on** in the z-split deck, now that #4044 is in.

The description has the full table of checks rerun on this head.

AMLattanzi:
@hgopalan Thank you. Let me iterate for a minute on the averaging of `hfx_z` at the bottom surface. I can make an argument for your original implementaiton and would like to have a little clarity. If you have a strong opinion, please leave a comment.

AMLattanzi:
@hgopalan For now, I am for the averaging. Can you just leave a comment above that line that notes the MOST flux alone is a defensible candidate at klo and may benefit near-surface TKE production at high vertical resolution?

AMLattanzi (approved):

Source/Diffusion/ERF_ComputeTurbulentViscosity.cpp
hgopalan: Follow-up in 910e14fed: your two points are now recorded in the code, so the reasoning does not live only in this thread.

- **`ERF_AddTKESources.H`:** that `hfx_z` (`Hfx3`) is z-nodal `(0,0,1)` at the w-faces, so `hfx_z(i,j,k)` is the flux at the cell's lower face rather than its cell-centred value; hence the average of the two faces in every cell, and no special case for the first cell, whose surface flux also sits at a w-face. It also notes the MOST-profile gradient at the first cell centre as the consistent alternative you are trying.
- **`ERF_ComputeTurbulentViscosity.cpp`:** each non-EB closure says why it stores no heat flux (a cell-centred value in a z-face array, overwritten by the theta diffusion of every stage before any read), and the EB write says it keeps that mismatch because the EB diffusion writes no `hfx_z`, with the note that moving the EB output to the EB face fluxes belongs with that diffusion.
- **`ERF_EddyViscosity.H`:** the LES declaration states that no subgrid heat flux is stored.

Comments only, no behaviour change: the build is warning-free, the unit tests and the buoyancy and RANS entries pass, and Sphinx and the style checks are clean.

asalmgren (commented):
Automated code review. The core change holds up — I traced it end to end and found no correctness problems in the source. Three low-severity findings, all inline below (two doc/comment accuracy, one test sensitivity).

<details>
<summary>Invariants checked and confirmed</summary>

- `hfx_z(i,j,k) = zflux(i,j,k)` is written **before** `zflux *= explicit_fac` at all 12 sites across `_N`/`_S`/`_T`, so the RHS and the reflux register are unchanged.
- Every non-diagnostic reader of `Hfx3`/`Q1fx3` reads only `klo`: `ImplicitDiffForStateLU_{N,S,T}` takes the `scalar_zflux(i,j,klo)` branch only under `at_zlo`; SHOC (`ERF_ShocPreprocess.cpp:98`), REMORA (`ERF_to_REMORA.cpp:576`) and `ERF_Plotfile2D.cpp` all read `klo`. The one case where the domain bottom is not the surface-layer value (native/EAMXX SHOC with `l_apply_surface_layer_fluxes_in_diffusion=false`) is safe because `host_owns_vertical_scalar_diffusion()` forces `implicit_{thermal,moisture,ke}_diffusion` off for the whole SHOC family, so the implicit path never reads it.
- The new `k+1` read is safe across box boundaries: `erf_slow_rhs_pre` runs the RhoTheta pass on `TileNoZ` tiles over the same BoxArray and writes `zbx = surroundingNodes(bx,2)`, so each valid box's top face is written before `erf_slow_rhs_post` (`TilingIfNotGPU`) reads it, and `Hfx3` carries a 1-cell ghost. `AddTKESources` fires exactly once per stage (`start_comp==end_comp==RhoKE_comp` only in Post; Pre passes `n_start=RhoTheta_comp, n_comp=1`).
- Dropping the closures' `hfx_x/hfx_y/hfx_z` writes is safe: they wrote zero into `Hfx1`/`Hfx2`, which are `setVal(0)` at allocation and re-written by `SurfaceLayer` at the boundary faces each stage (`ERF_SurfaceLayer.cpp:950-951`); `Hfx3` was overwritten by the theta diffusion before any read. `dtheta_dz`/`inv_Pr_t` remain used in all three lambdas, so no unused-variable fallout.
- The new aborts close real gaps (`DiffusionSrcForState_EB` has its `hfx_z` writes commented out and no `AddTKESources` include); `RANSType` has only `None`/`kEqn`, so EB+RANS is fully covered, and no in-tree deck hits either abort.
- `ABL_MOST` and `Deardorff_stationary` — the only gold-compared Deardorff decks — both set `erf.use_gravity=false`, so the buoyancy term is identically zero and their gold files do not need regenerating.

</details>

Docs/sphinx_doc/theory/RANS.rst
```diff
@@ -74,6 +74,18 @@ the plotfile ``diss`` holds :math:`\rho \varepsilon` from the start of the
 last step. A floor ``erf.tke_floor`` (default: machine epsilon on
 :math:`\rho k`) bounds :math:`k` from below.
 
+The heat flux in :math:`P_b` at the cell centre is the average of the
+fluxes of the :math:`\theta` diffusion at the two vertical faces of the
+cell, in every cell; at the bottom of the domain the face flux is the
+surface-layer flux when a surface layer is used. These are the full face
+fluxes, including the metric terms of stretched and terrain-fitted
```
asalmgren: **LOW — this claim is not true for terrain-fitted meshes.**

`ERF_DiffusionSrcForState_T.cpp:265` stores

```cpp
zflux(i,j,k) = -rhoAlpha * GradCz;
...
hfx_z(i,j,k) = zflux(i,j,k);
```

i.e. only the along-ζ component. The horizontal slope corrections and the Jacobian are applied *later*, in the RHS loop at `ERF_DiffusionSrcForState_T.cpp:790-806`:

```cpp
zflux_hi = zflux(i,j,k+1)
         - met_h_xi_hi  * mf_mx(i,j,0) * xfluxbar_hi
         - met_h_eta_hi * mf_my(i,j,0) * yfluxbar_hi;
...
stateContrib /= detJ(i,j,k);
```

None of that reaches `hfx_z`. So for Deardorff or k-eqn over terrain (Witch-of-Agnesi, Hemisphere), on the slopes both the TKE buoyancy production and the `hfx3` plotfile output differ from the physical vertical SGS heat flux by the omitted `met_h_xi`/`met_h_eta` terms — while the docs now assert they are accounted for.

The omission itself is pre-existing; only the claim that it is handled is new. Simplest fix is to narrow the sentence to stretched meshes. The same sentence appears at `Docs/sphinx_doc/theory/DNSvsLES.rst:237` ("These face fluxes carry the metric terms of stretched and terrain-fitted meshes").

Tests/test_files/RANS_Convective_Buoyancy_ZSplit/RANS_Convective_Buoyancy_ZSplit.i
```diff
@@ -0,0 +1,109 @@
+# ------------------  INPUTS TO MAIN PROGRAM  -------------------
+# Box parity of the TKE buoyancy source across z box boundaries: the convective
+# Canonical RANS case (Canonical_RANS/Convective_ABL_Flat), compressible with
+# explicit vertical diffusion, run on one box and on boxes split in z. The k
+# source averages the theta-diffusion fluxes at the two z-faces of each cell, so
+# the top cell of every box reads a face written for that box; the two runs must
+# agree. The splits (k = 24, 48, 72) sit in the stratified air above the
```
asalmgren: **LOW — the stated split locations are wrong, and the comment contradicts itself.**

With `amr.n_cell z = 100` and `amr.max_grid_size_z = 24`, `BoxList::maxSize` (`Submodules/AMReX/Src/Base/AMReX_BoxList.cpp`) first divides out the common factor of 2 twice:

```
bs = 24, nlen = 100  ->  ratio = 2, bs = 12, nlen = 50
                     ->  ratio = 4, bs =  6, nlen = 25   (25 is odd, stop)
numblk = (25 + 6 - 1)/6 = 5 ;  sz = 25/5 = 5 ;  extra = 0
klo = k * sz * ratio = k * 20
```

So the run gets five 20-cell boxes and the splits are at **k = 20, 40, 60, 80** (z = 400, 800, 1200, 1600 m), not k = 24, 48, 72.

Two of the four splits are therefore *below* the 937 m inversion, not "in the stratified air above" it — and the quoted k = 24 is z = 480 m, which is below the inversion as well, so the comment is inconsistent with itself. The "2 x 1 x 5" box count on line 22 is correct; only the split locations and their stated rationale are wrong, which misdescribes what the parity test actually covers.

`Tests/CTestList.cmake:1100` repeats the "z at 24" claim and should be updated too.

Tests/Unit/Diffusion/ERF_GTestTKEBuoyancySource.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+Real tol (Real scale)
+{
+    return Real(128.0) * std::numeric_limits<Real>::epsilon() * std::max(Real(1.0), scale);
+}
+
+// The buoyancy production of k reads the flux of the theta diffusion at the two faces of each cell.
+// With the implicit vertical solve (implicit_fac = 1) the stored face fluxes must still be the full
+// fluxes, not the explicit fraction (which is zero), and the source must not change.
+TEST(TKEBuoyancySource, FaceFluxesAreFullForExplicitAndImplicitDiffusion)
```
asalmgren, line 269: **LOW — this test is vacuous in single precision.**

`flux_scale` (line 253) includes a `K * a / dz` term = `0.8 * 300 / 0.5` = 480, which dominates everything else in it. With `tol(scale) = 128 * eps * scale` and single-precision `eps ~ 1.2e-7`, that gives `tol(flux_scale) ~ 7.4e-3`, while the quantities this test actually checks are `|hfx_z| = K*b = 3.2e-3` and `|qfx1_z| = K*|qb| = 2.4e-4`; the source is `~1.05e-4` against a `src_scale` tolerance of `~2.4e-4`.

Scenario: build in single precision and restore `hfx_z = zflux * explicit_fac`. With `implicit_fac = 1` the stored flux becomes exactly 0, and all three `EXPECT_LE` on lines 270-272 still pass — the test cannot see the mutant it was written to catch.

The other two tests carry `ASSERT_GT` resolvability guards (lines 282 and 298-299) and do still catch it, so the suite as a whole is not blind; this one should get the same guard, or drop the `K * a / dz` term from `flux_scale` for the `c = 0` case. In double precision the test is fine.

Docs/sphinx_doc/theory/RANS.rst
hgopalan: You're right, and the claim was mine to make wrong. Fixed in 90f7366de.

`DiffusionSrcForState_T` stores `zflux = -rhoAlpha * GradCz`, the component along ζ, and the slope corrections and the Jacobian are applied later, where the fluxes enter the RHS. So both the buoyancy production and the `hfx3` output omit the `met_h_xi`/`met_h_eta` terms over sloping terrain.

- **RANS.rst and DNSvsLES.rst** now say the stored fluxes carry the stretched-mesh spacing, and state the terrain-fitted caveat explicitly instead of claiming the metric terms are handled.
- **`ERF_AddTKESources.H`** carries the same caveat next to the average, so it stays with the code.

The omission itself is pre-existing and left alone here.

Tests/test_files/RANS_Convective_Buoyancy_ZSplit/RANS_Convective_Buoyancy_ZSplit.i
hgopalan: Confirmed and fixed in 90f7366de. I read the box list back from the split run's plotfile: five boxes of 20 cells, `((0,0,0)-(3,7,19))`, `((0,0,20)-(3,7,39))`, … so the splits are at k = 20, 40, 60, 80, exactly as your trace through `BoxList::maxSize` predicts.

That is z = 400, 800, 1200 and 1600 m: two below the 937 m inversion and two above it. The heat flux is non-zero at all four from the first step, so the coverage is fine, but the comment was wrong twice over. The deck comment now gives the real locations and says why `max_grid_size_z = 24` yields 20-cell boxes; `Tests/CTestList.cmake` is updated to match.

Tests/Unit/Diffusion/ERF_GTestTKEBuoyancySource.cpp
hgopalan: Confirmed and fixed in 90f7366de. I built with `ERF_PRECISION=SINGLE` and reproduced your scenario: with `hfx_z = zflux * explicit_fac` restored, the old test passed in single precision.

- **Larger gradients:** b = 0.1, qb = −0.1 (qa = 1.0) and c = 0.3, so the checked quantities are K b = 0.08 and K |qb| = 0.08 against a single-precision tolerance of ~7e-3.
- **The same guard as the other two:** `FaceFluxesAreFullForExplicitAndImplicitDiffusion` now asserts that each checked flux, and the source, exceed four tolerances, so it fails loudly rather than silently if the margin ever disappears.
- **The file header** records that the roundoff floor comes from the theta differences, eps · K a / dz, and that the gradients are chosen above it.

In single precision all three tests pass, and with the mutant all three fail. In double precision they pass, and the 575 unit tests are unchanged. (Single precision has 32 other unit-test failures, in two-stream radiation, Kessler and WENO; those files are untouched here.)

hgopalan:
Thanks @asalmgren — all three are fixed in 90f7366de, details in each thread.

1. **Terrain claim corrected.** The stored face flux is the ζ component only, so the docs no longer claim the terrain metric terms are carried; the caveat is in both theory pages and next to the code.
2. **Split locations corrected.** Read back from the split run's plotfile: five 20-cell boxes, splits at k = 20, 40, 60, 80 (z = 400–1600 m), two below the inversion and two above. Deck and `CTestList.cmake` comments updated.
3. **Single-precision sensitivity closed.** Built with `ERF_PRECISION=SINGLE` and reproduced the blind spot; larger gradients plus the same four-tolerance guard as the other two tests. In single precision the three tests now pass and all three fail with the mutant.

No behaviour change in this round: docs, comments, and test inputs and guards only. Release unit tests 575/575, the buoyancy, z-split parity and RANS entries pass, and Sphinx and the style checks are clean.

PR #4018, opened 2026-09-15
Fix 4012 rrtmgp surface temperature

### LINKED ISSUE #4012: RRTMGP uses the surface layer's potential temperature as an absolute temperature for the longwave boundary
The RRTMGP interface takes the surface layer's `t_surf` and uses it directly as an absolute temperature, but the surface layer works in potential temperature, so the longwave surface boundary is too warm by the Exner factor of the surface.

Follow-up to #4004, which fixed the same substitution on the two-stream side and deliberately left RRTMGP untouched (fixing it changes RRTMGP answers). Until this is fixed the two radiation models disagree on the longwave surface boundary in exactly the configuration #4004 set out to unify, while `Docs/sphinx_doc/Inputs.rst` now tells users the two models read the sun, site and surface-temperature inputs "with the same meaning".

## Where

`Source/Radiation/RRTMGP/ERF_Radiation.cpp` (`Source/PhysicsInterfaces/Radiation/ERF_Radiation.cpp` before #4004), in `mf_to_kokkos_buffers`:

```cpp
// We have a SurfLayer (enforce consistency with temperature)
else if (tsurf_arr && (ivar==0)) {
    rrtmgp_to_fill(icol) = tsurf_arr(i,j,k);
    if (lsm_in_arr) { lsm_in_arr(i,j,k) = tsurf_arr(i,j,k); }
}
```

`ivar == 0` is `t_sfc`, and `t_surf` is `SurfaceLayer::get_t_surf(lev)` (`ERF_AdvanceRadiation.cpp`, passed through `set_grids`).

## Why it is a potential temperature

MOST works in theta throughout. In `Source/BoundaryConditions/ERF_MOSTStress.H` the surface value is formed and consumed against the plane-averaged *potential* temperature at `zref`:

- flux mode (`theta_type == HEAT_FLUX`, which `erf.most.surf_temp_flux` selects): `t_surf_arr(i,j,k) = mdata.surf_temp_flux * (log(zref/z0) - psi_h) / (u* kappa) + tm_arr(i,j,k)`, with `tm_arr` the plane-averaged theta;
- temperature mode: `tstar = kappa * (tm_arr(i,j,k) - t_surf_arr(i,j,k)) / (C - psi_h)`, again differenced against theta.

## Consequences

The value reaches three places:

1. `t_sfc`, the surface Planck source of the longwave solver;
2. `t_lev(icol, 0)`, via the "enforce consistency between t_sfc and t_lev at bottom surface" loop right below;
3. the land-surface model's own `t_sfc` field, through the `lsm_in_arr(i,j,k) = tsurf_arr(i,j,k)` write-back on the same line, whenever an LSM is present but the column is not over land or has no valid data.

The size of the error is `(1/Pi)^4 - 1`, which vanishes at 1000 hPa and grows with elevation: at a surface pressure of 900 hPa, `Pi ~ 0.970`, so `sigma theta^4` exceeds `sigma T^4` by about 13% (theta is roughly 9 K above T for a 290 K surface). Low-elevation LES cases are barely affected; elevated terrain and real cases are not.

## Suggested fix

The same conversion #4004 added to the two-stream sweep: convert the surface-layer value with the Exner function of the lowest cell before it enters the buffer, and leave the LSM field and `erf.rad_t_sfc` alone since those are already temperatures. In the two-stream code that is

```cpp
if (t_sfc_from_field && t_sfc_is_theta) {
    t_surface /= get_inverse_exner(i, j, kmin, state_arr);
}
```

with `get_inverse_exner` returning `1/Pi` from `RhoTheta` and `qv` of the lowest cell (`Source/Radiation/TwoStream/ERF_TwoStreamColumn.H`). RRTMGP already has `p_lay` on hand in the same kernel, so `Pi = getExnergivenP(p_lay(icol,0), RdoCp)` is available without extra state lookups.

The write-back into the LSM field wants a decision too: writing a converted temperature there is correct if the LSM field is a temperature, but that is worth confirming rather than assuming.

## Note on testing

This changes RRTMGP answers wherever a surface layer is active, so any affected `Tests/ERFGoldFiles` entries need regenerating as part of the fix. A gtest mirroring `SurfaceLayerPotentialTemperatureIsConvertedBeforeEmission` (added in #4004 for the two-stream column) would pin the behaviour.

pressel:
Closes #4012.

## Summary

Issue #4012 identified that RRTMGP expects its surface temperature boundary condition to be an absolute temperature, while `SurfaceLayer::get_t_surf()` provides potential temperature.

The SurfaceLayer fallback path was passing that potential temperature directly to RRTMGP, causing it to be interpreted as absolute temperature. This affected both the RRTMGP `t_sfc` boundary value and the bottom `t_lev` value derived from it.

This PR converts the SurfaceLayer value to absolute temperature before it is used by RRTMGP.

## Changes

* Convert SurfaceLayer potential temperature to absolute temperature using the pressure in the lowest atmospheric cell.
* Use the converted value for RRTMGP `t_sfc`.
* Preserve the existing `t_lev(0) = t_sfc` boundary consistency.
* Write the converted absolute temperature back to the LSM `t_sfc` fallback field when applicable.
* Leave valid LSM surface temperatures unchanged, since those are already absolute temperature.
* Preserve the existing fallback behavior over water, where SurfaceLayer takes precedence over an available LSM value.
* Clarify in the user documentation that `erf.rad_t_sfc` is an absolute surface temperature in K.

The lowest-cell pressure is used for the Exner conversion, matching the existing TwoStream treatment introduced in #4004.

## Testing

Added focused RRTMGP unit coverage for:

* SurfaceLayer potential-temperature conversion at reduced pressure.
* Absolute-temperature writeback to the LSM fallback field.
* SurfaceLayer precedence over an available LSM value for water columns.
* Preservation of valid land-model `t_sfc` without an additional Exner conversion.
* Preservation of the configured absolute-temperature fallback when neither SurfaceLayer nor a valid LSM value is available.

The regression tests compute the expected temperature independently as

`T = theta * (p / p0)^(Rd/cp)`

at `p = 0.9 p0`, so the original potential-vs-absolute-temperature error from #4012 is observable rather than hidden by a near-reference-pressure case.


AMLattanzi (approved):

Source/Radiation/ERF_RRTMGP_SurfaceTemperature.H
```diff
@@ -0,0 +1,63 @@
...
+                              amrex::Real& t_sfc,
+                              amrex::Real* lsm_t_sfc_out)
+{
+    // A valid land-model value is already absolute temperature.  A valid LSM
+    // value over water follows the existing fallback chain instead.
+    const bool use_lsm_t_sfc = is_land && has_lsm_t_sfc && valid_lsm_t_sfc;
+
+    if (use_lsm_t_sfc) {
+        t_sfc = lsm_t_sfc;
+    } else if (has_surface_layer) {
```
asalmgren, line 47: **MEDIUM — the Exner conversion is applied to `t_surf` unconditionally, but `t_surf` holds an *absolute* temperature in several SurfaceLayer configurations.**

`resolve_surface_temperature` converts the SurfaceLayer value whenever `has_surface_layer` is true. But `SurfaceLayer::t_surf` is only a potential temperature in the MOST-derived paths. It is filled with absolute temperatures by:

- `fill_tsurf_with_sst_and_tsk` — WRF `SST` over water and `TSK` over land (`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:1317` and `:1320`),
- `fill_tsurf_with_sfc_sst` — the sfc-file SST (`:1361`),
- `fill_tsurf_with_coupled_sst` — coupled ocean SST (`:1519`).

ERF itself already consumes it as an absolute temperature in one place: `ERF_SurfaceLayer.cpp:1412` passes `t_surf_arr(i,j,k)` straight into `erf_qsatw`, which wants `T`, not theta. So the field's meaning is genuinely provenance-dependent rather than uniformly theta.

**Concrete failure scenario.** A WRF real case on elevated terrain (surface ~900 hPa) run with RRTMGP + SurfaceLayer and no valid LSM `t_sfc`: `TSK = 290 K` is a skin temperature read from `wrfinput`, and this code hands RRTMGP `290 * 0.970 ~= 281 K`. That is a ~13% error in the surface Planck source `sigma T^4` — the same magnitude as the bug #4012 reports, with the sign flipped. The same thing happens over water via SST, though there it is small (~0.4 K) because `p_lay(icol,0) ~= p_0`.

**Inherited, not introduced.** TwoStream makes the identical unconditional assumption — `ERF_TwoStreamRadiation.cpp:631` sets `t_sfc_is_theta = true` for *any* `t_surf` source — so this came in with #4004 rather than with this PR, and this PR is faithfully matching it. But since #4012's whole premise is unifying the two models' surface boundary, it is worth gating the conversion on the actual provenance of `t_surf` (e.g. skip it when `m_sst_lev` / `m_tsk_lev` / `m_use_sfc_sst` / coupled SST supplied the value) in both models, rather than assuming theta everywhere.

Source/Radiation/ERF_RRTMGP_SurfaceTemperature.H
```diff
@@ -0,0 +1,63 @@
...
+        // SurfaceLayer::get_t_surf() returns theta.  RRTMGP requires T.
+        t_sfc = getTgivenPandTh(surface_pressure, surface_layer_theta, RdoCp);
```
asalmgren, line 49: **LOW — no guard on `surface_pressure`, unlike the TwoStream routine this is modelled on.**

`getTgivenPandTh(P, th, RdoCp)` evaluates `th / std::pow(p_0/P, rdOcp)`. With `P == 0` that yields `0`; with a non-finite `P` it yields `NaN`. The reference implementation this PR is matching, `get_inverse_exner` (`Source/Radiation/TwoStream/ERF_TwoStreamColumn.H:410-418`), explicitly checks `std::isfinite` and `> 0` and falls back to `1.0` (i.e. no conversion) when the state is degenerate.

The consequence here is wider than in TwoStream, because the result propagates to three places: the RRTMGP `t_sfc` buffer, `t_lev(icol,0)` via the `t_lev_tab(icol,0) = t_sfc_tab(icol)` loop at the end of `mf_to_kokkos_buffers`, and the LSM `t_sfc` field via the write-back a few lines below. So a single bad column poisons the longwave boundary *and* persists into the LSM state, instead of degrading gracefully. Mirroring the TwoStream guard (fall back to the unconverted value when `surface_pressure` is not finite and positive) would keep the two implementations symmetric.

asalmgren (commented):
Review of the RRTMGP surface-temperature fix. Three findings (1 MEDIUM, 2 LOW). Two are posted inline; the third is below because it touches a file this PR does not modify.

---

### LOW — `Source/Radiation/Make.package` is not updated for the new header

`Source/Radiation/Make.package` lists every other header in that directory:

```
CEXE_headers += ERF_RadiationInterface.H
CEXE_headers += ERF_RadiationModels.H
CEXE_headers += ERF_OrbCosZenith.H
```

but `ERF_RRTMGP_SurfaceTemperature.H` was not added. The GNUmake build still compiles because `Exec/Make.ERF:361` puts `Source/Radiation` on `INCLUDE_LOCATIONS`, so this is a consistency/packaging gap rather than a break — worth adding the `CEXE_headers +=` line for symmetry with the CMake path.

---

### Things I checked that are *not* problems

Recording these so they don't get re-litigated:

- The new decision table is behaviourally identical to the old `if / else if / else` for every combination of `is_land`, LSM validity and `tsurf_arr`, and for `ivar != 0`. Only the intended theta-to-T conversion changed, and the write-back condition (`lsm_t_sfc_out != nullptr && !use_lsm_t_sfc`) reproduces the old "write back in branches B and C, not A" behaviour exactly.
- `p_lay_tab` is filled in the preceding `MFIter` loop in the same function, so `p_lay_tab(icol, 0)` is valid when read, and index `0` is the *bottom* layer — the same convention the pre-existing `t_lev_tab(icol,0) = t_sfc_tab(icol)` loop assumes. AMReX `Table2D`'s `end` is exclusive, so the `{extent0, extent1}` construction indexes correctly.
- Dropping the `t_lev_bottom` out-parameter in the second commit is safe: the pre-existing "enforce consistency between t_sfc and t_lev at bottom surface" `Kokkos::parallel_for` at the end of `mf_to_kokkos_buffers` still sets `t_lev(icol,0)` from `t_sfc`.
- The LSM write-back does **not** feed back into the surface layer. `SurfaceLayer::get_lsm_tsurf` only fires when an LSM data field is named `theta` or `t_surf` (`Source/BoundaryConditions/ERF_SurfaceLayer.H:498`), and the only model exposing `t_sfc` to RRTMGP is Noah-MP, whose `NOAHMP_LSMDATA_FIELDS` list contains neither name. SLM exposes `theta` but no `t_sfc`, so `lsm_input_ptrs[0]` is null there. So writing absolute temperature into the LSM `t_sfc` field cannot corrupt MOST's `t_surf`.
- Include paths are fine for both build systems; the new header picks up `AMREX_GPU_HOST_DEVICE` / `AMREX_FORCE_INLINE` transitively via `ERF_EOS.H` -> `AMReX.H`, and `Source/Radiation` is an unconditional CMake include directory (`CMake/BuildERFExe.cmake:665`), so the gtest compiles even with `ERF_ENABLE_RRTMGP=OFF`.
- No gold files need regenerating, contrary to the note in #4012: the `Radiation` ctest at `Tests/CTestList.cmake:1100` is guarded on `Tests/test_files/Radiation`, which does not exist.
- The unit-test tolerance (`64 * eps * T`) is comfortable in both double and single precision, even though the test computes `th * pow(P/p_0, .)` while the implementation computes `th / pow(p_0/P, .)`.

PR #4021, opened 2026-09-15
Fix Anelastic & Add Method

AMLattanzi:
# Summary

1. This PR fixes the default SSP RK2 integrator for the anelastic method and ensures it now obtains **2nd** order. 
2. The PR adds the midpoint method, which can be utilized in conjunction with implicit vertical diffusion in the first stage to retain **2nd** order diffusion (not possible with SSP RK2); see below

Proposed mode:
image

Factor through the stages with $$D = \Delta t \lambda_D$$. Then Taylor expand the result and compare to the Taylor expansion of the exponential solution (final result is second-order).c
image

## The scheme

Let $\mathbf{S}$ be the solution vector and $f$ the slow source. 

**MidPoint** — the first stage advances only a half timestep, and the second stage takes
no average at all:

$$\mathbf{S}^{\*}   = \mathbf{S}^n + \frac{\Delta t}{2} f(\mathbf{S}^{n})$$
$$\mathbf{S}^{n+1} = \mathbf{S}^n + \Delta t  f(\mathbf{S}^{*})$$

## Where it lives in the source

| File | Role |
| --- | --- |
| `Source/DataStructs/ERF_DataStruct.H` | `AMREX_ENUM(AnelasticType, RK2, MidPoint)`; per-level `anelastic_type` vector; parsing of `erf.anelastic_type`; the `vert_implicit_fac` overrides |
| `Source/TimeIntegration/ERF_MRI.H` | The two-stage loop itself — `MRISplitIntegrator::setAnelasticType` and the anelastic branch of `operator()` |
| `Source/ERF_MakeNewArrays.cpp` | `initialize_integrator` hands the per-level choice to the integrator |
| `Source/TimeIntegration/ERF_SlowRhsPre.cpp` | `l_anelastic_rk2` gates the trapezoidal averaging of the slow source |
| `Source/TimeIntegration/ERF_SlowRhsPost.cpp` | Same gate for the scalar (post) source |
| `Source/Prob/ERF_CloudChamberBudget.H` | Per-stage flux accounting, which differs between the two schemes |

### The stage loop

In the anelastic branch of `MRISplitIntegrator::operator()`, the only structural
difference is the stage-0 timestep:

```cpp
const bool l_midpoint = (anelastic_type == AnelasticType::MidPoint);

for (int nrk = 0; nrk < 2; nrk++)
{
    old_time_stage = time_stage;

    nsubsteps  = 1;
    dtau       = (l_midpoint && (nrk == 0)) ? myhalf * timestep : timestep;
    time_stage = time + nsubsteps * dtau;

    slow_rhs_pre (*F_slow, S_old, S_new, time, old_time_stage, time_stage, nrk);
    no_substep   (*S_sum,  S_old, *F_slow, time + nsubsteps*dtau, nsubsteps*dtau, nrk);
    slow_rhs_post(*F_slow, S_old, S_new, *S_sum, time, old_time_stage, time_stage, nrk);
}
```

Everything else — the projection, the boundary handling, the reflux on the final stage —
is shared. Under RK2, `dtau` is `timestep` in both stages and the second-stage averaging
in `erf_slow_rhs_pre` / `erf_slow_rhs_post` supplies the Heun weighting. Under MidPoint
that averaging is skipped: `l_anelastic_rk2` is false, so the second stage simply uses
$f(\mathbf{S}^{*})$ over the full step.

### The RK2-only averaging, and why MidPoint skips it

In `erf_slow_rhs_pre`, guarded by `l_anelastic && l_anelastic_rk2 && (nrk == 1)`:

```cpp
cell_rhs(i,j,k,Rho_comp) *= myhalf;
cell_rhs(i,j,k,Rho_comp) += half_dt * (cell_data(i,j,k,Rho_comp) - cell_old(i,j,k,Rho_comp));
```

Since $\mathbf{S}^{*} - \mathbf{S}^{n} = \Delta t\, f(\mathbf{S}^{n})$ under RK2, the second
term reconstructs $\tfrac{1}{2} f(\mathbf{S}^n)$ without keeping the stage-0 source around,
giving the trapezoidal weight. MidPoint needs no such reconstruction — the midpoint state is
the whole point of stage 0 — so the block is bypassed and the RHS stays $f(\mathbf{S}^{*})$.

## Why MidPoint enables the implicit solve

See the derivation at the top of this note for more details. Note that two things break that, and the code enforces both:

* **Solving again in the second stage** would reduce the pair to backward Euler — first order.
  So with `MidPoint`, `vert_implicit_fac[lev][1]` and `[2]` are zeroed. (Entry `[2]` is never
  reached anyway; the anelastic integrator takes only two stages.)
* **Under RK2 there is no midpoint stage at all**, so no stage can carry a second-order
  implicit solve. All three entries of `vert_implicit_fac` are zeroed and the vertical
  diffusion is explicit.

Both overrides live in `SolverChoice::read_params` and print a line at startup saying what was turned off.

## Input

```
erf.anelastic_type = MidPoint     # or RK2 (default); per-level values allowed
```

Ignored — with a printed notice — where no level is anelastic. The startup banner reports
the choice per level:

```
At level 0 :
   anelastic with no substepping, using the MidPoint integrator
```

`erf.vert_implicit` / `erf.vert_implicit_fac` interact with it as described above; the
`Inputs.rst` table carries the same rules.

## Flux accounting in the cloud-chamber budget

`CloudChamberBudget::capture_stage` takes a `trapezoidal` flag, passed as `l_anelastic_rk2`.
The two paths differ because the stages mean different things:

* **RK2** — `nrk = 0` is retained and `nrk = 1` applies the trapezoidal stage weight, so the
  two RHS evaluations are not counted as two physical timesteps.
* **MidPoint** — there is nothing to average. The first stage spans only $\Delta t/2$ and
  exists solely to produce the midpoint state; the second advances the whole step with the
  source evaluated there. Only `nrk = 1` contributes.

## Regression coverage

`AnelasticWallDiffusion_X_MidPoint` (added to `Tests/CTestList.cmake`) reruns the stationary
wall-diffusion state of `AnelasticWallDiffusion_X` with `erf.anelastic_type = MidPoint`, which
is the configuration that leaves the vertical implicit solve on (the `_X`/`_Y`/`_Z` cases opt
out via `vert_implicit`). The theta walls are in $x$, so the $z$ boundaries stay `foextrap`
and the tridiagonal solve is active for theta and for the momenta. A linear profile is a
steady state of the diffusion operator, so the implicit solve must reproduce it exactly.

## Related

* `Docs/sphinx_doc/TimeAdvance.rst` — the equations, alongside the compressible RK3 advance
* `Docs/sphinx_doc/Inputs.rst` — `erf.anelastic_type`, `erf.vert_implicit_fac`
* Findings: the anelastic RK2 path **was** first order in time, but is corrected here


Source/DataStructs/ERF_DataStruct.H
```diff
@@ -940,6 +952,24 @@ struct SolverChoice {
...
+        }
+
+        for (int i = 0; i <= max_level; i++) {
+            query_one_or_per_level_enum_case_insensitive(
+                pp, "anelastic_type", anelastic_type[i], i, max_level);
+        }
+
+        if (pp.contains("anelastic_type") && !any_anelastic) {
+            amrex::Print() << "Ignoring erf.anelastic_type because no level is anelastic" << std::endl;
+        }
```
asalmgren, line 972: **medium** - This prints "Ignoring erf.anelastic_type..." but does not actually reset `anelastic_type[i]` to `RK2`. The derived `l_anelastic_rk2` is then passed on to `CloudChamberBudget::capture_stage` in two places *without* an `l_anelastic` guard (see the comments on `ERF_SlowRhsPre.cpp` and `ERF_SlowRhsPost.cpp`), so the setting is not in fact ignored - it silently changes the budget accounting on a compressible level.

Either reset the vector here:

```cpp
if (pp.contains("anelastic_type") && !any_anelastic) {
    amrex::Print() << "Ignoring erf.anelastic_type because no level is anelastic" << std::endl;
    for (int i = 0; i <= max_level; i++) { anelastic_type[i] = AnelasticType::RK2; }
}
```

or gate both consumers. Resetting here is the narrower fix and makes the printed message true.

Source/TimeIntegration/ERF_SlowRhsPre.cpp
```diff
@@ -978,6 +987,6 @@ void erf_slow_rhs_pre (int level, int finest_level,
     if (cloud_budget && l_use_diff) {
             cloud_budget->capture_stage(CloudChamberBudget::RhoTheta, nrk,
                                         static_cast<Real>(dt), *dflux_x, *dflux_y,
-                                        *dflux_z, geom, 0);
+                                        *dflux_z, geom, 0, l_anelastic_rk2);
```
asalmgren: **medium** - `l_anelastic_rk2` is passed as the `trapezoidal` argument regardless of whether this level is anelastic.

On a *compressible* level (3 stages, `nrk = 0,1,2`) with `erf.anelastic_type = MidPoint` set in the inputs, `l_anelastic_rk2` is false, so `capture_stage` accumulates only the `nrk == 1` fluxes and drops stages 0 and 2 entirely. The RhoTheta boundary balance is then wrong and the cloud-chamber budget checker reports FAIL - for a configuration that `ERF_DataStruct.H` claims to have ignored.

Suggest `(!l_anelastic || l_anelastic_rk2)` here, or fix it at the source in `ERF_DataStruct.H`.

Source/TimeIntegration/ERF_SlowRhsPost.cpp
```diff
@@ -824,7 +825,8 @@ void erf_slow_rhs_post (int level, int finest_level,
             MultiFab qflux_z(*dflux_z, make_alias, qstate, 1);
             cloud_budget->capture_stage(
                 qstate == 0 ? CloudChamberBudget::RhoQv : CloudChamberBudget::RhoQc,
-                nrk, static_cast<Real>(dt_d), qflux_x, qflux_y, qflux_z, geom);
+                nrk, static_cast<Real>(dt_d), qflux_x, qflux_y, qflux_z, geom,
+                0, l_anelastic_rk2);
```
asalmgren: **medium** - Same defect as the RhoTheta capture in `ERF_SlowRhsPre.cpp`: `l_anelastic_rk2` is passed as `trapezoidal` with no `l_anelastic` guard, so a compressible level with `erf.anelastic_type = MidPoint` in the inputs silently mis-accounts the RhoQv/RhoQc moisture budget (only `nrk == 1` fluxes accumulate out of three stages).

Source/DataStructs/ERF_DataStruct.H
```diff
@@ -1358,13 +1384,30 @@ struct SolverChoice {
...
+                    vert_implicit_fac[lev][2] = zero;
+                } else if (anelastic_type[lev] == AnelasticType::MidPoint) {
+                    if (vert_implicit_fac[lev][1] > zero) {
+                        amrex::Print() << "Turning off the second-stage vertical implicit solve at level " << lev
+                                       << " because anelastic with MidPoint solves in the first stage only."
+                                       << std::endl;
+                    }
+                    // The anelastic integrator only takes two stages, so [2] is never used
+                    vert_implicit_fac[lev][1] = zero;
+                    vert_implicit_fac[lev][2] = zero;
```
asalmgren, line 1414: **low** - The MidPoint branch zeroes `vert_implicit_fac[lev][1]` and `[2]` but leaves `[0]` at whatever the user set. The second-order argument in the comment just above requires `[0] == 1` *exactly* - backward Euler over `dt/2` is what makes `S* = (S^n + S^{n+1})/2`.

With e.g. `erf.vert_implicit_fac = 0.5` the user gets a first-order, undocumented hybrid with no warning, while the startup banner still advertises "MidPoint integrator". Worth either warning or forcing `vert_implicit_fac[lev][0] = 1` when it is set to something else.

Docs/sphinx_doc/TimeAdvance.rst
```diff
@@ -56,6 +56,14 @@ where :math:`\mathbf{S}` is the solution vector, in the following two steps:
 
   \mathbf{S}^{n+1} &=& \mathbf{S}^n + \frac{\Delta t}{2} ( f(\mathbf{S}^{n}) + f(\mathbf{S}^{*}) )
 
+The midpoint stages are given by:
+
+.. math::
+
+  \mathbf{S}^{*}   &=& \mathbf{S}^n + \frac{\Delta t}{2} f(\mathbf{S}^n)
+
+  \mathbf{S}^{n+1} &=& \mathbf{S}^n + \Delta f(\mathbf{S}^{*})
```
asalmgren: **low** - Missing the `t`: should be `\Delta t f(\mathbf{S}^{*})`. As written this renders as a bare Delta right next to the correct `\frac{\Delta t}{2}` two lines up.

Source/TimeIntegration/ERF_SlowRhsPre.cpp
```diff
@@ -925,6 +929,11 @@ void erf_slow_rhs_pre (int level, int finest_level,
             if (l_moving_terrain) {
                  rho_w_rhs(i, j, k) *= myhalf * (detJ_arr(i,j,k) + detJ_arr(i,j,k-1));
             }
+
+            if ( l_anelastic && l_anelastic_rk2 && (nrk == 1) ) {
+                rho_w_rhs(i,j,k) *= myhalf;
+                rho_w_rhs(i,j,k) += half_dt * (rho_w(i,j,k) - rho_w_old(i,j,k));
+            }
```
asalmgren, line 936: **low (release process)** - This new `rho_w` trapezoidal term changes the answer for *every* existing anelastic run under the default RK2 type, not just the new MidPoint path.

Gold-file regressions such as `DensityCurrent_anelastic` (`Tests/CTestList.cmake:974`) and the cloud-chamber budget checkers will need regenerated benchmarks. The PR adds a new test but does not mention benchmark regeneration - worth calling out in the PR description so it is not missed at merge time.

asalmgren (commented):
Traced the anelastic stage loop end-to-end (`ERF_MRI.H` -> `ERF_TI_slow_rhs_pre.H`/`ERF_ImplicitPre.H` -> `ERF_TI_no_substep_fun.H` + `project_momenta` -> `ERF_TI_slow_rhs_post.H`/`ERF_ImplicitPost.H`) to check the new MidPoint path and the RK2 correction.

**What checks out.** The MidPoint staging is self-consistent: `slow_dt = new_stage_time - old_step_time` becomes `dt/2` in stage 0 and `dt` in stage 1, so `ERF_ImplicitPre.H`'s `stage_dt`, the `no_substep` update, the `else` branch in `erf_slow_rhs_post`, `gradp`'s `-1/l_dt` rescale in `project_momenta`, the point-implicit immersed-forcing rescale, and `realbdy_compute_interior_ghost_rhs` all pick up the right stage interval. The `(I - (dt/2)L)^-1` solve in stage 0 plus explicit `L S*` in stage 1 does reduce to the implicit midpoint rule as claimed. Every `l_anelastic && nrk==1` trapezoidal site got the new `l_anelastic_rk2` gate (grepped for stragglers - none). The new `rho_w` averaging is placed before the `level > 0` `zmom_crse_rhs` override, which is correct, and the extra grad-phi it injects is annihilated by the subsequent exact projection, so it is consistent with the existing `rho_u`/`rho_v` treatment. I also ran the existing `AnelasticWallDiffusion_X` deck at 2/4/8 ranks to confirm the grid is never chopped in z, so the new MidPoint test will not trip the `define_column_kextent` abort now that `vert_implicit_fac[0]` stays at 1.

Six inline comments below: one medium-severity issue with three sites, two low-severity, plus a release-process note about benchmark regeneration.

Source/DataStructs/ERF_DataStruct.H
AMLattanzi: Fixed in latest commit.

Source/TimeIntegration/ERF_SlowRhsPre.cpp
AMLattanzi: Fixed in latest commit.

Source/TimeIntegration/ERF_SlowRhsPost.cpp
AMLattanzi: Fixed in latest commit.

Source/DataStructs/ERF_DataStruct.H
AMLattanzi: Fixed in latest commit.

Docs/sphinx_doc/TimeAdvance.rst
AMLattanzi: Fixed in latest commit.

Source/TimeIntegration/ERF_SlowRhsPre.cpp
AMLattanzi: Ctests updated. Regtests will also need remaking.

AMLattanzi:
# Test changes
1. CTEST: Density current anelastic
2. REGTEST: All anelastic cases

PR #3960, opened 2026-09-04
Surface energy balance on immersed-boundary building faces (erf.ibseb), with immersed-boundary fixes for the forcing and the MRF/YSUNew schemes

hgopalan:
## Surface energy balance on immersed-boundary building faces

Adds a prognostic surface energy balance on the faces of buildings represented by immersed forcing (`erf.ibseb.*`, off by default). Every wall and roof face carries a skin temperature and a conduction slab and closes

    SW_abs + eps Q_ext + LW_net - H - LE - G = 0

each step, with the sensible flux deposited into the adjacent fluid cell. Each term has a verification case under `Exec/CanonicalTests/SEB/` (one and four ranks, checked against independent Python implementations or analytic solutions), documented in `Docs/sphinx_doc/theory/ImmersedBoundarySEB.rst` and `Docs/sphinx_doc/Inputs.rst`. CI runs the balance through the CTest `IBSEB_Cube` (the PrognosticSkin cube for 40 steps against a gold plotfile) and the immersed-boundary-aware PBL schemes through the tiling-parity tests `PBL_IBAware_MRF_Tiling` and `PBL_IBAware_YSUNew_Tiling`. The balance needs `erf.buildings_type = ImmersedForcing` on a uniform vertical grid without regridding; terrain by immersed forcing is rejected.

| case | what | check |
| --- | --- | --- |
| FaceStorage | compact per-rank face list (struct of device vectors), building ids, checkpoint slots | face counts, area, restart round trip |
| Shortwave | prescribed clear-sky provider (Spencer geometry, Bird beam, Liu-Jordan diffuse), ray-cast shadow | shadow flag vs an independent cast on every face |
| Longwave | hemisphere view fractions, gray sky, ground, isothermal walls | fractions vs independent sampling; closure |
| SensibleHeat | wall function per face, explicit source into rho-theta | formulas; internal-energy budget |
| SlabConduction | implicit slab per face, material CSV (SLUCM schema) | erfc and steady solutions; restart |
| PrognosticSkin | Newton on the skin, made exactly consistent with the implicit slab; `Q_ext` hook for external radiation; bounds as inputs | residual 1e-8 W/m2; closure; independent re-integration to 1e-9 K |
| IsolatedBuilding | isolated building over 24 h at Boulder | sequence of the day; roof shortwave integral to 0.01 %; slab energy |
[...]

PR #3860, opened 2026-08-17
Extend SurfaceLayer for all boundary sides

ckendrick:
This allows the MOST/Surface Layer boundary condition to be specified on all 6 faces instead of only zlo.

Main changes to `SurfaceLayer`/`MOSTAverage` classes:
- This adds a orientation parameter to the `SurfaceLayer` and `MOSTAverage` classes to determine behavior on different faces.
- MOST configuration options are prefixed with the face name if the SurfaceLayer BC type is specified on more than the zlo face (e.g, `erf.most.surf_temp`, can now be `erf.xlo.most.surf_temp`).
- MOST Averages and SurfaceLayer MultiFabs are constructed on planes corresponding to the face rather than at k=0.
- MOSTAverage is extended to compute averages of W and other planar velocities (XZ, YZ)
- MOSTAverage field ptrs are re-ordered from U/V/T/Qv/Qr/W to U/V/W/T/Qv/Qr.
- Local and regional averaging are supported, but not terrain interpolation, time windowing, EB, etc.
- `SurfaceLayer` writes to Tau and heat/moisture flux arrays for each face. Directional velocity averages and orientation is passed to MOSTStress.
- Only the Surface Temperature and Moeng flux pathway are updated for now to handle different faces.

Main changes to `ERF`:
- ERF now constructs a `SurfaceLayer` instance all 6 faces where the BC type is set to `surface_layer` (otherwise nullptr)
- Existing coupling dependent on reading surface layer values (SHOC, PBL, etc) should default to checking and reading on the zlo face only. Existing behavior on the zlo face should not be modified by these changes.
- X and Y hfx and qfx are read in the diffusion routines (N/S/T versions) from the surface layer on the face (if set).
- Each face writes its variables to checkpoint with a face index prefix ("UStar_0, UStar_1", etc). Existing checkpoints without a face prefix should be supported.

A follow-up PR will be made to adapt this for the new cloudchamber interface and problem configuration.

PR #4026, opened 2026-09-15
IBSEB: read the ground surface layer from the per-face vector (development does not compile)

hgopalan:
Development does not compile since #3860 merged. #3860 made `ERF::m_SurfaceLayer` a vector of surface layers indexed by face, and #3960 (merged about an hour earlier) still used it as a single pointer in `ERF::ibseb_advance` (`ERF_IBSEB.cpp` lines 143-146):

```
error: value of type 'amrex::Vector<std::unique_ptr<SurfaceLayer>>' is not contextually convertible to 'bool'
error: no member named 'get_olen' in 'amrex::Vector<std::unique_ptr<SurfaceLayer>>'
```

The wall function beyond neutral reads the ground's Obukhov length and boundary-layer height, so it now takes them from `m_SurfaceLayer[Orientation::zlo()]`, the same form `ERF_Advance.cpp` uses.

Checked on `1dc36961e`:
- A `-k` build of `erf_exec` on development: the only errors are these 8, all in `ERF_IBSEB.cpp`.
- With this change: `erf_exec` and `erf_unit_tests` build (Release, MPI, FFT, all warnings) with no warnings, and 568 unit tests pass.
- `git grep` finds no other single-pointer use of `m_SurfaceLayer` in `Source`.


hgopalan:
Closing as a duplicate: #4025 (merged as ab52daa25) fixes the same lines in ERF_IBSEB.cpp, taking the zlo surface layer the same way.

PR #4032, opened 2026-09-16
Regtest Real W/ NoahMP + RRTMGP

AMLattanzi:
# Summary
This PR modifies the inputs for the WPS regtest directory. The following changes are made:

1. Clean up `inputs_wps_ml`
2. Remove `inputs_wps_noah`
3. Remove `namelist.erf`
4. Add `inputs_noah_rrtmgp` for nightly regtesting of noah + radiation

## Additional Items
1. The `namelist.erf` , `noah.tbl`, and `wrfinput/bdy` files will be uploaded to the regtest machines. There is no need to track the changes of those in github.


Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
```diff
@@ -0,0 +1,143 @@
+# ===== Problem Setup ==========
+erf.prob_name        = ABL
+prob.KE_0            = 0.4
+prob.KE_decay_height = 1000.
+prob.KE_decay_order  = 3
+
+start_datetime = "2024-08-05 00:00:00"
+stop_datetime  = "2024-08-10 00:00:00"
```
asalmgren, line 10: **No `max_step` anywhere in this file, so the test never finishes.**

`ERF.cpp:2981` defaults `max_step` to `INT_MAX` when unspecified, which leaves `stop_datetime` as the only bound. With `start_datetime = 2024-08-05` to `stop_datetime = 2024-08-10` (432,000 s) and `erf.fixed_dt = 0.5`, that is **864,000 steps**.

Every sibling test bounds this explicitly: `inputs_wps_ml` uses `max_step = 1`, `inputs_radiation` uses `max_step = 100`, and the Katrina NoahMP+RRTMGP setup uses `max_step = 86400`. Unless the regtest driver injects `max_step` via runtime params, this needs one here.

Separately, `erf.fixed_dt = 0.5` s against `dx = 25 km` (2,500 km / 100 cells) is ~4 orders of magnitude below the advective limit. That's fine for a short smoke test, but it is what makes the missing `max_step` so expensive.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
```diff
@@ -0,0 +1,143 @@
...
+
+amr.v = 1
+erf.v = 1
+erf.sum_interval = 1
+
+erf.check_file = chk
+erf.check_int  = -1
+
+erf.plot_file_1 = plt
+erf.plot_int_1  = 1
```
asalmgren, line 142: **`erf.plot_int_1 = 1` writes a plotfile every step.**

Combined with the missing `max_step`, that is 864,000 plotfiles of a 100x100x79 domain with 9 variables -- it will fill the regtest machine's disk long before the run ends. The comparable Katrina NoahMP+RRTMGP setup uses `erf.plot_int_1 = 1000`.

`erf.sum_interval = 1` on line 135 compounds the per-step I/O.

Exec/RegTests/WPS_Test/inputs_wps_ml
```diff
@@ -39,21 +41,8 @@ amr.max_level      = 1       # maximum level number allowed
...
-#amr.blocking_factor_y = 8
-#amr.blocking_factor_z = 68
-
-# Upper bound on box length in any direction
-#amr.max_grid_size_x = 64
-#amr.max_grid_size_y = 64
-
-#amr.refine_grid_layout_z = 0
+# Do not chop in z-direction
+amr.refine_whole_domain_dir = 2
```
asalmgren, line 45: **This hunk also removes `amr.n_error_buf_x = 2` and `amr.n_error_buf_y = 2`, not just the z entry.**

`amr.refine_whole_domain_dir = 2` replaces only the z behaviour, but `Source/main.cpp:50` defaults `amr.n_error_buf` to 0. So the horizontal tag buffer silently drops from 2 to 0 and the level-1 footprint now hugs the tagged cells exactly.

That changes the answer of an existing benchmarked regression test, for a reason the "Do not chop in z-direction" comment doesn't mention. If only the z behaviour was meant to change, `n_error_buf_x/y = 2` should be kept.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
```diff
@@ -0,0 +1,143 @@
...
+erf.variable_coriolis = true
+erf.coriolis_3d       = true
+
+erf.rayleigh_damping_type = FastImplicit
+erf.rayleigh_damp_W   = true
+erf.rayleigh_zdamp    = 5000.0
+erf.rayleigh_dampcoef = 0.2
+
+erf.radiation_model        = "RRTMGP"
+erf.profile_rad_int        = 100
```
asalmgren: **`erf.profile_rad_int` is set but `erf.rad.datalog` is not, so you pay the cost and get no output.**

`profile_rad_int` feeds `setDataLogFrequency`, and `datalog_int > 0` makes RRTMGP allocate the clear-sky flux/heating arrays at full `m_ncol` instead of `m_ncol_chunk` and compute clear-sky radiation on every radiation call (`Source/Radiation/RRTMGP/ERF_Radiation.cpp:395-420`, `1513-1514`). But the write is guarded by `rad[0]->hasDatalog()` at `Source/ERF.cpp:477`, which is false without `erf.rad.datalog` (`Source/Radiation/ERF_RadiationInterface.H:68-75`).

The deleted `inputs_wps_noahmp` had `erf.rad.datalog = "radiation"`. Either restore that line or drop `erf.profile_rad_int`.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
```diff
@@ -0,0 +1,143 @@
...
+
+
+# ============================================================
+# WRF Initialization
+# ============================================================
+
+erf.use_real_bcs   = true
+erf.real_width     = 5
+erf.init_type      = "WRFInput"
+erf.nc_init_file_0 = "wrfinput_d01"
```
asalmgren, line 70: **The referenced input data isn't in the tree, and the namelist being deleted won't match it.**

This points at `wrfinput_d01` / `wrfbdy_d01`, but `Exec/RegTests/WPS_Test` ships `wrfinput_chisholmview_d01` and `wrfbdy_chisholmview_d01_width1` -- and `erf.real_width = 5` (line 66) while the only checked-in bdy file is width 1. The PR description only promises to stage `namelist.erf` and `NoahmpTable.TBL` on the regtest machines; nothing says where these WRF files come from.

Two things to watch when staging them:

1. `start_datetime = "2024-08-05 00:00:00"` must match `SIMULATION START DATE` in `wrfinput_d01` exactly or ERF aborts (`Source/ERF.cpp:2998-3006`).
2. The `namelist.erf` being deleted has `ERF_SETUP_FILE_01 = "wrfinput_chisholmview_d01"` and `START_YEAR = 2023`. The uploaded namelist has to be regenerated for the new file and the 2024-08-05 date, not just copied over.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
```diff
@@ -0,0 +1,143 @@
+# ===== Problem Setup ==========
+erf.prob_name        = ABL
```
asalmgren, line 4: **`erf.prob_name = ABL` perturbs the real WRF initial condition.**

`Source/ERF.cpp:2528` skips `init_custom()` only for `InitType::NCFile`, not `InitType::WRFInput`, and `Exec/ERF_Prob.cpp:114` matches `"abl"`. So the `prob.KE_*` values on lines 3-5 are live: they add `r_hse*KE_0*(1-z/1000)^3` to `RhoKE_comp` and assign `RhoScalar_comp` on top of the WRF state (`Source/Prob/ERF_InitCustomPert_ABL.H:74-179`).

The sibling WPS tests use `erf.prob_name = "WPS"`, which matches no branch and leaves the real IC untouched (it just prints "does not add any state perturbations"). If a clean real-data init was intended, `"WPS"` is the right prob_name here.

Worth noting these are the same three `prob.KE_*` lines this PR removes from `inputs_wps_ml`, where they genuinely were inert.

Exec/RegTests/Radiation/inputs_radiation
```diff
@@ -68,8 +68,8 @@ erf.covmr                  = 1.5e-7
 erf.ch4vmr                 = 1.7e-6
 erf.o2vmr                  = 0.209
 erf.n2vmr                  = 0.7906 
-erf.rrtmgp_file_path       = /home/alattanz/git/data-files
-erf.rrtmgp_coeffs_sw       = rrtmgp-gas-sw-g112.nc #rrtmgp-data-sw-g224-2018-12-04.nc
-erf.rrtmgp_coeffs_lw       = rrtmgp-gas-lw-g128.nc #rrtmgp-data-lw-g256-2018-12-04.nc
+erf.rrtmgp_file_path       = ./  # Path to data-files
```
asalmgren: Nit: `./` is already the default -- `rrtmgp_file_path` initializes to `"."` at `Source/Radiation/RRTMGP/ERF_Radiation.H:297`, and the join at `ERF_Radiation.cpp:220` then produces `.//file.nc`. Harmless on POSIX, but the line could just be dropped. (The inline `# Path to data-files` comment is stripped correctly by ParmParse, so no issue there.)

Good change to get the hardcoded `/home/alattanz/...` path out of the file either way.

asalmgren (commented):
Review of the inputs changes. All parameter names check out against source; the findings below are about values and about two files outside the diff that this PR invalidates.

**Two items must be fixed before this can run nightly** (missing `max_step`, `plot_int_1 = 1`), and one is an unintended answer change to an existing benchmarked test (`inputs_wps_ml` tag buffer). Inline comments have the details.

Two more that can't be inline because the files aren't in the diff:

**`Exec/RegTests/WPS_Test/README:5` is now stale.** It says "Use `inputs_wps_noahmp` to run this test using the Noah-MP land surface model" and tells the reader to copy `NoahmpTable.TBL` into the directory. This PR deletes both `inputs_wps_noahmp` and `namelist.erf`, so anyone following the README hits a missing file. It should name `inputs_noah_rrtmgp` and point at the surviving copies in `Docs/sphinx_doc/namelist.erf` and `Docs/sphinx_doc/NoahmpTable.TBL`.

**The new nightly test isn't in the regression-test table** in `Docs/sphinx_doc/RegressionTests.rst` (the WPS_Test rows are at lines 239-245).

For the record, these all verified fine: `amr.refine_whole_domain_dir = 2` is valid and is the documented replacement for the `n_error_buf_z = 180` hack (`Docs/sphinx_doc/MeshRefinement.rst:389-407`); dropping `erf.nswbands/nlwbands/nswgpts/nlwgpts` is correct since those are read from the coefficient files now (`ERF_Radiation.cpp:229`), so switching to `g112`/`g128` needs no companion settings; `erf.rad_t_sfc` is a required `pp.get` for RRTMGP and is only a fallback when an LSM is present (`ERF_Radiation.cpp:711` vs. the `else` at 723), so it does not override NoahMP; `qc` is a valid plot var under `MoistNoCondensation` (`n_qstate_moist_size = 2`); BC strings are lowercased before matching (`Source/Initialization/ERF_InitBCs.cpp:237-269`); and `erf.most.*` is still accepted.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
AMLattanzi: Modified to `5 s` since CFL output showed max of `12 s`.

Exec/RegTests/WPS_Test/inputs_wps_ml
AMLattanzi: No benchmark to consider, this was propped up to test dynamic AMR. We may want to regtest this case though.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
AMLattanzi: Good catch. Removed in last commit.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
AMLattanzi: Comment updated.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
AMLattanzi: This is intentional. The MYNN PBL scheme needs an initial `TKE` profile. WPS therefore, is not appropriate. We could consider adding a prob type for real simulations with `TKE ` transport so that ONLY a `TKE` profile is built and users can't accidentially perturb other variables.

Exec/RegTests/Radiation/inputs_radiation
AMLattanzi: Removed the `/`.

Exec/RegTests/WPS_Test/inputs_noah_rrtmgp
AMLattanzi: Good catch. Corrected in last commit.

asalmgren (approved):

Issue #3618, opened 2026-08-08
Noah-MP never observes ITIMESTEP==1, so its first-call initialization is dead


WeiqunZhang:
**Severity:** medium · **Category:** correctness · **Fix order: 100 of 150** — fix this 100th.

<sub>Filenames are numbered in reverse fix order: `001` = fix last, `150` = fix first. This file is `051`.</sub>

**Location:** `Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP_Advance.cpp:412`

**Based on commit `4d7cf58636893f9`** (line numbers refer to that tree).

## The defect

**What is wrong** — time_to_fire() increments m_itimestep before the block loop, so the value mirrored into blk.io->itimestep is one greater than the call number; Noah-MP's DriverMain therefore never observes ITIMESTEP==1 and its first-call initialization block is dead code.

## Why it matters

Cold start: ReadLandMain (NoahmpReadLandMod.F90:221) leaves ITIMESTEP=1; time_to_fire bumps m_itimestep to 2 before the first DriverMain, so NoahmpDriverMainMod.F90:58-64 (EAHXY/TAHXY from actual forcing, CHXY=CMXY=0.1) never executes -- CHXY/CMXY stay at InitMain's 0.0 and the soil-update cycle mod(ITIMESTEP,SOIL_UPDATE_STEPS) is shifted by one substep.

## Suggested fix

`time_to_fire()` increments `m_itimestep` before the block loop, so the value mirrored into `blk.io->itimestep` is one greater than the call number and Noah-MP's `DriverMain` never sees `ITIMESTEP == 1` — its first-call initialization block never runs. Increment after the loop (or pass `m_itimestep` before incrementing) so the first call reports 1.

```diff
--- a/Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP_Advance.cpp
+++ b/Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP_Advance.cpp
@@ time_to_fire(): the counter is bumped BEFORE the block loop, so the value mirrored into
@@ blk.io->itimestep is one greater than the call number and Noah-MP never observes
@@ ITIMESTEP == 1 — its first-call initialization block is dead code.
-    ++m_itimestep;            // (in time_to_fire, before the loop below)
     ...
             // (4) Drive Noah-MP. Mirror the authoritative counter into the block first.
             blk.io->itimestep = m_itimestep;
             blk.io->DriverMain();
     ...
[...]

PR #4036, opened 2026-09-16
[Noah-MP] Address #3618, Noah-MP bump, and misc: Docs/, dev/ management

akashdhruv:
- [ ] Test on Aurora.


Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP_Advance.cpp
```diff
@@ -430,6 +430,10 @@ NOAHMP::Advance_With_State (const int& lev,
             read_results(mfi, blk, bx, gbx, cons_in);
         }
 
+        // Retire the firing index only after every block has consumed it, so the first
+        // firing reports ITIMESTEP == 1 and reaches Noah-MP's first-call init (#3618).
+        m_itimestep += 1;   // once per firing, in lockstep on every rank
```
asalmgren, line 435: This is the right placement, and I verified the three things it could have broken are all fine: the increment is outside the `MFIter` loop (so a rank owning no land boxes and a rank owning several both advance exactly once, keeping the broadcast counter in lockstep), the gate `(m_itimestep-1)*m_dtbl` still sees one increment per firing so firing times are bit-identical, and `Get_LSM_Step()` still returns the same integer it did before, so the `lsm_step` checkpoint file format is unchanged.

Two notes, neither blocking:

**1. Restart across the version boundary deserves a release note.** The stored integer is numerically identical before and after, but its *convention* flips: it was "index already consumed, +1", it is now "index the next firing will consume". Restarting a pre-PR checkpoint with this build therefore hands Noah-MP an `ITIMESTEP` one lower than the old code would have, shifting the `mod(ITIMESTEP, SOIL_UPDATE_STEPS)` soil-update phase and -- with the bumped submodule -- the derived calendar by one `DTBL`. That is the intended fix rather than a regression, but it does break bitwise continuation across the boundary, which the checkpoint code advertises.

**2. Minor robustness.** The increment now sits ~40 lines downstream of the `time_to_fire()` gate, with `collect_precip_sources`, `prepare_precip_snapshots` and the whole block loop in between. There is no early return in that span today, but if one is ever added the gate re-fires forever at the same `NOAH_time` -- a livelock rather than a crash, which is the unpleasant kind. A scope guard, or a comment marking the span as no-early-return, would be cheap insurance.

asalmgren (commented):
**The core fix is correct.** I traced it end to end: `NoahmpReadLandMod.F90:221` sets `ITIMESTEP = 1`, `ERF_NOAHMP_Init.cpp:208` picks that up into `m_itimestep`, and with the increment moved after the `MFIter` loop the first firing now mirrors `1` into `blk.io->itimestep` and reaches the `IF (NoahmpIO%ITIMESTEP == 1)` blocks in `NoahmpDriverMainMod.F90` (lines 60 and 149) -- exactly what #3618 asked for. The gate schedule `(m_itimestep-1)*m_dtbl` is unchanged so firing times are bit-identical; the increment stays outside the loop so ranks owning zero land boxes and ranks owning several all advance exactly once; and `time_to_fire` has no other caller. I also checked the legacy-checkpoint path: `lsm_step` and `noahmp_restart` were added in the same commit (`8059fd322`), so a checkpoint missing `lsm_step` also cold-initializes the land state, where `ITIMESTEP == 1` is the right thing to do.

The findings below are all in the surrounding material, not in the moved increment. Inline comments have the details; one more can't be inline because the file isn't in the diff:

---

### `Exec/CanonicalTests/Hurricanes/InputFiles/Katrina/namelist.erf:5` -- medium

`START_YEAR = 2013 / START_MONTH = 11 / START_DAY = 21`, copy-pasted from `Docs/sphinx_doc/namelist.erf` rather than set to Katrina's date. Katrina is August 2005, and `inputs_Katrina_adv_most_bulk_Morrison_MRF_smag2d_LSM_RRTMGP` sets no `erf.start_datetime` -- ERF takes its start from `wrfinput_d01`'s `SIMULATION_START_DATE`.

This was inert before: `NoahmpIO%YR` / `JULIAN` sat frozen at the `NoahmpIOVarInit` defaults for the whole run. The bumped submodule's new `NoahmpCalendarAdvance` derives the wall-clock date from these fields, so this case now runs Noah-MP phenology/greenness and `CAL_MON_DAY` on a late-November 2013 calendar while the atmosphere is in hurricane season -- a several-month seasonal error in LAI/vegetation.

`Exec/RegTests/WPS_Test/namelist.erf` (2023-08-17) looks plausible by comparison, but worth the same check. The shipped sample at `Docs/sphinx_doc/namelist.erf` is where the bad date is inherited from and could use a "must match your host start date" comment now that the field is load-bearing.

---

### Checked and clean

- Every reference to the deleted `dev/` specs is removed -- no dangling pointers in source or docs -- and the RST section underlines still match their shortened titles.
- The new calendar math is sound: multi-year rollover loops in both directions, whole seconds carried in `integer(i8)` with only the sub-second remainder through the real accumulator (protects single-precision builds), and `NoahmpYearLength` uses the full proleptic-Gregorian rule, which also fixes `CAL_MON_DAY`'s old `mod(julyr,4)`-only test.
- The `character(len=1)` -> `len=32` and `I1.1` -> `I0` changes in the three I/O modules are straight overflow fixes for `amr.max_level > 9`.

Submodules/Noah-MP
```diff
@@ -1 +1 @@
-Subproject commit e03243d2c5eac19b1c6a02a2c3ca12f03f92cceb
+Subproject commit 14b961ae5ca7dd5000f6b27135218386824dcd61
```
asalmgren: **This bump is not just a bump, and the PR description doesn't say so.**

`e03243d2c5..14b961ae5c` is three commits, including **+166/-56 in `NoahmpDriverMainMod.F90`**. Alongside the new calendar subsystem (`NoahmpCalendarAdvance`, `NoahmpYearLength`) it carries two answer-changing correctness fixes:

- `CAL_MON_DAY`'s `MONTH` array was `DATA`-initialized and therefore implicitly `SAVE`d, so one leap-year call left `MONTH(2)=29` set for the rest of the run.
- `NoahmpIO%ICE` is a scalar carried across the `ILOOP` iterations, so a land point following a sea-ice point inherited `ICE=1` and the first point saw `undefined_int`.

Both will move regression baselines. "Noah-MP bump, and misc" plus an unchecked "Test on Aurora" undersells that -- worth spelling out so the baseline churn isn't a surprise.

**The two halves of this PR are interdependent.** The new calendar computes the wall-clock date as `(ITIMESTEP-1)*DTBL` after the namelist start date, which is correct *only* with the ERF-side increment fix in this same PR. If either half is ever back-ported or reverted alone, every Noah-MP call is dated one `DTBL` off. Worth a line in the description so nobody splits them.

**One nit in the bumped code** (`NoahmpDriverMainMod.F90:153`): the "sea-ice at water point" diagnostic went from an unconditional `print*` to `if ((XICE(I,J) == 1.0) .and. (NoahmpIO%rank == 0))`. Each rank owns a different part of the domain, so an inconsistent `XICE`/`XLAND` pair anywhere off rank 0 is now silently ignored. The rank guard is right for the once-per-run calendar banner just above it, but wrong for a per-cell data anomaly.

Docs/sphinx_doc/CouplingToNoahMP.rst
```diff
@@ -193,9 +184,7 @@ types are identical *by construction*.
 
 To expose a new variable you therefore add a single line here and rebuild; the
 only remaining manual steps are the runtime wiring the generator does not own (the
-namelist guard for a namelist-read scalar, and optional NetCDF output). See
-``Submodules/Noah-MP/drivers/erf/dev/spec-add-coupled-variable.md`` for the full
-procedure.
+namelist guard for a namelist-read scalar, and optional NetCDF output).
```
asalmgren, line 185: **The submodule bump turns `namelist.erf`'s `START_*` fields into load-bearing user inputs, and that isn't documented anywhere.**

Before the bump, `NoahmpIO%YR` and `%JULIAN` sat frozen at their `NoahmpIOVarInitMod` defaults (2000, Jan 1) for the entire run. After it, `NoahmpCalendarAdvance` derives the wall-clock date from `START_YEAR/START_MONTH/START_DAY/START_HOUR/START_MIN`, which now drive phenology, `CAL_MON_DAY`, and the urban irrigation window. So those fields must agree with ERF's own start date (`erf.start_datetime`, or `SIMULATION_START_DATE` from `wrfinput`).

Nothing cross-checks them. The new Fortran says so itself -- *"Nothing cross-checks this against the host's own start date"* -- and only prints an advisory banner, so a mismatch is silent apart from one log line. An outright invalid date is now a hard `NoahmpIO_abort()`, which is also new behavior.

This PR edits this file but only to drop the `dev/` references; the calendar requirement should be documented here, per the repo convention of documenting user-facing behavior. The Katrina `namelist.erf` (see the review body) is what that gap costs in practice.

A hard check would be better than prose. `ERF_NOAHMP_Init.cpp:158-165` already asserts namelist `NSOIL` against `erf.lsm_nsoil` right after `ReadNamelist()`, which is the natural place. Note it would first need `start_year`/`start_month`/`start_day` added to the `@NoahmpMacro:Source m_noahmpio` block in `NoahmpIO.H-mc` -- they currently live only in `NoahmpIOVarType.F90-mc`, not the C++ ABI.

Docs/sphinx_doc/CouplingToNoahMP.rst
```diff
@@ -74,8 +74,6 @@ The ERF-side driver is split by concern across several files under
 
 -  **ERF_NOAHMP_IO.cpp**: The land plotfile and checkpoint/restart.
 
-Developer design specifications for these files live under
-**Source/LandSurfaceModel/Noah-MP/dev/** (start with ``dev/README.md``).
```
asalmgren: Nit: removing the two-line `dev/` pointer here leaves a double blank line before the next paragraph. Same thing happens at line 138, after the `WriteRestart`/`ReadRestart` bullet, and there's a stray extra blank before the `Working on the ERF driver` heading.

Docs/sphinx_doc/CouplingToNoahMP.rst
akashdhruv: Removed support for `START_*` in namelist, load those variables directly from WPS file.

Docs/sphinx_doc/CouplingToNoahMP.rst
akashdhruv: Fixed.

PR #4038, opened 2026-09-16
Normalize surface temperature handling across SurfaceLayer and radiation

pressel:
## Summary

This PR establishes a consistent surface-temperature contract across ERF:

* `SurfaceLayer::t_surf` stores **dry potential temperature (`θ`)**.
* Inputs supplied as **absolute temperature (`T`)** are converted to `θ` before entering `SurfaceLayer`.
* Consumers that require physical temperature explicitly convert `θ -> T` using the appropriate surface pressure.
* Noah-MP surface temperature remains an **absolute radiative temperature** and is not treated as a `SurfaceLayer` `θ` source.

This addresses the surface-temperature inconsistency described in #4012 and removes several places where absolute temperature had previously been interpreted as potential temperature.

### SurfaceLayer SST handling

Absolute-temperature SST sources are converted to `θ` before being stored in `t_surf`, including:

* prescribed/text SST,
* coupled SST,
* WRFInput SST/TSK initialization, and
* Metgrid SST/SKINTEMP initialization.

The text and coupled SST runtime paths now diagnose the physical surface pressure before applying the `T -> θ` conversion.

Dry configurations no longer access the moisture component when converting SST: `qv` is zero unless moisture is enabled and the conserved moisture component is present.

### Surface pressure and qsat

Surface-pressure handling used by the new conversions is face-consistent:

* x/y surface layers use the cell-centered pressure (`Δz = 0`);
* z-low uses the signed offset from the cell center to the lower W face;
* z-high uses the signed offset from the cell center to the upper W face.

The same pressure is used for both `θ -> T` and saturation-mixing-ratio calculations.

The qsat path validates conserved state before division. Invalid authoritative cells retain the collective failure diagnostic, while invalid halo state is ignored rather than turning an unused ghost value into a fatal error.

### Radiation

RRTMGP and TwoStream now consume surface temperature according to the same contract.

For RRTMGP, `SurfaceLayer` `θ` is converted to physical `T` using the diagnosed surface pressure and the configured `solverChoice.rdOcp`. Valid LSM absolute temperatures remain absolute temperatures, and the existing bottom-level invariant

```text
t_lev(0) = t_sfc
```

remains the single owner of the lower radiation boundary value.

TwoStream likewise converts `SurfaceLayer` `θ` to absolute temperature when needed while preserving the existing source precedence between LSM, prognostic SEB, SurfaceLayer, and scalar fallback temperatures.

### Metgrid

Time-dependent Metgrid SST/SKINTEMP values are converted from absolute temperature to potential temperature using the corresponding surface pressure.

When PSFC is present, it must now be finite and strictly positive. Missing PSFC continues to use the analytic fallback; invalid present values do not silently fall back. The analytic fallback also validates its thermodynamic inputs, including requiring positive `TLP`.

Metgrid initialization does not currently support a state layout decomposed in z. The code now checks the actual runtime `lev_new[Vars::cons].boxArray()` before terrain/SST/SKINTEMP accesses and aborts explicitly for that unsupported configuration rather than allowing flattened surface boxes to address the wrong 3-D FAB.

### Surface-layout / MPI handling

Review uncovered two additional decomposition issues in the new SST conversion paths.

The text and coupled SST conversions now operate only on the physical surface owner before `PlanarBoundary` propagates values to internal, periodic, and duplicate surface copies.

For coupled SST, the final planar-boundary exchange is the authoritative halo operation. A second coupled-SST fill after that exchange was removed because it could overwrite a correct neighboring value with a locally clamped donor value.

The coupled-SST regression therefore uses a **spatially varying, zero-ghost donor** so that neighbor exchange and local clamping produce distinguishable values.

### Restart compatibility

The checkpoint format now records the surface-temperature contract version. Restart logic distinguishes checkpoints written under the normalized contract from older checkpoints, and legacy Metgrid SST/TSK state is rejected where its previous semantics would be ambiguous.

### Documentation and diagnostics

Documentation and diagnostic descriptions were updated to reflect that `SurfaceLayer::t_surf`, `erf.most.surf_temp`, and the associated surface-heating quantity use potential-temperature semantics.

The `ABL_MOST_SST` gold result is updated because the old result encoded absolute SST being consumed directly as `θ`.

## Validation

Local validation on the final branch included:

* full DOUBLE unit suite: **715/715 passed**;
* focused serial SurfaceLayer tests: **11/11 passed**;
* SurfaceLayer MPI tests: **5/5 on 1 rank and 5/5 on 2 ranks**;
* focused SINGLE serial SurfaceLayer tests: **11/11 passed**;
* focused SINGLE parallel SurfaceLayer tests: **5/5 on 2 ranks**;
* NetCDF-enabled `erf_unit_tests` and `erf_exec` builds;
* focused NetCDF / RRTMGP / TwoStream / Metgrid / checkpoint tests: **16/16 passed**;
* normal Metgrid initialization smoke test;
* forced z-split Metgrid test confirming the expected early diagnostic; and
* the `ABL_MOST_SST` production regression.


Docs/sphinx_doc/Inputs.rst
```diff
@@ -3173,14 +3173,14 @@ the ones marked **Required** abort the run if they are not given.
 | **erf.most.pblh_calc**                | which scheme diagnoses the PBL height used by the *w*\*  | none, MYNN25,       | none             |
 |                                       | correction                                               | MYNNEDMF, YSU, MRF  |                  |
 +---------------------------------------+----------------------------------------------------------+---------------------+------------------+
-| **erf.most.surf_temp**                | prescribed surface temperature [K]; a positive value     | Real > 0            | -1.0 (not set)   |
-|                                       | selects the surface-temperature formulation              |                     |                  |
+| **erf.most.surf_temp**                | prescribed surface potential temperature [K]; a positive | Real > 0            | -1.0 (not set)   |
+|                                       | value selects the surface-temperature formulation         |                     |                  |
```
asalmgren: **Malformed RST table (medium).** This reflowed continuation line is 142 characters while every other row in the grid table is 141, so the third column's `|` is out of alignment. Running the file through docutils gives:

```
(ERROR/3) Malformed table. Right border not aligned or missing.
```

The effect is that the entire ~130-row `erf.most.*` / `erf.surface_layer.*` reference table is dropped from the rendered docs (and the Sphinx build errors). Fix is to delete one space before the closing `|`:

```rst
|                                       | value selects the surface-temperature formulation        |                     |                  |
```

Source/IO/ERF_ReadFromMetgrid.cpp
```diff
@@ -185,6 +183,9 @@ read_from_metgrid (int lev, int itime,
     if (!flag_hgt) {
         Abort("HGT_M was not found in " + fname + "; it is required to build the terrain.");
     }
+    if ((flag_sst || flag_tsk) && !flag_psfc) {
+        Abort("Metgrid SST/SKINTEMP requires PSFC in each forcing file so it can be normalized to potential temperature: " + fname);
```
asalmgren: **Removes an existing graceful path (medium).** `init_base_state_from_metgrid` (`Source/Initialization/ERF_InitFromMetgrid.cpp:1637-1645`) explicitly handles `flag_psfc == 0` by deriving `psurf` from the closed-form standard atmosphere, and `erf.metgrid_debug_psfc` exists precisely to run without a usable PSFC.

So a met_em set that carries SST/SKINTEMP but no PSFC - or a user running with `erf.metgrid_debug_psfc = true` - initialized fine before this PR and now aborts at startup. Consider falling back to the same analytic `psurf` for the theta conversion instead of aborting.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp
```diff
@@ -599,41 +601,37 @@ TwoStreamRadiation::advance (int lev,
...
-                        t_sfc_arr = lsm_ptr->const_array(mfi);
-                        has_t_sfc_field = true;
+                        lsm_t_sfc_arr = lsm_ptr->const_array(mfi);
+                        has_lsm_t_sfc = true;
                     }
                 }
-                if (!has_t_sfc_field && rad_choice.seb_prognostic_enable && rad_choice.seb_enable && m_t_sfc[lev]) {
-                    t_sfc_arr = m_t_sfc[lev]->const_array(mfi);
-                    has_t_sfc_field = true;
+                if (!has_lsm_t_sfc && rad_choice.seb_prognostic_enable &&
```
asalmgren, line 630: **Array-level precedence still short-circuits the SEB candidate (medium).** The comment above this block states the precedence is "valid LSM absolute temperature, valid prognostic SEB absolute temperature, SurfaceLayer theta, then the scalar fallback," and the stated point of moving resolution into the per-cell kernel is that "an LSM field can exist globally while carrying an undefined sentinel in an individual column." But `has_seb_t_sfc` is still gated on `!has_lsm_t_sfc`, so whenever an LSM exposes a `t_sfc` array the prognostic SEB candidate is never handed to the kernel at all.

Concretely: Noah-MP + `seb_prognostic_enable` over an ocean column where Noah-MP wrote `lsm_undefined` - the resolver skips the evolved SEB temperature and falls through to SurfaceLayer theta (or `erf.rad_t_sfc`), which is exactly the array-level precedence bug this PR set out to remove.

Suggest dropping the `!has_lsm_t_sfc` gate and letting `resolve_surface_temp_k` order the candidates per cell.

Source/Radiation/RRTMGP/ERF_Radiation.cpp
```diff
@@ -610,13 +612,15 @@ Radiation::mf_to_kokkos_buffers (iMultiFab* lmask,
...
     Real rad_t_sfc = m_rad_t_sfc;
+    Real rdOcp = m_rdOcp;
 
     for (MFIter mfi(*m_cons_in); mfi.isValid(); ++mfi) {
         const auto& vbx  = mfi.validbox();
         const int nx     = vbx.length(0);
         const int imin   = vbx.smallEnd(0);
         const int jmin   = vbx.smallEnd(1);
         const int offset = m_col_offsets[mfi.index()];
+        const int k_surface = vbx.smallEnd(2);
```
asalmgren: **Unused variable (low).** `k_surface` is added to this first `MFIter` loop of `mf_to_kokkos_buffers` but is never referenced here; the loop that actually uses it declares its own copy at line 748. It trips `-Wunused-variable`, harmless today only because the `-Werror` line in `.github/workflows/gcc-rrtmgp.yml` is commented out. Suggest removing the line.

asalmgren (commented):
Automated code review (Claude Code). The core theta/T contract work looks self-consistent: `pressure_at_surface` adds rho*g*dz in the right direction, the text-SST producer and `fill_qsurf_with_qsat` consumer use the identical pressure expression so they round-trip exactly, all `SurfaceLayer` / `make_two_stream_params` / `define_level` call sites were updated, and the removed `clamp_finite` / `is_finite_positive` / old `resolve_surface_temp_k` overload have no remaining references. The changed non-NetCDF translation units and the three GTest files compile clean with `-Wall -Wextra`, and the `ABL_MOST_SST` gold regrid (2 boxes -> 1) is harmless because `add_test_r` passes `--allow_diff_grids` to `fcompare`.

Five inline findings below. One more does not sit on a changed line, so it is recorded here:

**`Source/IO/ERF_Checkpoint.cpp:502` (low)** - this PR changes the meaning of two persisted fields with no checkpoint-version guard: `Tsurf<face>` (written/read at lines 502 and 1749) and, via the new metgrid ingestion, the `sst_lev` / `tsk_lev` arrays that `ERF_Checkpoint.cpp:1280-1285` restores verbatim rather than re-reading from NetCDF. Restarting a metgrid-driven run from a checkpoint written by current `development` feeds absolute SST (e.g. 290 K) into a field the new code interprets as potential temperature, so the MOST thermal boundary is off by the Exner factor - roughly 3 K near 900 hPa - with no warning. There is no version tag in the checkpoint to detect this.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1861,10 +1880,13 @@ SurfaceLayer::fill_qsurf_with_qsat (const int& lev,
...
                 auto RTh  = cons_arr(i,j,k,RhoTheta_comp);
                 auto Qv   = cons_arr(i,j,k,RhoQ1_comp) / Rho;
-                auto P_cc = getPgivenRTh(RTh, Qv);
-                P_cc += Rho*CONST_GRAV*deltaZ;
-                P_cc *= Real(0.01);
-                erf_qsatw(t_surf_arr(i,j,k), P_cc, q_surf_arr(i,j,k));
+                const auto P_cc = erf_surface_temperature::pressure_at_surface(
+                    Rho, RTh, Qv, deltaZ);
+                Real t_surface = t_surf_arr(i,j,k);
+                if (erf_surface_temperature::theta_to_temperature(
```
asalmgren, line 1987: **Silent skip leaves a stale surface humidity (low).** `fill_qsurf_with_qsat` previously always wrote `q_surf`; now `erf_qsatw` is only called when `theta_to_temperature` succeeds. A column with a non-finite or non-positive diagnosed surface pressure, or `t_surf <= 0`, silently retains whatever `q_surf` held - the `default_land_surf_moist` initial value, or a stale value from a previous step - and the run continues with a plausible-looking but wrong sea-surface humidity.

The same pattern appears in `fill_tsurf_with_sfc_sst` (~line 1782) and `fill_tsurf_with_coupled_sst`. Worth counting or asserting on the rejected cells rather than dropping them without a trace.

Docs/sphinx_doc/Inputs.rst
pressel: Thanks — good catch. Fixed the alignment by removing the extra space on that continuation row. The grid table is aligned again, and the documentation build is passing on the current head.

Source/IO/ERF_ReadFromMetgrid.cpp
pressel: Agreed. I removed the new hard requirement on PSFC and made the surface-temperature conversion use the same pressure policy as the existing Metgrid base-state initialization.

The production path now uses:

100000 Pa when erf.metgrid_debug_psfc=true;
the current forcing file's PSFC when present; and
the existing analytic standard-atmosphere pressure at the physical terrain height when PSFC is absent.

Both SST and SKINTEMP go through the same metgrid_surface_theta() helper, and I added coverage for file PSFC, missing PSFC, and the debug-pressure override.

Thanks for pointing out the compatibility regression.

Source/Radiation/TwoStream/ERF_TwoStreamRadiation.cpp
pressel: Thanks — I traced this through the prognostic SEB ownership path before changing the gate.

I kept the !has_lsm_t_sfc condition intentionally. When Noah exposes t_sfc on a level, the post-dycore prognostic SEB update is skipped for that entire level. In other words, m_t_sfc is not an evolved per-ocean-column fallback on a Noah level; it remains non-owned/default state there.

So in the concrete mixed land/ocean case:

a valid Noah t_sfc is used where available;
a Noah sentinel falls through to SurfaceLayer theta;
if SurfaceLayer is unavailable/invalid, the scalar radiation fallback is used.

Offering m_t_sfc as a candidate on that level would instead make the resolver consume a field that the prognostic SEB does not advance.

I clarified the comment in this block to make the level-wide ownership rule explicit. If we later make prognostic SEB ownership genuinely per-column alongside Noah, then I agree the resolver should receive both candidates and decide per cell.

Source/Radiation/RRTMGP/ERF_Radiation.cpp
pressel: Thanks — fixed. I removed the unused declaration from the first MFIter loop. k_surface is now declared only in the later surface-field loop where it is actually used.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
pressel: Agreed. I changed these conversions to fail fast rather than silently retain the previous/default value.

fill_qsurf_with_qsat, fill_tsurf_with_sfc_sst, and fill_tsurf_with_coupled_sst now set a device-side failure flag when the theta/T conversion is invalid, synchronize it, reduce the flag across MPI ranks, and abort with a path-specific diagnostic if any participating cell failed.

That prevents a bad pressure or temperature from leaving a plausible-looking stale q_surf or t_surf in the simulation.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1758,16 +1767,37 @@ SurfaceLayer::fill_tsurf_with_sfc_sst (const int& lev,
...
+        const auto cons_arr = cons_in.const_array(mfi);
+        const auto z_arr    = (z_phys_nd) ? z_phys_nd->const_array(mfi) :
+                                            Array4<const Real> {};
 
         ParallelFor(gtbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept
         {
             int is_land = (lmask_arr) ? lmask_arr(i,j,k) : 0;
             if (!is_land) {
-                t_surf_arr(i,j,k) = sfc_sst;
+                const Real rho = cons_arr(i,j,klo,Rho_comp);
```
asalmgren: **blocking** — these `klo` reads are out of bounds on a level whose grids are decomposed in z.

The iterator is over `*t_surf[lev]`, whose BoxArray is every grid flattened by `b.setRange(dir, sm_index)` with a `{1,1,0}` ghost vector. For a zlo face `sm_index == klo`, so `gtbx.smallEnd(2)` is *always* `klo` and the guard at `:1765` never fires. On a z-split level the flattened boxes are duplicates, and the copy whose local index maps to an upper 3-D box then evaluates `cons_arr(i,j,klo,…)` and `Compute_Zrel_AtCellCenter(i,j,klo,z_arr)` on a FAB covering, say, `k` in [nz/2, nz-1].

This hazard is new. On `development` the body of this loop touches only `t_surf_arr` and `lmask_arr`, both on the flattened layout, so the dead guard was harmless; the 3-D reads added here are what makes it bite.

Two correct patterns already exist in this file: `fill_qsurf_with_qsat` clips with `gtbx &= cons_in.fabbox(mfi.index())` and tests `m_planar_bndry[lev].is_surface_copy(mfi.index())`, and `compute_sfc_params_from_lsm_fluxes` iterates `cons_in` and tests the 3-D `validbox`. Either would do here.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1969,11 +2028,29 @@ SurfaceLayer::fill_tsurf_with_coupled_sst (const int& lev)
             int li = amrex::min(amrex::max(i, i_lo), i_hi);
             int lj = amrex::min(amrex::max(j, j_lo), j_hi);
 
-            if (has_valid && valid_arr(li,lj,k) == 0) { return; }
-
-            t_surf_arr(i,j,k) = coupled_sst_arr(li,lj,k);
+            if (!has_valid || valid_arr(li,lj,k) == 0) { return; }
+
+            const Real rho = cons_arr(li,lj,klo,Rho_comp);
```
asalmgren, line 2135: **blocking** — same defect as in `fill_tsurf_with_sfc_sst`: the guard at `:2004` is a no-op on the flattened `t_surf` layout, so on a z-split level these `klo` reads land outside the 3-D FAB.

Clamping to `li`/`lj` guards only the lateral indices; the `klo` index is untouched by it. A coupled REMORA run with `erf.max_grid_size_z` below `nz` reads garbage here &mdash; and because the garbage flows into `pressure_at_surface`, it can then trip the new `Abort` at `:2051`, reporting an SST conversion failure when the real fault is the out-of-bounds read.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1861,13 +1894,25 @@ SurfaceLayer::fill_qsurf_with_qsat (const int& lev,
...
+                    amrex::Gpu::Atomic::Max(conversion_failed, 1);
+                }
             }
         });
     }
+    amrex::Gpu::streamSynchronize();
+    int conversion_failed_host = d_conversion_failed.dataValue();
+    amrex::ParallelDescriptor::ReduceIntMax(conversion_failed_host);
+    if (conversion_failed_host != 0) {
+        amrex::Abort("SurfaceLayer fill_qsurf_with_qsat: failed to convert the authoritative "
```
asalmgren, line 2000: **medium** — this abort escalates what used to be a harmless ghost-cell value into a killed run.

`gtbx` is deliberately intersected with `cons_in.fabbox(...)` above, so state ghost cells are visited. For a cell with `rho_theta == 0`, `pressure_at_surface` returns `getPgivenRTh(0,qv) + rho*g*dz`, whose first term is 0; if `delta_z` is also 0 the pressure is 0, `valid_pressure` is false, `theta_to_temperature` returns false, and the reduced flag stops the job. Before this PR such a cell simply got a nonsense `q_surf` that nothing downstream consumed.

That this is reachable is shown by the PR itself: `Tests/Unit/BoundaryConditions/ERF_GTestSurfaceLayer.cpp` had to add an explicit pre-fill of the state ghost cells, with the comment "initialize those cells so this test exercises qsat rather than an unrelated invalid-density path."

Treating an invalid cell as a no-op &mdash; which is what `get_lsm_tsurf` now does &mdash; or restricting the abort to the valid box would keep the diagnostic without the new failure mode.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -1861,13 +1894,25 @@ SurfaceLayer::fill_qsurf_with_qsat (const int& lev,
                 auto Rho  = cons_arr(i,j,k,Rho_comp);
```
asalmgren: **low** — on a lateral (x/y) surface layer this pressure is badly wrong, and the PR now propagates that error into the temperature.

For `dir != 2` the box is grown over the whole column, so `k` sweeps every level and `deltaZ` at `:1892` is the height of cell `k` above the terrain rather than an offset from the surface. `P_cc = p(i,j,k) + rho*g*z(k)` is then inflated by roughly 36 kPa at z = 5 km. The `myhalf*dz` no-terrain branch has the mirror-image problem.

The inflated pressure was pre-existing as the `erf_qsatw` argument, but it is now also fed to `theta_to_temperature`, so `t_surface` itself is wrong at every k of a lateral face.

asalmgren (commented):
Review of the surface-temperature normalization. I checked each finding against this head (`cd1465f9a`) and against `development`, to separate what the PR introduces from what it inherits.

The three highest-value findings are all the same shape, and it is a shape this repo has just been working through in #4043 and #4044: a `MFIter` over a z-collapsed BoxArray whose local index maps to a 3-D FAB that need not contain the surface plane. Moving the surface-temperature conversion into these loops turns three previously 2-D-only kernels into 3-D reads at `klo`/`k_surface`, and the `gtbx.smallEnd(2) != klo` guard that looks like it covers this is a no-op on the flattened layout. Two of them are in `SurfaceLayer` and one in the metgrid SST/TSK path.

Beyond those: `metgrid_surface_pressure` hands back an unvalidated file PSFC into a `(p_0/psfc)^rdOcp`, and the new `fill_qsurf_with_qsat` abort fires on state ghost cells that were previously harmless &mdash; the PR's own unit test had to be adjusted around that.

Things I checked and think are right: the flipped `!has_valid || valid_arr(...) == 0` matches the documented `update_coupled_sst_ptr` contract; dropping the `m_t_sfc` fallback in `TwoStreamRadiation::advance` is behavior-preserving, since `fill_or_copy_seb_field` already fills it with `rad_choice.rad_t_sfc`; removing `clamp_finite` is an improvement, because the old code clamped the `lsm_undefined` sentinel to 1.0; and `read_contract_version`, the `SST_0_H`/`TSK_0_H` legacy scan, the three constructor call sites, `m_rdOcp` initializer ordering and the Pa/mb split in `erf_qsatw(..., P*0.01)` all look correct.

Seven comments inline: two blocking, four medium, one low.

*Reviewed with Claude Code.*

Source/Initialization/ERF_InitFromMetgrid.cpp
```diff
@@ -411,11 +415,22 @@ ERF::init_from_metgrid (int lev)
...
+                const Array4<const Real> psfc_arr = flag_psfc ? NC_psfc_fab.const_array() :
+                                                               Array4<const Real>{};
+                const Array4<const Real>& z_nd_arr = z_phys_nd[lev]->const_array(mfi);
                 ParallelFor(gtbx, [=] AMREX_GPU_DEVICE (int i, int j, int) noexcept
                 {
                     int li = min(max(i, i_lo), i_hi);
                     int lj = min(max(j, j_lo), j_hi);
-                    dst_arr(i,j,0) = src_arr(li,lj,0);
+                    const Real z_sfc = Real(0.25) *
+                        (z_nd_arr(li,lj,k_surface) + z_nd_arr(li+1,lj,k_surface) +
```
asalmgren, line 444: **medium** — `z_nd_arr` is indexed at `k_surface` from an iterator over `ba2d[lev]`.

`sst_lev[lev][itime]` lives on `ba2d[lev]`, i.e. `grids[lev]` flattened with `setRange(2,0)`, while `z_phys_nd[lev]` is the 3-D nodal field. On a z-split level the duplicate 2-D box maps by local index to a nodal FAB that does not contain `k_surface`, and these four reads go outside it.

This function already guards exactly this case in the grid-integrity check above, at `:367`:

```cpp
if (klo_nd < vbx.smallEnd(2) || klo_nd >= vbx.bigEnd(2)) { continue; }
```

Source/Initialization/ERF_InitFromMetgrid.cpp
```diff
@@ -431,11 +446,22 @@ ERF::init_from_metgrid (int lev)
...
```
asalmgren, line 488: Same out-of-bounds `z_nd_arr(…, k_surface)` read as in the SST loop above &mdash; see the comment on `:426`.

Source/Initialization/ERF_MetgridUtils.H
```diff
@@ -15,6 +15,57 @@
...
+                          const amrex::Real z_sfc,
+                          const amrex::Real P00,
+                          const amrex::Real T00,
+                          const amrex::Real TLP)
+{
+    if (debug_psfc) {
+        return amrex::Real(100000.0);
+    }
+    if (flag_psfc == 1) {
+        return file_psfc;
```
asalmgren: **medium** — when `flag_psfc == 1` this returns the file value with no validity check, and the new SST/TSK conversion adds no guard of its own.

`metgrid_surface_theta` then evaluates `T*(p_0/psfc)^rdOcp`, so a met_em file carrying a fill value or a missing-data constant in PSFC at some cells silently yields `Inf`/`NaN` in `sst_lev`/`tsk_lev`, and that propagates into `t_surf` and the MOST fluxes.

The base-state path this helper was factored out of still protects itself &mdash; `AMREX_ALWAYS_ASSERT(psurf > zero)` at `ERF_InitFromMetgrid.cpp:1665` &mdash; while the new path does not, and it runs over grown boxes at every forcing time. Worth putting the positivity check inside the helper so both callers get it.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
pressel: Fixed in the follow-up. The text-SST path now restricts the conversion to the physical z-low surface copy via is_surface_copy before any 3-D state/terrain reads, clamps tangential ghost sampling to the physical source box, and then propagates the result with the planar-boundary exchange. Added z-split regression coverage for this ownership path.

Source/Initialization/ERF_MetgridUtils.H
pressel: Fixed. When PSFC is present, the shared helper now requires the value to be finite and strictly positive; an invalid present value does not silently fall back to the analytic profile. Missing PSFC still uses the analytic fallback, the explicit debug-pressure override remains supported, and NetCDF fill/missing/default metadata are canonicalized before the conversion path. The analytic fallback also rejects non-positive TLP, with regression coverage for the negative-TLP case.

Source/Initialization/ERF_InitFromMetgrid.cpp
pressel: Fixed by the same early runtime-layout guard. Because the actual lev_new[Vars::cons].boxArray() is checked before SST/SKINTEMP and terrain reads, the duplicate flattened SKINTEMP copy can no longer index a 3-D nodal FAB at an out-of-range k_surface. The forced-z-split runtime test covers the early rejection.

Source/Initialization/ERF_InitFromMetgrid.cpp
pressel: Fixed. The Metgrid admission check now inspects the actual runtime state layout, lev_new[Vars::cons].boxArray(), before terrain/SST/SKINTEMP processing. A z-split state layout is rejected explicitly with the documented diagnostic rather than reaching the flattened-surface access. The forced-z-split runtime test confirms the expected early abort.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
pressel: Fixed. Pressure handling is now face-consistent: x/y surface layers use the cell-centered pressure (delta_z = 0), z-low uses the signed offset to the lower W face, and z-high uses the signed offset to the upper W face. The same diagnosed pressure is used for both theta-to-temperature conversion and qsat. The z-high z-split regression uses a nonuniform vertical metric to check the signed upper-face result and duplicate-copy agreement.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
pressel: Fixed. The qsat path now validates rho, rho_theta, and rho_qv before division. Invalid authoritative cells still set the collective conversion-failure flag, while invalid halo cells are ignored without setting that flag, so a bad state ghost can no longer abort the run. The lateral x-low regression exercises this policy directly.

PR #4039, opened 2026-09-16
[WIP]: Issue 4037

### LINKED ISSUE #4037: [Bug]: PBL QKE Source Error
# Summary

PBL vertical gradients at the first cell are inconsistent under a surface_layer BC

Version: main @ 42934d88703346b4f608ec307e395652005c393e

## The issue

ComputeVerticalDerivativesPBL (Source/PBL/ERF_PBLModels.H:281) selects a one-sided, boundary-to-center stencil at k == izmin only when the relevant ext_dir_on_zlo flag is set. With zlo.type = surface_layer none of those flags is set, so the first cell falls through to the interior centered stencil and differences against a ghost cell that carries no surface information (FOEXTRAP value).

For potential temperature this yields a gradient that is exactly half the one-sided value. For velocity the stencil collapses to a first-order difference located at z_{klo+1/2} rather than at the cell center.

Where the flags come from

Source/Initialization/ERF_InitBCs.cpp translates zlo.type = surface_layer into:

cell-centered variables (including RhoTheta): ERFBCType::foextrap (~line 733)
tangential u, v: ERFBCType::hoextrap; only the wall-normal component gets ext_dir (~line 525)

So in ComputeQKESourceTerms (Source/PBL/ERF_PBLModels.H:353) the booleans c_ext_dir_on_zlo, u_ext_dir_on_zlo, v_ext_dir_on_zlo are all false, and the k == izmin branches of ComputeVerticalDerivativesPBL are unreachable at zlo. dz_inv.dirichlet_lo is never used there.

Resulting discretization at k = klo

With dz_inv.interior = 1/(2*dz) (PBLDerivativeDzInv_N, ERF_PBLModels.H:416):

Potential temperature. foextrap sets θv(klo-1) = θv(klo), giving
```
dθv/dz|klo = (θv_{klo+1} - θv_{klo}) / (2*dz)
```

This is half the one-sided difference, and contains no information about the surface heat flux.

Velocity. hoextrap at zlo is linear extrapolation, u(klo-1) = 2*u(klo) - u(klo+1) (Source/BoundaryConditions/ERF_BoundaryConditionsXvel.cpp:258). Substituting into the centered stencil, the ghost cancels:

```
du/dz|klo = (u_{klo+1} - u_{klo}) / dz
```

No factor-of-two error, but the natural location of this estimate is z_{klo+1/2}, not z_{klo}, and it still carries no dependence on u*.

## Magnitude

For a neutral log layer the first two cell centers sit at z1 = dz/2 and z2 = 3*dz/2, so the ratio of the discrete velocity gradient to the exact one at z1 is
```
[ln(z2/z1) / (z2 - z1)] / (1/z1) = ln(3)/2 ≈ 0.55
```

independent of dz. Shear production scales as the square, so the first-cell value is roughly 30% of the correct one at every vertical resolution. The halved θ gradient puts first-cell dθv/dz at about ln(3)/4 ≈ 0.27 of the log-layer value.

## Why the BC itself is not the problem

The surface layer is flux-based: SurfaceLayer::compute_SurfaceLayer_bcs (Source/BoundaryConditions/ERF_SurfaceLayer.cpp:803) takes the state as Vector<const MultiFab*> and writes only into Tau_lev / hfx / qfx; it never populates ghost cells. The diffusion operator then substitutes those fluxes directly at the wall face (Source/Diffusion/ERF_DiffusionSrcForState_N.cpp:210, with zflux = 0 for RhoKE), so foextrap is the correct logical BC for the transport terms — as the comment in ERF_InitBCs.cpp notes. The PBL gradient stencil is the one consumer that reads the ghost cell as if it were physically meaningful.

## Scope

ComputeVerticalDerivativesPBL is called from:

Source/Diffusion/ERF_AddQKESources.H (via ComputeQKESourceTerms) — shear and buoyancy production
Source/PBL/ERF_ComputeDiffusivityMYNN25.cpp:146 — also feeds the buoyancy length scale l_B and the stability functions, so Km/Kh at klo inherit the bias
Source/PBL/ERF_ComputeDiffusivityMYNNEDMF.cpp:4332
Source/PBL/ERF_ComputeDiffusivityMRF.cpp:1027,1085
Source/PBL/ERF_ComputeDiffusivityYSU.cpp:235, ERF_ComputeDiffusivityYSUNew.cpp:1799
Source/PBL/ERF_ComputeDiffusivityMYJ.cpp:230

MYNN25 and MYNNEDMF both require zlo.type = surface_layer (TurbChoice::pbl_requires_surface_layer, Source/DataStructs/ERF_TurbStruct.H:352), so this is the only path those models take, not an edge case.

Nothing downstream corrects the first cell: the dirichlet_k pin in Source/TimeIntegration/ERF_SlowRhsPost.cpp:716 is gated on the RANS k-equation and does not apply to MYNN.

## Suggested fix

Add a surface-layer branch at k == izmin that builds the gradients from MOST rather than from ghost cells:

```
du/dz = (u*/(κ z)) φm(z/L) * (u_klo / |U_klo|)      (similarly for v)
dθv/dz = (θ*/(κ z)) φh(z/L)
```

The ingredients are already local in the diffusivity routines — u_star_arr, t_star_arr, q_star_arr, tm_arr and l_obukhov are computed within a few lines of the derivative call in both ERF_ComputeDiffusivityMYNN25.cpp and ERF_ComputeDiffusivityMYNNEDMF.cpp. The QKE path needs more plumbing: ERF_AddQKESources.H currently forwards only tm_arr(i,j,0), so u* and θ* would have to be threaded through Source/Diffusion/ERF_SetupDiff.H into DiffusionSrcForState_{N,S,T}.

A smaller interim step is to pass a surface_layer_on_zlo flag into ComputeVerticalDerivativesPBL and take the existing one-sided branch with dz_inv.dirichlet_lo, which at least removes the factor of two in the θ gradient.

It may also be worth applying the usual MYNN lowest-level QKE surface condition (proportional to B1^(2/3) u*² in the WRF implementation) instead of relying on resolved first-cell production. No such condition is present in ERF today.

### Steps to reproduce

N/A

### Expected behavior

N/A

### Git commit hash

42934d88703346b4f608ec307e395652005c393e

### Environment

N/A

### Relevant logs or output

```shell

```

AMLattanzi:
# Summary
This PR attempts to resolve #4037 .


hgopalan:
@AMLattanzi, I merged this branch (de0de9560) onto #4051 to test it and found a problem in the MRF helper on stretched and terrain-fitted meshes.

In `ApplySurfaceLayerGradientsAtKsrfMRF` (`ERF_ComputeDiffusivityMRF.cpp` L45-47):
```cpp
const Real z_sfc = (use_terrain_fitted_coords)
                 ? Compute_Zrel_AtCellCenter(i, j, ksrf, z_nd_arr) : Real(0);
const Real zrel  = amrex::max(zval - z_sfc, Real(1.0e-4));
```
`Compute_Zrel_AtCellCenter` already returns the cell-centre height above the column's bottom node (`z_cc - z0_cc`). So on the terrain-fitted path `zval` is already the height above the surface, and `z_sfc` here is the first cell's own centre height. At `k == ksrf` that makes `zrel = 0`, which is clamped to 1e-4 m, and the MOST gradients come out about 10^4 times too large. This is the same first-level offset that #4051 (fix 5) took out of the K-profile's linear factor.

**GABLS1** (`Exec/CanonicalTests/ABL/mrf_stable_gabls`, stretched mesh with `erf.initial_dz = 4`, which takes the terrain-fitted path). A debug print in the first cell at the first steps shows `zval = 2`, `dudz = 17180` s⁻¹ (MOST gives about 0.9) and `Km = 15107`, clamped to the 300 m²/s cap. Over the full 9 h:

| | first-cell Kmv | max Kmv | u* |
| --- | --- | --- | --- |
| #4051 without this branch | 0.15 | 2.65 | 0.383 |
| with this branch | 399 (the cap, from the first 10 min on) | 399 | 0.341 |
| with this branch and `zrel = max(zval, 1e-4)` | 0.40 | 2.71 | 0.377 |

With the helper as it is, the same deck run anelastic with explicit diffusion at dt = 1 s aborts (`RhoTheta is negative at 0 0 0`). Uniform flat meshes take the `z_sfc = 0` branch, so they are unaffected.

**Suggested fix:** use `zval` as the height, `const Real zrel = amrex::max(zval, Real(1.0e-4));`. The other helpers look fine: MYJ, MYNN25, MYNNEDMF and YSU pass `zval` directly, and YSUNew subtracts `zib`, which is right for its absolute heights.

PR #4042, opened 2026-09-16
Abort when acoustic substepping runs on grids split in z

hgopalan:
## The bug

The acoustic substep (`erf.substepping_type = Implicit`, the default for compressible runs) solves for `w` with one tridiagonal system per box column. Each column is closed with a Dirichlet row at the box's own `lo.z` and `hi.z+1` (`ERF_MakeFastCoeffs.cpp`, "If at the bottom of the grid"; the `w` solves in `ERF_Substep_{T,NS,MT}.cpp`). This dates from 5bbcef32e ("generalizations for multiple grids", #1762, 2024-08-28).

On grids decomposed in the vertical, every internal box face therefore acts as a rigid lid, and the answer depends on the decomposition:

| development binary | `w` difference, split vs one box (relative) |
| --- | --- |
| `ABL_MOST_WOA_ZSplit_NoSub` deck (12 boxes, split at k = 16, 32), 10 steps, dt = 0.02 s, `substepping_type = None` | 2.8e-14 |
| same deck, `substepping_type = Implicit` | 3.4e-5 |
| Canonical RANS convective deck, 10 boxes in z, dt = 2 s, one step | 39 (Smagorinsky) |

With dt = 2 s the split k-eqn and Deardorff runs reach negative density by step 2; the single-box runs do not. `ABL_MOST_WOA_ZSplit_NoSub.i` already turns substepping off, citing this per-box column assumption.

## The fix

`ERF::define_column_kextent` already aborts when implicit vertical diffusion runs on grids split in z (#3998). It now also aborts when `substepping_type = Implicit` at that level, both at start-up and at a regrid that creates z-stacked grids. The message names `erf.substepping_type = None` as the way to run such a case, and `Inputs.rst` gets a matching note next to the implicit-diffusion note.

This makes the defect loud. A column solve that spans boxes would make these cases run correctly, but that is a larger change to the fast integrator (and #3471 is currently rewriting those files).

## Tests

- **`ABL_MOST_WOA_ZSplit_Substep_Abort`:** the `ABL_MOST_WOA_ZSplit_NoSub` deck with `erf.substepping_type=Implicit` must stop at start-up with the new message.
  - It uses a new `Tests/RunExpectedAbort.cmake` / `add_test_expected_abort`: the run must exit non-zero, and its output must contain every expected string, so a crash for another reason does not pass.
  - It is inside the existing `ERF_ENABLE_MPI AND NOT WIN32` block.
  - It passes with this change and fails with the development binary, where the run goes ahead.
- **`Terrain2Lev_BTF_ZSplit`:** now sets `erf.substepping_type = None`. The test compares only the mesh at step 0, and its split run would otherwise abort.

## Which existing inputs are affected

ERF splits a column in z only when an input asks for it (`amr.max_grid_size_z`, or a single `amr.max_grid_size`, smaller than `n_cell` in z; see Inputs.rst). To find affected inputs:
- **Decks run:** every input file under `Exec/` and `Tests/test_files/` (428) was started for zero steps with this binary, one at a time. None stopped at the new abort, apart from `Terrain2Lev_BTF_ZSplit` above.
- **Not reached:** 64 decks stopped earlier for unrelated reasons (particles, NetCDF, or FFT not built here), and 4 timed out.
- **Not started:** 17 real-size decks (over 2e6 cells) were left out for memory. They use the default grid layout, which does not split in z.

## Merge of development (10cc4e503)

development was merged in to resolve a conflict in `Tests/CTestList.cmake`: #4043 gave `ABL_MOST_WOA_ZSplit_NoSub_BoxParity` a `DATALOG` argument on the lines where this branch adds `ABL_MOST_WOA_ZSplit_Substep_Abort`. Both are kept. No other file conflicted, and the change of this PR against development is unchanged (5 files, 101 lines added).

development gained two input decks since this PR was opened (`Closure_BoxParity`, `PBLH_ZSplit`, from #4046 and #4044). Neither reaches the new abort: the first does not split in z, the second sets `erf.substepping_type = None`. Both pass below, so the sweep above still holds.

## Verification (on the merged head)

| Check | Result |
| --- | --- |
| Release build (MPI, FFT, all warnings) | no warnings |
| Debug build (MPI, no FFT: the CI configuration, bounds-checked) | no warnings |
| Unit tests, Release | 572 of 572 |
| Parallel tests, 1 and 2 ranks, Release and Debug | 47 of 47 |
| Release `ctest -L regression` | 121 of 124 (see below) |
| Debug `ctest -R "ZSplit\|Abort\|SelfTest"` | 11 of 11, including `ABL_MOST_WOA_ZSplit_Substep_Abort` and `PBLH_ZSplit_BoxParity` |
| Debug `ctest -R "BoxParity\|Tiling\|Restart"` | 28 of 28 |
| Mutation: abort disabled (`if (false && ...)`), Debug | `ABL_MOST_WOA_ZSplit_Substep_Abort` fails (the run goes ahead); passes again with the source restored and rebuilt |
| Single precision, `-Wshadow`: `ERF_MakeNewArrays.cpp` | no warnings |
| MPI off | configures; the abort and box-parity tests are not registered |
| Sphinx | no warnings |
| codespell, tab and trailing-whitespace scripts, `git diff --check` | clean |
| GPU lint scripts (device constants, private device lambdas) | no hits; no device code in this PR |
| Submodule entries | all six are gitlinks (mode 160000) |

The three Release regression failures are not the new abort:
- `ABL_MOST_WOA_ZSplit` (anelastic, no substepping) and `ABL_MOST_Cloudchamber` run to completion and fail their gold comparisons on this machine (Mac arm64), as they did before the merge with the same error norms as the development binary.
- `Closure_BoxParity_MYNNEDMF` (new with the merge) stops in its one-box run with the invalid-operation trap inside `ComputeDiffusivityMYNNEDMF`, the trap reported in #4046. It fires in Release on this machine only and passes in Debug; the deck is not split in z.

**Not run:** CUDA, HIP and SYCL compiles (no device code is touched) and a GNU Make build (no source files are added).
**Not applicable:** no new inputs, no new plot variables, no checkpointed state.

Found while addressing the review of #4017.


asalmgren:
@hgopalan -- I am going to close this after all -- I think now that we have the default to make grids not decompose in z we can just have a simple test that enforces no decomposition in z if the substepping type is Implicit.

PR #4044, opened 2026-09-16
PBL height: scan whole columns on grids split in z

hgopalan:
## The bug

The MYNN25 PBL-height estimator (`erf.most.pblh_calc = MYNN25`, `ERF_PBLHeight.H`) sets each tile's z range to the whole domain (`gtbx.setSmall(2,domain.smallEnd(2)); gtbx.setBig(2,domain.bigEnd(2));`, since f73c64bbc, #1756). It then writes the planar `pblh` of that box.

On grids whose boxes do not all span the domain in z, this goes wrong in three ways:
- **Reads outside the box:** a box stacked on another reads `cons` outside its own data, both at `k = 0` (the maximum TKE) and at every level below the box. A box that stops below the top of the domain (a refined level of partial height) is scanned past its top.
- **Unsynchronized copies:** each stacked box writes its own copy of the planar field, and those copies are never synchronized.
- **Levels away from the ground:** a box with no ground below it has no column to scan.

Consumers read `pblh` through their 3D `MFIter` index: the k-eqn length cap (`erf.rans_lscale_from_pblh`), the 2D plotfile and IBSEB. So the PBL height they see depends on the decomposition.

On the test deck below, after 40 steps, the split run differs from the single-box run by 4.0e-2 in KE and by 0.97 in the eddy viscosity (relative). With `erf.rans_lscale_from_pblh = false` the two runs agree to the last bit. The dependence is in the PBL height, not in the closure.

## The fix

- **Estimator:** it scans the cells its box holds rather than the domain, so the last `k+1` read ends in the box's top ghost cell, and it asserts that the box starts at the ground. Full-height boxes are scanned exactly as before.
- **Grids whose boxes all start at the ground** (full height or not) go straight to the estimator, unchanged.
- **Any other grids** (boxes stacked in z, boxes aloft) go through columns: the runs of cells that start at the ground, each joined into one box with `join_boxes_stacked_in_z` (the helper of the fine BTF mesh, now declared in `ERF_TerrainMetrics.H`). Every valid cell of a column is a valid cell of the level, so nothing needs a fill value.
  - **Copy in:** `Rho`, `RhoTheta`, `RhoKE` and the span of the moisture species that enter theta_v, with one ghost cell; `z_phys_cc` and the land mask likewise. Ghost cells go first, valid cells second, so every cell the level owns comes from the box that owns it.
  - **Copy out:** onto every planar box; planar cells over which no box reaches the ground hold zero, the estimator's value for a height it did not find. A level entirely aloft gets zero everywhere.
  - **Cost:** the columns and their DistributionMapping (each column on the rank that owns its lowest corner cell) are built when the grids of the level change, not per call. Each PBL-height update on such grids copies the components above onto the columns.
- **EB:** the PBL height lives on the 3D grids and is written into their k = 0 plane, so the run aborts, naming the input, if some box does not start at the bottom of the domain.

## Tests

`PBLH_ZSplit_BoxParity` uses `add_test_box_parity` and a new deck, `Tests/test_files/PBLH_ZSplit`.
- **Deck:** the Canonical RANS convective case with the PBL-height length cap, compressible with explicit diffusion and without acoustic substepping. Both the implicit vertical solve and the substep need unsplit columns (#3998, #4042).
- **Comparison:** one box against 10 boxes of 4 x 8 x 20 (split at k = 20, 40, 60, 80, read from `Level_0/Cell_H`), `rel_tol 1e-9`.
- **Results:** identical plotfiles with this change; without it the test fails (KE 4.0e-2, Kmv 0.97 relative). Kmv spans 0.034 to 14.6 m²/s in the split run, so the comparison is not between constants.

Unit tests (`erf_parallel_tests`, `SurfaceLayerParallel.*`, 1 and 2 ranks), each against an exact column (theta_v steps by 2 K, dz = 1 m, no TKE, so the height is k_jump + 0.125 m):
- `PBLHeightScansWholeColumnsOnZSplitGrids`: one box and 16x16x2 boxes (two stacked per column) both give 2.125 m on every planar copy. The per-box path gives 0 to 7.87 m.
- `PBLHeightStopsAtTheTopOfPartialHeightGrids`: grids over the lower half of the domain. A jump between the top cell and its ghost cell is found, a jump above is not. Scanning to the top of the domain aborts in Debug: `(-1,-1,3,1) is out of bound (-3:18,-3:18,-1:2,0:4)`.
- `PBLHeightOnUnevenStacksAndBoxesAloft`: a box stacked on three quarter boxes with a different footprint, partly over a quarter with no ground box. 3.125 m where the column reaches the ground and holds the jump, 0 elsewhere, on every copy. One column per surface box extended to the top of the domain returns 1e150.
- `PBLHeightIsZeroOnGridsEntirelyAloft`: zero everywhere; building a DistributionMapping on the empty column set trips ``! boxes.empty()`` in Debug.
- `PBLHeightSeesMoistureOnZSplitGrids`: the jump made by water vapour alone. With the moisture span not copied the split layout gives 0.
- The harness takes an explicit BoxArray for grids that do not cover the domain.

**Coverage:** flat ground only. The terrain path (`z_phys_cc`) is copied the same way but is not covered by a test here. The EB abort was checked by hand (`inputs_most_nbl_EB` at 8 x 8 x 261 with `pblh_calc = MYNN25`: aborts at start-up when split in z, steps normally otherwise); a CTest entry needs `add_test_expected_abort` from #4042.

## Verification

- **Release** (MPI, FFT, all warnings): no warnings.
  - `ctest -L regression`: 120 of 123 pass. `ABL_MOST_WOA_ZSplit` and `ABL_MOST_Cloudchamber` fail their gold comparisons on this machine, identically with the development binary. `Closure_BoxParity_MYNNEDMF` stops in its one-box run with the invalid-operation trap inside `ComputeDiffusivityMYNNEDMF` (the trap reported in #4046; Mac arm64 Release only, passes in Debug; the one-box run does not take the column path).
  - Unit tests pass 572/572, parallel tests 47/47 on 1 and 2 ranks.
- **Debug, MPI, no FFT** (CI configuration, bounds-checked): no warnings; parallel tests 47/47 on 1 and 2 ranks; `ctest -R "ZSplit|PBLH|Closure_BoxParity|Tiling"` passes 26 of 26.
- **Single precision:** `ERF_SurfaceLayer.cpp`, `ERF_TerrainMetrics.cpp` and the unit test compile with no warnings, also with `-Wshadow`.
- **MPI off:** both sources compile, and the tests are not registered (they sit in the existing `ERF_ENABLE_MPI AND NOT WIN32` block and in `erf_parallel_tests`).
- **Docs:** `SurfaceLayer.rst` describes the MYNN25 height on decomposed and partial-height grids and the EB restriction; Sphinx builds with no warnings.
- **Also clean:** codespell, the tab and trailing-whitespace scripts, the GPU lint scripts.
- **Not applicable:** no device code added, no new source files (GNU Make unchanged), no new inputs.

**Related:** #4042 (substep abort on z-split grids) and #4043 (surface history on z-split grids, merged) come from the same audit, found while addressing the review of #4017.


asalmgren (commented):
Review of the z-split PBL-height path. The core mechanism is sound — I checked it rather than taking the description's word for it:

- the column layout comes from `PlanarBoundary::surface_boxes()`, which asserts the surface boxes are disjoint;
- `dm_col` is shared by `ba_col` and `ba_sfc`, so the estimator's `min_thetav.array(mfi)` / `pblh->array(mfi)` local-index aliasing still lines up;
- `setRange(2, smallEnd, length)` is the correct 3-argument form;
- the ghost-then-valid `ParallelCopy` ordering does give in-domain cells from their owner and out-of-domain ghosts from the BC-filled source ghosts.

The unit-test arithmetic checks out too (min theta_v = 300, jump at k=1 -> 2, zi = 1.5 + 1.25/2 = 2.125; the TKE branch is inert with qke = 0), and the CTest entry follows the existing `add_test_box_parity` conventions.

Six findings inline, three of them medium. The theme of the medium ones is that `split_in_z` is a predicate about *box count*, and the property it needs to capture is about *box extent* — which is why partial-height and entirely-aloft levels fall through it.

One non-blocking note on the description rather than the code: it says the split run breaks at k = 24, 48, 72, but with `amr.max_grid_size_z = 24`, `n_cell_z = 100` and `blocking_factor = 4`, AMReX chops into five 20-cell boxes, so the splits are at k = 20, 40, 60, 80. The deck's own comment ("2 x 1 x 5 boxes") is right; only the PR text is off.

*Reviewed with Claude Code.*

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -2026,9 +2026,64 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+
+    // The estimator scans each column from the bottom of the domain to the top, one box
+    // column at a time, and writes the planar pblh of that box.  That is only correct when
+    // every box spans the full height: on grids split in z an upper box would read cons
+    // outside its data and write its own (wrong) copy of the planar field.  In that case run
+    // the estimator on boxes that span the full height of the domain, one per surface box,
+    // and copy the result onto every planar box.
+    const BoxArray& ba_sfc = m_planar_bndry[lev].surface_boxes();
+    const bool split_in_z = (m_terrain_type != TerrainType::EB) && (m_face.coordDir() == 2) &&
+                            (ba_sfc.size() < pblh[lev]->boxArray().size());
```
asalmgren: **medium** — this guard detects *duplicate planar boxes*, not boxes that fail to reach `domain.bigEnd(2)`.

Consider a partial-height refined level: a level-1 patch spanning k = 0..99 of a 200-cell fine domain, one box per column. Every box touches the surface, so `ba_sfc.size() == pblh[lev]->boxArray().size()`, `split_in_z` is false, and the old path runs. `ERF_PBLHeight.H:69` then sets `khi = domain.bigEnd(2)` and the k-loop reads `cons_arr(i,j,k+1)` up to k = 200 — far past the FAB's `ngrow_state` halo. That is a heap read overflow with `erf.most.pblh_calc = MYNN25`.

This is pre-existing, but it is the same failure class the PR is fixing, and ERF explicitly contemplates partial-height levels (e.g. `ERF_MakeNewArrays.cpp:1102`). The robust predicate is "some box does not span `[domain.smallEnd(2), domain.bigEnd(2)]`", not "the planar BoxArray has duplicates" — and it would subsume the current test.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -2026,9 +2026,64 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+    for (int i = 0; i < ba_sfc.size(); ++i) {
+        Box b(ba_sfc[i]);
+        b.setRange(2, domain.smallEnd(2), domain.length(2));
+        bl_col.push_back(b);
+    }
+    BoxArray ba_col(std::move(bl_col));
+    DistributionMapping dm_col(ba_col);
+
+    MultiFab cons_col(ba_col, dm_col, cons.nComp(), cons.nGrowVect());
+    cons_col.setVal(zero);
```
asalmgren: **medium** — `setVal(zero)` leaves `Rho_comp = 0` in every column cell that no source box covers.

On a level that is both split in z *and* partial-height, `split_in_z` is true, `ba_col` spans the full domain height, and the cells above the grid top stay zero. `GetThetav` (`ERF_MoistUtils.H:78`) then evaluates `RhoTheta/Rho` = 0/0, and the TKE branch (`ERF_PBLHeight.H:161-163`) does the same — NaN, or an immediate abort under `amrex.fpe_trap_invalid=1`. The scan runs to `khi` unconditionally unless both `zi` and `zi_tke` are already non-zero, which in a CBL they are not.

`zcc_col->setVal(zero)` a few lines below compounds it: `zphys = 0 < thetamin_height` is true, so the uncovered cells are also pulled into the min-theta_v scan.

Either clip the scan to the covered height, or seed the fill so uncovered cells are inert (`Rho = 1` with a large theta_v) rather than singular.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -2026,9 +2026,64 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+    const MultiFab& cons = vars[lev][Vars::cons];
+    const iMultiFab* lmask = m_lmask_lev[lev][0];
+
+    // The estimator scans each column from the bottom of the domain to the top, one box
+    // column at a time, and writes the planar pblh of that box.  That is only correct when
+    // every box spans the full height: on grids split in z an upper box would read cons
+    // outside its data and write its own (wrong) copy of the planar field.  In that case run
+    // the estimator on boxes that span the full height of the domain, one per surface box,
+    // and copy the result onto every planar box.
+    const BoxArray& ba_sfc = m_planar_bndry[lev].surface_boxes();
```
asalmgren: **medium** — when no box on the level reaches the surface, `ba_sfc` is empty, and `0 < pblh[lev]->boxArray().size()` makes `split_in_z` true.

This is the fine-level-entirely-aloft case that `ERF_PlanarBoundary.H` documents and that `fill()` handles explicitly, so it is reachable. `ba_col` is then empty, and `DistributionMapping dm_col(ba_col)` below trips AMReX's `BL_ASSERT(!boxes.empty())` in `DistributionMapping::SFCProcessorMap` (`AMReX_DistributionMapping.cpp:1482`) in a Debug build.

In Release it is quieter but worse: the copy-out has no source, so `pblh[lev]` keeps the `bogus_large_value` (1e18) it was given at `ERF_SurfaceLayer.H:682`. The old path at least did `pblh->setVal(zero)` first. Downstream `erf.rans_lscale_from_pblh`, the 2D plotfile and IBSEB would all read 1e18.

An `if (ba_sfc.empty())` early-out that keeps the old path (or zeroes `pblh`) would cover it.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -2026,9 +2026,64 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+    const bool split_in_z = (m_terrain_type != TerrainType::EB) && (m_face.coordDir() == 2) &&
```
asalmgren: **low** — `m_terrain_type != TerrainType::EB` silently routes EB terrain back to the old path.

With EB, `ba_flux = ba` (3D) and `pblh[lev]` is a 3D MultiFab, so on a z-split EB grid the estimator writes `pblh_arr(i,j,0)` and `min_thv_arr(i,j,0)` into a FAB whose z range is e.g. 24..47 with one ghost — an out-of-bounds *write*, not merely a wrong value.

Pre-existing, but this line is now where the decision is made, so it is the natural place to make it loud: an `AMREX_ALWAYS_ASSERT` that EB + MYNN25 grids span the full height would beat silent corruption.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -2026,9 +2026,64 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+    const Periodicity period = m_geom[lev].periodicity();
+
+    BoxList bl_col(ba_sfc.ixType());
+    for (int i = 0; i < ba_sfc.size(); ++i) {
+        Box b(ba_sfc[i]);
+        b.setRange(2, domain.smallEnd(2), domain.length(2));
+        bl_col.push_back(b);
+    }
+    BoxArray ba_col(std::move(bl_col));
+    DistributionMapping dm_col(ba_col);
```
asalmgren: **low (efficiency)** — this builds a fresh SFC/knapsack mapping on every call (every timestep, every level), unrelated to where the data already lives.

`m_planar_bndry[lev].surface_dm()` is already the right size and order, and puts each column on the rank that owns its surface box — which makes the copy-out mostly rank-local and removes the per-step mapping construction.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
```diff
@@ -2026,9 +2026,64 @@ SurfaceLayer::compute_pblh (const int& lev,
...
+
+    MultiFab cons_col(ba_col, dm_col, cons.nComp(), cons.nGrowVect());
```
asalmgren: **low (efficiency)** — `cons_col` is allocated and copied with all `cons.nComp()` components at full ghost width, twice, every call.

The estimator reads only `Rho_comp`, `RhoTheta_comp`, `RhoKE_comp` and the moisture indices. On a moist run this roughly doubles the state volume communicated per step; copying just the needed component span would cut it substantially.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
hgopalan: Confirmed and fixed in 6122e826d. The new unit test `PBLHeightStopsAtTheTopOfPartialHeightGrids` (four boxes over k = 0..1 of the 4-cell domain) aborts on the previous code in Debug with `(-1,-1,3,1) is out of bound (-3:18,-3:18,-1:2,0:4)`.

I put the fix in the estimator rather than in the predicate: `ERF_PBLHeight.H` now scans the cells its box holds (`mfi.validbox()` in z) instead of the domain, so the last `k+1` read ends in the box's top ghost cell, and it asserts that the box starts at the ground. A partial-height level with one box per column is then scanned directly, without the copies, and full-height boxes are scanned exactly as before. The test checks both outcomes: a jump between the top cell and its ghost cell is found (2.125 m), a jump above that is not (0).

The predicate is now "some box does not start at the ground" (stacked boxes, or boxes aloft), computed from the 3D BoxArray rather than from the planar duplicates.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
hgopalan: Agreed, and fixed by construction rather than with a fill value (6122e826d). The columns are no longer "surface box extended to the top of the domain": they are the runs of cells that start at the ground, joined with `join_boxes_stacked_in_z` (the helper the fine BTF mesh already uses, moved out of the anonymous namespace in `ERF_TerrainMetrics.cpp`). It cuts stacks at each other's lateral edges, so every valid cell of a column is a valid cell of the level and each column ends where the grids above that ground end. With the scan clipped to the box (other thread), no uncovered cell is ever read, and both `setVal(zero)` calls are gone.

I looked at seeding inert values first and dropped it: on flat ground the min-theta_v scan cannot exclude a seeded cell below 200 m, and a large theta_v seed makes the scan "find" the PBL top at the top of the refined region.

Test: `PBLHeightOnUnevenStacksAndBoxesAloft` stacks a box with a different footprint (8 <= i <= 23, k = 2..3) on three quarter boxes (k = 0..1), partly over a quarter with no ground box. Expected 3.125 m under the upper box where the ground is covered and 0 elsewhere, on every planar copy; the previous code returns 1e150. Writing it also showed that the ghost-cell pass of the copy-out reached valid planar cells next to the gap, so the copy-out now zeroes those before the valid pass and ends with a FillBoundary (all planar copies agree at that point).

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
hgopalan: Confirmed both ways on the previous code: the new test `PBLHeightIsZeroOnGridsEntirelyAloft` trips ``Assertion `! boxes.empty()' failed`` at `AMReX_DistributionMapping.cpp:1482` in Debug. Fixed in 6122e826d: on the column path `pblh` is set to zero first (the estimator's own value for a height it did not find), and with no column the routine returns there, before any DistributionMapping is built. The same zero covers the planar cells of a box aloft on a level that also has ground boxes (tested in `PBLHeightOnUnevenStacksAndBoxesAloft`).

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
hgopalan: Done in 6122e826d: `define_pblh_columns` aborts when the terrain is EB and some box does not start at the bottom of the domain, naming `erf.most.pblh_calc`, the level and `amr.max_grid_size_z`. I made the condition "starts at the ground" rather than "spans the full height" because that is what the write to the k = 0 plane needs now that the scan stops at the top of the box.

Checked with `Exec/CanonicalTests/ABL/inputs_most_nbl_EB` shrunk to 8 x 8 x 261 with `erf.most.pblh_calc = MYNN25`: with `amr.max_grid_size_x=4 amr.max_grid_size_y=8 amr.max_grid_size_z=64 amr.blocking_factor=1` the run stops at start-up with the message; unsplit, it steps normally. There is no CTest entry for it, since `add_test_expected_abort` arrives with #4042; I can add one once that is in.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
hgopalan: Done in 6122e826d, a bit further than suggested: the columns and their DistributionMapping are now built only when the level's BoxArray or DistributionMapping changes (`SurfaceLayer::define_pblh_columns`, cached per level), so nothing is constructed per step. Since the columns are joined runs of cells and no longer one per surface box, `surface_dm()` does not line up with them index for index; each column goes to the rank that owns its lowest corner cell, which gives the same locality. Two data-less MultiFabs keep the ParallelCopy metadata alive between calls, as `column_grids` does for the BTF mesh.

Source/BoundaryConditions/ERF_SurfaceLayer.cpp
hgopalan: Done in 6122e826d. The columns now receive `Rho`, `RhoTheta`, `RhoKE` and the span of the moisture species that enter theta_v (qv, qc, qi, qr, qs, qg; scalars and number concentrations are skipped), with one ghost cell instead of the state's full width. Components keep their index, since the estimator indexes the state by component number.

Because a wrong span would be silent on the dry decks, there is a test for it: `PBLHeightSeesMoistureOnZSplitGrids` makes the 2 K theta_v jump with water vapour alone. With the moisture copy removed the split layout gives 0 instead of 2.125 m.

hgopalan:
Thanks for the review. All six findings are addressed in 6122e826d (pushed on top of a merge of development, which brought in `surface_dm()` from #4043; no force-push).

| # | Finding | Change | Test (previous code) |
|---|---------|--------|----------------------|
| 1 | Partial-height level scans past the FAB | Estimator scans the cells its box holds; asserts the box starts at the ground | `PBLHeightStopsAtTheTopOfPartialHeightGrids` (Debug: `(-1,-1,3,1) is out of bound`) |
| 2 | Zero fill gives 0/0 in uncovered column cells | Columns are the runs of cells from the ground (`join_boxes_stacked_in_z`), so no cell needs a fill | `PBLHeightOnUnevenStacksAndBoxesAloft` (returns 1e150) |
| 3 | Level entirely aloft: empty BoxArray, stale sentinel | `pblh` = 0 where no box reaches the ground; early return | `PBLHeightIsZeroOnGridsEntirelyAloft` (``! boxes.empty()``) |
| 4 | EB silently on the old path | Abort naming the input | EB deck, split vs unsplit (by hand; no expected-abort harness before #4042) |
| 5 | Mapping rebuilt every call | Columns cached per level, rebuilt when the grids change | - |
| 6 | All components, full ghost width | 3 components + moisture span, one ghost cell | `PBLHeightSeesMoistureOnZSplitGrids` (mutant gives 0) |

The description now says the split run breaks at k = 20, 40, 60, 80 (checked in `Level_0/Cell_H`: ten boxes of 4 x 8 x 20).

Verification on this head:
- Release (MPI, FFT, all warnings) and Debug (MPI, no FFT): no warnings. Unit tests 572/572; parallel tests 47/47 on 1 and 2 ranks in both builds.
- Debug `ctest -R "ZSplit|PBLH|Closure_BoxParity|Tiling"`: 26/26, including `PBLH_ZSplit_BoxParity` and `Closure_BoxParity_kEqn_PBLHCap`.
- Release `ctest -L regression`: 120/123. `ABL_MOST_WOA_ZSplit` and `ABL_MOST_Cloudchamber` fail their gold comparisons on this machine as before. `Closure_BoxParity_MYNNEDMF` (new with the development merge) stops in its one-box run with the invalid-operation trap inside `ComputeDiffusivityMYNNEDMF`; that is the trap reported in #4046, it fires on this Mac (arm64, Apple clang, Release) and not in Debug, and the one-box run does not take the column path.
- Single precision with `-Wshadow`: the two sources and the unit test compile without warnings. MPI off: both sources compile. Sphinx: no warnings. codespell, tabs, trailing whitespace: clean. No device code added.

PR #4050, opened 2026-09-17
EB: Avoid invalid density access in EB VelocityToMomentum

skang67:
This PR fixes the `SIGILL` issue in `VelocityToMomentum` for EB, and applies the same guard everywhere else the EB code reads density or eddy viscosity on a covered cell.

## Problem

For EB, momentum is computed using volume-fraction-weighted average of density. Density values are accessed regardless whether they are covered. Some compilers trigger `SIGILL` when covered-cell density is evaluated, even though it is multiplied by a zero volume fraction.

## Solution

Guard accessing rho using conditional operator. The same guard is applied in `MomentumToVelocity` and in the off-diagonal averages in `ComputeStress_EB`, for both `rho` and `mu_turb`. `ERF_MakeTauTerms.cpp` now takes `vfrac` from the EB factory directly rather than relying on `detJ` aliasing the volume fraction.

The `USE_FC_FACTORY` change that was originally part of this PR has been reverted (b66318f); its effect is wider and it will be handled in a separate PR.

## Effect on the regression tests

Over 20 steps of `Exec/RegTests/WitchOfAgnesi/inputs_EB_x`, this branch is bitwise identical to `development` (`fcompare`: `PLOTFILE AGREE`). The EB reference solutions therefore do **not** need to be regenerated.


Source/EB/ERF_EB.H
```diff
@@ -95,7 +95,7 @@ class eb_ {
      * \def USE_FC_FACTORY
      * \brief Select native AMReX face-centered factories instead of ERF auxiliary factories.
      */
-    #define USE_FC_FACTORY 0 // Default to 0: use eb_aux_ (original)
+    #define USE_FC_FACTORY 1 // Default to 0: use eb_aux_ (original)
```
asalmgren: **HIGH — this flip makes every EB + multigrid run abort at startup, because `EB2::BuildFC()` is never called on that path.**

In `Source/EB/ERF_EBInit.cpp`, `EB2::BuildFC()` appears only in the *non*-multigrid arm of each geometry branch (lines 103, 153, 169, 186). When `build_eb_for_multigrid` is true (line 36: `terrain_type == EB && (project_initial_velocity[0] == 1 || anelastic[0] == 1)`) the `max_coarsening_level` overload of `EB2::Build` is taken at lines 98/148/164/181 and `BuildFC()` is skipped entirely.

But `ERF::MakeNewLevelFromScratch` calls `eb[lev]->make_all_factories(...)` *unconditionally* for `terrain_type == EB` (`Source/ERF_MakeNewLevel.cpp:82`, and likewise `:349` and `:725`). With `USE_FC_FACTORY 1`, `eb_::make_all_factories` (`Source/EB/ERF_EB.cpp:50-80`) constructs `EBFArrayBoxFactory(..., idim)` for all three faces, which hits

```
AMREX_ALWAYS_ASSERT_WITH_MESSAGE(a_level.hasFCData(face_dir),
    "EBDataCollection: FC data not available for face_dir");
```

(`Submodules/AMReX/Src/EB/AMReX_EBDataCollection.cpp:101`) — an ALWAYS assert, so it fires in optimized builds too.

Existing inputs that hit this: `Exec/RegTests/WitchOfAgnesi/inputs_EB_anelastic`, `inputs_EB_anelastic_twolevel`, `Exec/RegTests/EB_SquareCylinder/inputs_anelastic`, `Exec/RegTests/EB_Poiseuille/inputs_compressible`.

The comment at `Source/EB/ERF_EB.cpp:51` ("EB2::BuildFC called in ERF.cpp") is only conditionally true. Since `EB2::BuildFC()` takes no arguments and just calls `IndexSpace::top().buildAllFCData()`, hoisting it out of the `if`/`else` in all four geometry branches is enough to fix this.

---

Minor, same line: the trailing comment `// Default to 0: use eb_aux_ (original)` is now false and will mislead the next reader or bisect.

asalmgren (commented):
Automated code review.

The `ERF_VelocityToMomentum.cpp` change itself looks correct: `vfrac_i + vfrac_im1 > 0` is equivalent to the old `||` test for non-negative volume fractions, the ternaries genuinely prevent the covered-cell load (they are sequence points, not just hints), and zeroing momentum on a fully covered face is consistent with `mask_fc_from_cc`, which marks a face covered exactly when both CC neighbors are covered.

The issues are in the `USE_FC_FACTORY` flip (inline comment below) and in what the fix leaves behind elsewhere.

---

### The SIGILL is only half fixed — `Source/Utils/ERF_MomentumToVelocity.cpp` (MEDIUM)

Lines 79-81, 87-89 and 95-97 still contain byte-for-byte the pattern this PR is removing:

```cpp
if (c_vfrac_arr(i,j,k) > zero || c_vfrac_arr(i-1,j,k) > zero) {
    Real rho = (c_vfrac_arr(i,j,k) * dens_arr(i,j,k,Rho_comp)
            + c_vfrac_arr(i-1,j,k) * dens_arr(i-1,j,k,Rho_comp))
            / (c_vfrac_arr(i,j,k) + c_vfrac_arr(i-1,j,k));
```

Any x-face with `vfrac(i,j,k) > 0` and `vfrac(i-1,j,k) == 0` — the outermost fluid face against the EB, precisely the condition described in the PR — enters the `if` and evaluates `dens_arr(i-1,j,k,Rho_comp)` on a covered cell. `MomentumToVelocity` runs on the same EB data every step, so the crash this PR is chasing can still occur. The same ternary guard applies.

### Third instance — `Source/Diffusion/ERF_ComputeStress_EB.cpp:57,70,86` (LOW/MEDIUM)

In all three off-diagonal lambdas, the `if (vol_sum > 1e-16)` branch multiplies covered-cell densities by a zero `vfrac` (same trap: the multiply does not prevent the load), and the `else` branch averages all four densities unconditionally — which is reached exactly when all four cells are covered, the case this PR just stopped doing in `VelocityToMomentum`.

### Scope of the flip is wider than the two named reg tests (MEDIUM)

Besides `inputs_EB_x` / `inputs_EB_y`, setting `USE_FC_FACTORY 1` also:

- switches `TerrainType::ImmersedForcing` / `BuildingsType::ImmersedForcing` from `make_cc_factory` to `make_all_factories` (`Source/ERF_MakeNewLevel.cpp:85-88`, `:352-355`, `:728-731`), and
- allocates and populates `terrain_blanking_{x,y,z}face` (`Source/ERF_MakeNewArrays.cpp:108-110`, `:1009-1040`, `:1194-1225`), which flips immersed forcing from the cell-centered average to a face-centered blanking field in `Source/Initialization/ERF_InitImmersedForcing.cpp:48-60`, `Source/TimeIntegration/ERF_TI_slow_rhs_pre.H:65-73`, `ERF_TI_slow_rhs_post.H:49-55`, and `ERF_TI_substep_fun.H:47-53`.

So every ImmersedForcing/buildings test, plus MOST (`ERF_MOSTAverage.cpp:2689`) and surface-layer (`ERF_SurfaceLayer.cpp:1389`) EB results, change too and need regolding. Note also that there is no `USE_FC_FACTORY` override anywhere in `.github/`, `CMakeLists.txt`, `Build/`, or `CMake/`, so the `=1` branches have never been compiled in CI.

---

<sub>Checked and clear: the FC-factory API surface used through `get_{u,v,w}_const_factory()` (`getVolFrac`/`getCentroid`/`getBndryCent`/`getBndryNormal`/`getAreaFrac`/`getFaceCent`/`getMultiEBCellFlagFab`) all exists on `amrex::EBFArrayBoxFactory`; both `solve_with_EB_mlmg` instantiations are present (`ERF_SolveWithEBMLMG.cpp:162-179`); `ng_sub` stays within the allocated ghost widths for the face blanking MultiFabs; all `terrain_blanking_*face` consumers null-check; and `mask_fc_from_cc`'s ghost reach (`grow(face_box, nGrow-1)` = 4) is within the CC flags' 5.</sub>

Source/EB/ERF_EB.H
skang67: 1. I applied similar guard to accessing rho in `MomentumToVelocity` and `ComputeStress_EB`, which resolves the first and second issues raised.
2. For flipping `USE_FC_FACTORY`, I set it back to original; Since the effect of modifying this is wider, it will be better to handle in another separate PR.

asalmgren (commented):
Automated code review (revision 2).

Net diff is 3 files (`ERF_VelocityToMomentum.cpp`, `ERF_MomentumToVelocity.cpp`, `ERF_ComputeStress_EB.cpp`); the `USE_FC_FACTORY` change nets out to zero.

### Follow-up on the previous review

| Earlier finding | Status |
|---|---|
| SIGILL only half fixed — `MomentumToVelocity` still has the unguarded pattern | **Fixed** (889de77). But the follow-on "Enable fma" commit introduces a new problem — below. |
| Third instance — `ComputeStress_EB` off-diagonal lambdas | **Mostly fixed** (030a0e9). The `else` that averaged four covered densities is gone. Two residual holes remain. |
| `USE_FC_FACTORY 1` aborts every EB+multigrid run (`EB2::BuildFC()` never called) / scope wider than the two named reg tests | **Moot** — reverted in b66318f. The `BuildFC()` gap still exists in `development` and will need fixing in whichever PR eventually flips the macro. |

---

### HIGH — the `std::fma` commit is what changes the EB answers, and it costs a libm call per face

I built this branch and a baseline (same tree, the three files reverted to `development`) and ran `Exec/RegTests/WitchOfAgnesi/inputs_EB_x` for 20 steps:

* **PR vs development:** `x_velocity` 1.30e-17, `y_velocity` 6.09e-20, `z_velocity` 8.67e-19 absolute error.
* **PR with `std::fma` replaced by the plain guarded sum, vs development:** `PLOTFILE AGREE` — bitwise identical.

So commit 8bb3310 is the *sole* source of the answer change it was apparently meant to prevent.

The reason: ERF's default Release flags are `-O3 -DNDEBUG -std=c++20 -Werror=return-type -finline-limit=43210` — no `-march`, so the baseline x86-64 target has no FMA instruction and GCC never contracts `a*b + c*d`. Confirmed in codegen: `a*b+c` compiles to `mulsd`/`addsd`, while `std::fma(a,b,c)` compiles to `jmp fma@PLT`. The actual built object `ERF_MomentumToVelocity.cpp.o` carries **three `R_X86_64_PLT32 fma` relocations** (one per lambda) and zero `vfmadd`. Only under `-march=native` do both forms collapse to a single `vfmadd132sd`.

On the default build that means the `std::fma` both (a) changes the rounding relative to `development`, and (b) puts an un-inlinable libm call in the innermost per-face loop, which also blocks vectorization of that loop.

**Suggestion:** drop 8bb3310 and the `#include <cmath>`, and keep the symmetric `(vfrac > zero ? vfrac * rho : zero) + …` form already used in the other two files. The PR then becomes answer-preserving, and the request to @AMLattanzi to regenerate the WitchOfAgnesi EB reference solutions can come out of the description.

### MEDIUM — covered-density reads this PR leaves behind in `ComputeStress_EB`

* `ComputeStressConsVisc_EB`, cell-centered loop (`Source/Diffusion/ERF_ComputeStress_EB.cpp:99`): `Real rhoAlpha = cell_data(i, j, k, Rho_comp) * mu_eff;` with no `vfrac` guard — while the matching loop in `ComputeStressVarVisc_EB:196` *does* guard with `if (vfrac(i,j,k) > zero)` and zeroes the taus otherwise. Same asymmetry this PR closes everywhere else; it is a two-line addition.
* The new `vol_sum >= four` fast path reads all four `cell_data` unguarded. That is safe when `vfrac` really is the EB volume fraction, but in `Source/TimeIntegration/ERF_MakeTauTerms.cpp:122-135` `vfrac` is aliased to `detJ` whenever the fab type is not `singlevalued` — including `FabType::covered`, where `detJ == 1` so `vol_sum == 4` and four covered densities get read. Pre-existing (`development` read them via the weighted branch), but the PR's own goal is not met on that path.

### LOW — `four - Real(1.e-16)` is not a tolerance

`4.0 - 1e-16` rounds to exactly `4.0`: the double spacing near 4 is 2^-50 ≈ 8.9e-16, so 1e-16 is below half an ulp (checked in both double and float). The test is therefore exactly `vol_sum >= 4.0`. Harmless, since regular `vfrac` is exactly 1.0 and the four-term sum is exact — but the code reads as if it has slack it does not have. Either drop the epsilon or write it as `four*(one - eps)`.

### LOW — none of this is covered by `ctest`

No input under `Tests/test_files/` sets `erf.terrain_type = EB`, and `Tests/CTestList.cmake` never overrides `terrain_type`. All three modified code paths are gated on `TerrainType::EB`, so the regression suite never executes them — which is also why no gold files move here. (`IBSEB_Cube` uses `buildings_type = ImmersedForcing`, a different gate.) The inputs that do exercise this — `WitchOfAgnesi/inputs_EB_{x,y}`, `EB_Poiseuille/inputs_compressible`, `EB_SquareCylinder/*` — are not registered as tests. Adding one short EB case would be worth it.

### Question on the root cause

It would be useful to record in the PR which compiler and flags reproduce the SIGILL. Covered density is not obviously a trap value here: `vars_new/old[Vars::cons]` is `setVal(zero)` at define time (`Source/ERF_MakeNewArrays.cpp:177`), and density is deliberately defined with one more ghost cell than velocity so the `i-1` access stays in bounds (`Source/ERF_MakeNewArrays.cpp:155-161`), so it is not an out-of-range index either. If covered cells really do hold signaling NaNs, an `EB_set_covered` on the state would close the whole class at once rather than guarding each use site — there will be more of these. Worth noting too that the guards rely on the compiler not if-converting the ternary into an unconditional multiply plus select; GCC will not under its default `-ftrapping-math`, but Clang's default FP exception behavior is `ignore`.

---

<sub>Checked and clear: the guard logic in `VelocityToMomentum` is equivalent to the old `||` test for non-negative volume fractions; zeroing momentum on a fully covered face changes nothing measurable (it is inside the bitwise-identical run above) and is strictly better than the old garbage-weighted average; in `ComputeStress_EB` the new fast path and the weighted branch are both bitwise identical to `development` for `vol_sum == 4` and for `0 < vol_sum < 4` (same summand order, and `1*x` and `/4` vs `*0.25` are all exact); with `eb2.small_volfrac` in play the `0 < vol_sum <= 1e-16` window where the branch condition flipped is unreachable; and the PR adds no new ParmParse keys, so nothing is owed to `Docs/sphinx_doc/Inputs.rst`.</sub>

**Verdict:** good to merge once 8bb3310 is dropped — that single change makes the PR answer-preserving and removes the per-face libm call. The `ComputeStressConsVisc_EB` cell-centered guard is worth folding in while you are here.

skang67:
The first three comments from the second PR review were addressed. And the last comment regarding `ctest` should be handled in management.

PR #4051, opened 2026-09-17
MRF Modifications

AMLattanzi:
# MRF PBL scheme — correctness fixes

Branch: `FixMRF` · Base: `development` (`9929d94a9`)

The MRF boundary-layer scheme in `Source/PBL/ERF_ComputeDiffusivityMRF.cpp` was reviewed
line-by-line against its stated benchmark, WRF's `phys/module_bl_mrf.F` (fetched from the WRF repository at `master`), with WRF YSU (`v4.4/phys/module_bl_ysu.F`) and ERF's own `ERF_ComputeDiffusivityYSUNew.cpp` used as secondary references. Five of the findings are fixed here; the rest are listed at the end as not addressed.

Line numbers below refer to the post-fix file.

---

## 1. Guard the mixed-layer `pblh_rel` division (`:1029`)

**Was:** `pblh_rel = pblh - z_sfc`, used as a denominator with no guard.

**Problem:** the Rib passes floor the diagnosed height at `pblh_min = max(pblh_emp, 10)`, and `pblh_emp` is the *same expression* as the first-level height. So for any column whose height lands on that floor with a first layer thicker than 20 m, `pblh - zl1` is exactly zero — at the same time as `zval - zl1` is zero at `k == ksrf`, making the quotient `0/0`. `amrex::max` propagates the resulting NaN straight into `K_turb`.

**Now:** `pblh_rel = max(pblh - zl1, 1e-4)`, matching `ERF_ComputeDiffusivityYSUNew.cpp:1701`,
which already guarded the identical expression.

**Status of evidence:** the degeneracy is structural (`pblh_min` and `zl1` are the same expression), but it was *not* reproduced in a live run. With the guard temporarily removed and a terrain case built with a 64 m first layer — so `pblh_min == zl1` exactly — neither an unstable nor a stable/QNSE configuration triggered it, because no column clamped to its PBLH floor in those runs. Reaching it requires `Rib` at the first cell to already exceed `Ribcr`, plausible with a sharp, poorly resolved surface inversion. The guard is free and matches the sibling scheme, so it was kept, but this is reasoned rather than demonstrated.

---

## 2. Floor `HGAMT` at zero (`:768-778`)

**Was:** `HGAMT = min(-b·u*·θ*/wstar, GAMCRT)` — upper limit only.

**WRF** (`module_bl_mrf.F`):

```fortran
HGAMT(I)=MIN(GAMFAC*HFX(I)/CPM(I),GAMCRT)   ! L874
VPERT=HGAMT(I)+EP1*THX(I,KL)*HGAMQ(I)       ! L877
HGAMT(I)=MAX(HGAMT(I),0.0)                  ! L880
```

**Problem:** the floor is not redundant. The unstable branch is selected on `sign(L)`, but the numerator is the sensible flux `-u*·θ*` alone. Over water with a small downward sensible flux and a large latent flux, `L` is negative while `-u*·θ*` is not — and without the floor the "countergradient" term adds to the down-gradient flux instead of opposing it. ERF already applied the matching `max(HGAMQ, 0)` (WRF L881), so the omission looked accidental.

**Now:** `max(min(...), 0)` in Pass 4. Pass 2 deliberately keeps the unfloored value for VPERT, because WRF forms VPERT at L877 *before* flooring at L880.

---

## 4. Pass 4 uses the corrected PBL height (`:749`)

**Was:**

```cpp
const Real zi = amrex::max(pblh_pred_arr(i, j, 0), Real(10.0));  // or pblh_corr_arr in Pass 4
```

a verbatim copy of the Pass 2 line, trailing comment included.

**Problem:** Pass 4 exists "to ensure internal consistency between the K-profile amplitude and the countergradient fluxes" and its `HOL` correctly uses `pblh_corr_arr`, but `wstar_conv` was still built from the *predictor* height. `wstar` feeds both the K-profile amplitude and `HGAMT`/`HGAMQ`, so the pass did not achieve what it was there for.

**Now:** `pblh_corr_arr`. The Pass 2 counterpart was re-commented so the two no longer read as copies.

---

## 5. Move the `ZL1` offset into `zfac` only (`:1016-1033`)

**WRF** (`module_bl_mrf.F:976-978`):

```fortran
ZFAC=MAX((1.-(ZQ(I,K)-ZL1(I))/(PBL(I)-ZL1(I))),ZFMIN)
XKZM(I,K)=XKZO+WSCALE(I)*KARMAN*ZQ(I,K)*ZFAC**PFAC
```

The first-level offset `ZL1` belongs to `ZFAC` alone; the linear factor is the full height above ground, `ZQ(I,K)`.

**Problem:** ERF subtracted `z_sfc` from *both*, so under terrain-fitted coordinates `K_m` was identically zero in the first fluid cell — the one cell where the surface-layer flux is delivered. The non-terrain path had the mirror-image error (`z_sfc = 0`, so the linear factor was right but `zfac` omitted the offset). `ERF_ComputeDiffusivityYSUNew.cpp:1699-1720` splits the two correctly.

**Why the split matters.** `zl1` is the height of the lowest model level above the surface (WRF `ZL1(I) = ZA(I,KL)`, `module_bl_mrf.F:809`) — the surface-layer reference height, where `u*`, `θ*` and `L` are handed to the PBL scheme, and the same height WRF uses for the log law (L643) and the bulk Richardson number (L660). The two factors need different origins:

| factor | origin | job |
|---|---|---|
| linear `z` | ground | reduce to the surface-layer form `K → κ·w_s·z` so the profile joins the log law |
| `zfac` | `zl1` | vanish at the PBL top `z = h`, and equal exactly 1 at `z = zl1` |

**Now:**

```cpp
const Real zl1 = (use_terrain_fitted_coords)
               ? Compute_Zrel_AtCellCenter(i, j, ksrf, z_nd_arr)
               : ((ksrf + myhalf) * gdata.CellSize(2) - zib);
const Real pblh_rel = amrex::max(pblh - zl1, Real(1.0e-4));
const Real zfac = amrex::max(Real(1) - (zval - zl1) / pblh_rel, Real(1.0e-8));

K_turb(i, j, k, EddyDiff::Mom_v) = rho * wstar * KAPPA * zval * zfac * zfac;
```

`zl1` is `zval` evaluated at the column's first fluid cell, so `zval - zl1` is exactly zero
there in both the terrain and the non-terrain path.

> **Fixes 1 and 5 are coupled — do not ship 5 without 1.** Previously the non-terrain path had
> `z_sfc = 0`, so `pblh_rel = pblh ≥ 10` and could never vanish; the `0/0` was terrain-only.
> Now that fix 5 gives the non-terrain path a real `zl1 = dz/2`, that path is exposed to the
> same degeneracy, and fix 1's guard is what covers it.

---

## 11. Retire the dead `mrf_moistvars` option

**Problem:** the option was a complete no-op. All three call sites computed identical values in both branches:

- K-profile branch: `Prq_base = phit_eff/phiM_eff` was character-for-character `Prt_base`, and `Prq` applied the same `+ const_b·κ·sf` and the same bounds, so `K_m/Prq == K_m/Prt == K_θ`, which is what the `else` branch assigned.
- Both local-mixing branches: `K_q = rl2wsp*ft`, and `K_θ` had just been set to `rl2wsp*ft`.

`K_q = K_θ` is also WRF's actual behaviour — moisture is diffused with the heat coefficient `XKZH` (`module_bl_mrf.F:1069, 1180`).

**Now:** branches collapsed to `K_q = K_θ` with the WRF citation. The option was MRF-only, so it was removed from `ERF_TurbStruct.H` (declaration, ParmParse query, startup print, doxygen), from `Docs/sphinx_doc/Inputs.rst`, and the now-meaningless `erf.mrf_moistvars` line was stripped from the 43 input files that set it. `grep` confirms zero remaining references repo-wide.

---

## Files touched

| file | change |
|---|---|
| `Source/PBL/ERF_ComputeDiffusivityMRF.cpp` | all five fixes; header docstring corrected for `K_q` |
| `Source/DataStructs/ERF_TurbStruct.H` | `mrf_moistvars` removed (4 sites) |
| `Docs/sphinx_doc/Inputs.rst` | `erf.mrf_moistvars` row removed |
| 43 × `Exec/CanonicalTests/**/inputs*` | stale `erf.mrf_moistvars` line removed |

## Verification

- Full CMake build (MPI, Release, GCC 15.2): **0 errors, 0 warnings**.
- Five MRF cases run 200–300 steps each under `amrex.fpe_trap_invalid=1`, no NaN, no abort:
  - `Exec/CanonicalTests/ABL/MRF_YSUNew_Enhancements/canonical/{unstable,neutral,veryunstable_cgcheck}_abl`
  - `Exec/CanonicalTests/ABL/MRF_YSUNew_Enhancements/countergradient_terrain_check/{stretched,uniform}`
  — the last two exercising the terrain path that fixes 1 and 5 touch.

**Answers change.** Fixes 2, 4 and 5 all alter results in the mixed layer; fix 5 changes the
first-level `K_m` under terrain from zero to the surface-layer-matched value. Regression
baselines for MRF cases need regenerating, and a comparison against the `WRF_ideal`
unstable/stable ABL cases named in the file header would be worth doing before merge.


hgopalan (approved):

hgopalan:
@AMLattanzi, an MRF finding in the block this PR edits, the K branch below the PBL top. It is not a clean fix, because it trades one problem for another, so I would like your view before anything goes into code.

**What the code does.** Below the PBL top, the kernel uses the K-profile only when `SFCFLG || use_qnse_stable_profile`. With a stable or neutral surface layer (L > 0, no QNSE), it uses the free-atmosphere mixing `rho l^2 |dU/dz| f(Ri)` instead, so K follows the local shear inside the PBL.
- **WRF:** `module_bl_mrf.F` has no regime test below `KPBL` (L968-986). `XKZM = XKZO + WSCALE*KARMAN*ZQ*ZFAC**PFAC`, and stability enters only through `WSCALE = UST/PHIM`. Ri mixing is used only above `KPBL` (L988 on).
- **ERF before #3367:** the profile was used in every regime up to 740a64750. The switch came in with edbd74ea6 (#3367), whose header still documents the stable `phi_m` of the profile and puts the Ri mixing "above PBL". #3486 then brought the profile back for QNSE only.
- **YSUNew** keeps a profile inside a stable PBL (`ERF_ComputeDiffusivityYSUNew.cpp` L1775).

**Why it matters: anelastic MidPoint.** On a horizontally uniform neutral column (4 x 4 x 200, dz = 5 m, 10 m/s geostrophic, no surface heat flux) with the implicit vertical solve, MRF on this branch (47be816aa) develops a grid-scale wind mode inside the PBL once dt K / dz^2 is above about 2.
- The mode flips sign every step and never decays. The centred shear stencil at level k reads only the other-parity levels, so the K profile goes jagged and keeps the mode alive.
- The table below is 1 h from an explicit spin-up, with explicit anelastic at dt = 0.125 s as reference (Kmv 21.9).

| dz | dt | this branch: max Kmv | du/dz sign changes in the PBL |
| --- | --- | --- | --- |
| 5 m | 2 s | 21.9 | 0 |
| 5 m | 5 s | 61.5 | 46 |
| 5 m | 20 s | 326 | 58 |
| 5 m | 60 s | 335 (u off by 1.45 m/s, theta by 5 K) | 69 |
| 20 m | 40 s | 21.8 | 0 |
| 20 m | 60 s | 52 | 8 |

YSUNew on the same columns is clean at every dt up to MidPoint's own limit (dt 120 s at dz 5 m, the limit kEqn and Deardorff hit too).

**The trade-off.** Restoring the profile in every regime, as in WRF, removes the mode: 0 sign changes at every dt on both grids, and within 4e-3 m/s of explicit at dt = 60 s. But on GABLS1 (`Exec/CanonicalTests/ABL/mrf_stable_gabls`, 9 h) it mixes about twice as much as the current code and YSUNew:

| GABLS1, 9 h | max Kmv | u* | theta at 16 / 208 / 400 m |
| --- | --- | --- | --- |
| this branch (shear K inside the PBL) | 2.65 | 0.383 | 263.34 / 263.78 / 264.13 |
| profile in every regime (WRF) | 4.72 | 0.420 | 263.64 / 264.46 / 264.74 |
| YSUNew | 2.27 | 0.344 | 263.34 / 263.62 / 263.87 |

Over-mixing stable layers is the known MRF behaviour that Hong (2010) addressed in YSU, and I suspect it is why #3367 switched to shear K.

**Middle options tried; none survives MidPoint.** Each was measured on the column at dt 5 / 20 / 60 s and on GABLS1:

| inside a stable PBL | du/dz sign changes | GABLS1 Kmv, u* |
| --- | --- | --- |
| min(profile, shear K) | 0 / 50 / 64 | 2.57, 0.397 |
| shear K from compact face differences | 42 / 62 / 74 | 2.67, 0.381 |
| min(profile, compact shear K) | 0 / 36 / 44 | 2.57, 0.397 |

Any K that depends on the local shear inside the PBL seems to keep the mode once dt K / dz^2 is large, because MidPoint does not damp stiff modes. So the choice looks like one of:
1. the WRF profile in every regime, with WRF's over-mixing in stable layers;
2. the current shear K, with MRF + MidPoint unusable above about dt = dz^2 / (2K) on fine near-surface grids (a start-up warning or abort, perhaps);
3. an integrator-side remedy, such as a slightly off-centred first MidPoint stage, which would damp stiff modes for every closure.

**Also on this branch:** the `PBL_IBAware_MRF_Smoothing` gold file already fails here (Kmv 13 % relative), independent of this, so it needs regenerating with #4051 either way.

I have option 1 ready on top of 47be816aa. It comes with a CTest, `MRF_Neutral_MidPoint`: a 4 x 4 x 100 column at dz = 10 m, dt = 20 s, 120 steps, MLMG. It fails on this branch with 24 and 27 sign changes and Kmv 58, and passes with option 1 (0 and 1, Kmv 25). It runs in 6 s in Release and 113 s in Debug no-FFT, and a checker self-test is included. The test also serves options 2 and 3 as they stand. Which way would you like MRF to go?

hgopalan:
Follow-up on option 3: I tried it. Solving the first MidPoint stage over θ·dt instead of dt/2 (θ = 0.5 reproduces this branch bit for bit) does not damp the mode, it destabilises the run. MRF blows up at θ = 0.55, 0.6 and 1.0 (steps 49, 36 and 21), and kEqn at θ = 1.0 (step 45), so a simple off-centring is not viable in the current stage structure. The integrator-side scheme that did remove the mode, with MRF's shear K kept, was RK2 with the solve in both stages and a half step on the second, the one #4021 removed as first order in time. So options 1 and 2 remain, plus a MidPoint stage redesign if you want a second-order remedy.

hgopalan:
Thanks, you're right, and I got the history wrong in the follow-up. #4021 did not remove anything from RK2: upstream has had the vertical implicit solve off for all anelastic runs since #3329, and #4021 kept it off for RK2 and added the single first-stage solve for MidPoint. The RK2 variant I was describing was a local change in our fork, never upstream.

What that variant did: the Heun stages are unchanged (both advance dt from Sⁿ, and stage 2 averages the slow sources), but the implicit solve runs in both stages. Because the solve is folded into the slow tendency, the stage-1 tendency that stage 2 averages in, F₀ = (S₁ − Sⁿ)/dt, already contains the full stage-1 implicit increment, so the Heun average passes half of it into stage 2. The stage-2 solve was therefore taken over dt/2. With dt instead, the column gained 1.45–1.48 times the surface heat flux in one step, against 0.975–0.991 with dt/2. For pure diffusion, with z = dt·λ:
- stage 1: S₁ = (1 − z)⁻¹ Sⁿ;
- stage 2: (1 − z/2) Sⁿ⁺¹ = Sⁿ + (dt/2) λS₁;
- together: Sⁿ⁺¹ = Sⁿ / (1 − z).

That is backward Euler: first order and L-stable. It is exactly what the #4021 comment says two solves reduce to, and it is why it damped the MRF mode. So it is not a second-order remedy, and I should not have offered it as one.

On a stable, second-order MidPoint redesign: I don't have one I have verified. The standard candidate would be an IMEX pair with a second-order, L-stable implicit part, e.g. ARS(2,2,2) (Ascher, Ruuth & Spiteri 1997): SDIRK2 with γ = 1 − 1/√2 and two implicit solves over γ·dt, paired with a second-order explicit RK. That means moving the first stage to t + γ·dt and adding a second solve in stage 2, so it is a change to the stage logic in `ERF_MRI.H`, not a factor on the solve. The off-centring I tried only stretched the first-stage solve to θ·dt while the explicit part of that stage stayed at dt/2. A pure-diffusion analysis says that should damp stiff modes, and in ERF it went unstable instead, so the stages couple through more than that model captures and I would not trust a redesign until it is tested in the code. If you think ARS(2,2,2) is worth a look, I can prototype it on the neutral column and GABLS1 before anyone commits to it.

That leaves options 1 and 2 from my first comment as the ones available now.

hgopalan:
I prototyped ARS(2,2,2) on current development (ac1db3bcc), as a local scratch `erf.anelastic_type = ARS222`; it is not pushed.

**How it maps onto the two anelastic stages.** γ = 1 − 1/√2 and δ = 1 − 1/(2γ) = −1/√2.
- Stage 1 advances γ·dt, with the implicit solve at factor 1, so Y₂ = Sⁿ + γ·dt·[f_E(Sⁿ) + f_I(Y₂)].
- Stage 2 combines the sources as (1 − δ)·F(Y₂) + δ·F₀, where F₀ = (Y₂ − Sⁿ)/(γ·dt) is the reconstructed stage-1 tendency. It keeps an explicit fraction 2√2 − 2 of the vertical diffusion at Y₂ (`vert_implicit_fac[1]` = 3 − 2√2) and solves with coefficient γ·dt.
- With these, the stage-2 update is the ARS(2,2,2) one: yⁿ⁺¹ = yⁿ + dt·[δ·f_E(yⁿ) + (1 − δ)·f_E(Y₂) + (1 − γ)·f_I(Y₂) + γ·f_I(yⁿ⁺¹)].
- The changes are in `ERF_MRI.H` (stage-1 step), `ERF_SlowRhsPre.cpp` / `ERF_SlowRhsPost.cpp` (stage-2 weights), and `ERF_ImplicitPre.H` / `ERF_ImplicitPost.H` (stage-2 solve step).

**Conservation.** Convective RANS deck, 40 steps at dt = 5 s: the column heat gain over ρ_sfc·F·t is 0.9907, against 0.9908 for MidPoint.

**MRF with the current shear K, neutral column** (4 x 4 columns, 10 m/s geostrophic, 1 h from an explicit spin-up, compared with explicit anelastic at a small dt). The table gives du/dz sign changes inside the PBL at dt = 5 / 20 / 60 / 120 s:

| dz | MidPoint | ARS(2,2,2) |
| --- | --- | --- |
| 5 m | 46 / 58 / 69 / fails | 30 / 28 / 20 / fails (max Kmv 62 to 152, explicit 21.9) |
| 20 m | 0 / 0 / 8 / 10 | 0 / 0 / 0 / 0 |

ARS weakens the mode and clears it at dz = 20 m, but not at dz = 5 m. The profile in every regime (option 1) remains the only change that removed it on both grids.

**kEqn and Deardorff, same column at dz = 5 m.** The table gives max |u − u_explicit| after 1 h in m/s, at dt = 5 / 20 / 60 / 120 s:

| closure | MidPoint | ARS(2,2,2) |
| --- | --- | --- |
| kEqn | 8.9e-4 / 2.2e-3 / 4.8e-3 / fails | 2.4e-3 / 2.7e-3 / 4.8e-3 / 4.4e-2 |
| Deardorff | 5.5e-4 / 2.1e-3 / 6.7e-3 / fails | 5.5e-4 / 2.1e-3 / 6.6e-3 / 1.4e-2 |

ARS runs at dt = 120 s where MidPoint fails. It matches MidPoint's accuracy for Deardorff and is less accurate for kEqn at small dt. None of these errors falls like dt², because the eddy diffusivity is computed once per step from the old state and that lag dominates, so these runs cannot confirm second order in the coupled problem.

**GABLS1, run anelastic** (the shipped deck is compressible, where neither scheme applies). The explicit reference at dt = 1 s has max Kmv 2.64.
- At dt = 5 and 20 s, both schemes are within 4e-4 m/s of the reference.
- At dt = 60 s both break: max Kmv 10.9 (MidPoint) and 11.6 (ARS), with the wind off by 0.91 and 0.43 m/s. The Kmv profile inside the PBL alternates between the floor value (0.13) and about 10 from level to level, so MRF's Ri arm is switching on and off with the step at dt·K/dz² of only about 0.6.
- That is a closure problem, not stiffness, and no integrator fixes it.

**Summary.** ARS(2,2,2) conserves and extends the stable step for the TKE closures (a separate improvement, if you want it), but it does not fix MRF. For MRF the choice is still option 1 (the WRF profile, with WRF's stable-layer over-mixing) or option 2 (keep the shear K and limit MidPoint's step), and GABLS1 suggests the shear K has a large-step problem of its own in stable conditions.

hgopalan:
Following up on the question about #4039: I merged #4039 (de0de9560) onto this branch (e3fdb1431) and reran the same cases. It merges cleanly.

- **The helper has a height bug on stretched and terrain-fitted meshes**, which I have written up on #4039. With it, GABLS1 puts the first-cell K at the 300 m²/s cap.
- **With that fixed** (`zrel = max(zval, 1e-4)`), #4039 changes MRF only slightly and does not change the MidPoint behaviour:

| case | this branch | + #4039 (height fixed) |
| --- | --- | --- |
| neutral column, dz = 5 m: du/dz sign changes at dt = 5 / 20 / 60 s | 46 / 58 / 69 | 50 / 64 / 65 |
| neutral column, dz = 20 m: sign changes at dt = 60 / 120 s | 8 / 10 | 10 / 10 |
| GABLS1 anelastic, MidPoint dt = 60 s: max Kmv (explicit ≈ 2.7) | 10.9 | 10.95 |
| GABLS1 as shipped (compressible): max Kmv, u* | 2.65, 0.383 | 2.71, 0.377 |

This fits the picture. #4039 corrects the first-cell gradients, while the MidPoint mode lives in the middle of the PBL and comes from the shear-dependent K throughout the stable/neutral PBL. So #4039 is worth having for the surface gradients, but for MRF under MidPoint the choice is still option 1 (the profile in every regime) or option 2 (keep the shear K and limit the step).

AMLattanzi:
@hgopalan Thank you for all the valueable feedback. From all of these analysis, it is starting to look like a matter of `A-stable` vs `L-stable` for the time integration scheme. Your stabilization with the implicit solve in the last stage makes the method first-order **AND** `L-stable`. The `RK3` scheme in ERF is also `A-stable`, just like the `MidPoint` method, since it too converges to Crank-Nicolson diffusion. If the test fails with `RK3`, that is confirmation of sorts for the `A-stable` issue. If it works, we have something else going on.

hgopalan:
Ran it. RK3 fails too, which supports the A- vs L-stable reading.

MRF neutral column, compressible (acoustic substeps pinned at a fast step ≤ 2 s), 1 h from a compressible explicit spin-up, compared with compressible explicit at a small dt. Entries are du/dz sign changes inside the PBL:

| K inside a stable/neutral PBL | RK3 `vert_implicit_fac` | dz = 5 m, dt = 5 / 20 / 60 / 120 s | dz = 20 m, dt = 5 / 20 / 60 / 120 s |
| --- | --- | --- | --- |
| shear K (this branch) | `1 1 0` (default) | 46 / blows up (\|u\| 2170 m/s) / NaN / NaN | 0 / 0 / 8 / 10 |
| shear K (this branch) | `1 1 1` | 0 / 0 / 0 / 0 | 0 / 0 / 0 / 0 |
| profile in every regime | `1 1 0` | 0 / 0 / 0 / 0 | 0 / 0 / 0 / 0 |

- **Default RK3 behaves like MidPoint.** It has the same mode at dz = 20 m (8 and 10 sign changes at 60 and 120 s), and at dz = 5 m it is worse: at dt = 5 s the mode is there (Kmv 54 against 21.9), and from dt = 20 s it diverges.
- **A solve in all three stages (`1 1 1`) removes the mode everywhere.** At dz = 5 m and dt ≥ 60 s, K still grows (Kmv 74–85, wind error 0.2 m/s), but smoothly, with no grid-scale mode. That is the same large-step error the RK2 two-solve variant showed.
- **With the profile K, the default `1 1 0` is clean on both grids** and within 2e-3 m/s of explicit at dt = 120 s. So the damping is only needed because K follows the local shear.

A side note: the dt = 60 and 120 s runs at dz = 5 m finished with NaN in every field and exit status 0. `check_for_negative_theta` tests `rho <= zero` and `rhotheta <= zero`, which are false for NaN, so a NaN state passes. `!(rhotheta > zero)` (and the same for `rho`) would catch it.

hgopalan:
GABLS1 run anelastic with option 1 (the profile below the PBL top in every regime) on top of this branch's current head (e3fdb1431). This closes the one case I had not tested for option 1.

Deck: `Exec/CanonicalTests/ABL/mrf_stable_gabls` (9 h, stretched mesh), run with `erf.anelastic = 1`, FFT projection. Each setup is compared with its own explicit run at dt = 1 s. The table gives max Kmv / number of extrema of the Kmv profile in the lowest 30 levels / max |ΔU| (m/s) / max |Δθ| (K):

| setup | explicit dt = 1 s: max Kmv, u* | MidPoint dt = 5 s | dt = 20 s | dt = 60 s | dt = 120 s |
| --- | --- | --- | --- | --- | --- |
| this branch (shear K) | 2.64, 0.383 | 2.64 / 1 / 6.7e-5 / 1.7e-4 | 2.64 / 1 / 3.9e-4 / 8.6e-4 | **10.9 / 13 / 0.91 / 0.52** | **137 / 12 / 1.7 / 1.2** |
| option 1 (profile K) | 4.74, 0.421 | 4.74 / 1 / 2.4e-4 / 2.9e-4 | 4.74 / 1 / 1.0e-3 / 1.3e-3 | 4.74 / 1 / 3.2e-3 / 4.2e-3 | aborts at step 22 of 270 (ρθ < 0 in the first cell) |
| YSUNew | 2.27, 0.344 | 2.27 / 3 / 5.1e-4 / 1.2e-3 | 2.27 / 3 / 1.9e-3 / 4.8e-3 | 2.26 / 3 / 5.4e-3 / 1.7e-2 | 2.26 / 3 / 1.6e-2 / 4.3e-2 |

- **At dt = 60 s option 1 fixes the GABLS1 failure.** With the shear K, the Kmv profile breaks into 13 extrema and the wind and θ are off by 0.91 m/s and 0.52 K. With option 1 the profile stays smooth and the run is within 3.2e-3 m/s and 4.2e-3 K of its explicit reference.
- **At dt = 120 s** the shear K runs to the end but with max Kmv 137 and errors of 1.7 m/s and 1.2 K. Option 1 aborts early in the first cell; I have not looked into why. dt = 120 s is 96 % of the 125 s CFL-1 step ERF prints for this deck, and YSUNew survives it.
- **The cost of option 1 is the one already noted.** Its stable mixing is about 1.8x the current branch and 2.1x YSUNew (explicit max Kmv 4.74 against 2.64 and 2.27; u* 0.421 against 0.383 and 0.344). The same happens in the compressible run of this deck (4.72, 0.420), so it is a property of the WRF profile, not of the integrator.

Taken with the neutral-column and RK3 results above, option 1 is the one change that removes the mode under MidPoint and RK3 on every case I have tried up to dt = 60 s, at the price of more mixing in stable layers. If that price is acceptable, a docs line pointing stable cases to YSUNew would cover it.

hgopalan:
Follow-up on the one open item above, why option 1 aborts on GABLS1 (anelastic, MidPoint) at dt = 120 s. It is not the MRF formula. It is a first-cell instability from how the surface stress enters the implicit momentum solve, and it affects every closure.

**What happens.** The first-cell wind runs away from the first steps, well above the 8 m/s geostrophic wind: 8 → 12 → 21 → 32 → 40 m/s by step 4, and 76 m/s by step 10. The second cell flips sign every step. Once the first cell reaches about 75 m/s the advective Courant number is about 9, and the run aborts at step 22.
- YSUNew does the same: it peaks at 60 m/s at step 10, then settles to about 15 m/s by step 20. It survived dt = 120 s by recovering, not by being immune.
- It is not the stretched mesh: a uniform 4 m mesh (512 levels) does the same and aborts at step 19.

**Where the momentum comes from.** On the uniform 4 m mesh, the column-integrated ρu changes by only −30, −8, +5, 0 per step, while MOST's u* implies the surface should remove about 80 per step. So momentum is pulled from the levels above into the first cell, and the surface stress removes far less than u*².

**What controls it.** The ratio dt/dz₁, not the diffusion number dt·K/dz₁²:

| dt (s) | first cell dz₁ (m) | dt/dz₁ | dt/dz₁² | result |
| --- | --- | --- | --- | --- |
| 60 | 4 | 15 | 3.75 | clean |
| 120 | 8 | 15 | 1.9 | clean |
| 30 | 2 | 15 | 7.5 | clean |
| 120 | 4 | 30 | 7.5 | aborts at step 22 |
| 60 | 2 | 30 | 15 | first cell runs away (164 m/s by step 13) |

The 30 s / 2 m run has the failing case's dt/dz₁² but half its dt/dz₁, and it is clean. At start-up (u* = 0.70, U₁ = 8 m/s) the passing runs have a drag number D = dt·u*²/(dz₁·U₁) of 0.92 and the failing ones 1.84. YSUNew recovers because its u* falls to 0.34 (D ≈ 1.1); option 1's higher mixing keeps u* at 0.42 (D ≈ 1.6).

**In the code.** With a surface layer, the implicit momentum solve applies the MOST stress as a fixed flux on the right-hand side, with no drag term on the diagonal:
- `ERF_ImplicitDiff_N.cpp` L459: `RHS_a(i,j,klo) += Fact * tau(i,j,klo);`
- `ERF_ImplicitDiff_S.cpp` L490 (with `dz_inv`)
- `ERF_ImplicitDiff_T.cpp` (same form)

So the surface stress stays explicit even with `vert_implicit`, and at large dt/dz₁ it no longer controls the first cell.

So option 1's dt = 120 s abort is a separate, closure-independent limit, roughly dt·u*²/(dz₁·U₁) ≲ 1. The usual remedy is an implicit surface drag (C = u*²/|U| on the first-cell diagonal, as WRF's PBL solvers do). I am prototyping that now and will post what it does.

hgopalan:
Implicit surface drag prototype, as promised above. It keeps dt = 120 s alive but costs accuracy, so it is not a fix on its own.

**What I tried** (scratch switch, `ERF_ImplicitDiff_N.cpp` and `_S.cpp`, momentum bottom row with a surface layer). For u and v the row's `tau(klo)` terms cancel (`gfac = 1`), and the surface stress sits inside `face_data`, the explicitly updated momentum. So the prototype:
- removes it from the right-hand side (`RHS -= Fact * tau`);
- applies it implicitly as −C·u₁ⁿ⁺¹ on the diagonal, with C = |τ| / |u₁ without the stress| (floored at 0.1 m/s).

**Result.** GABLS1 anelastic, option 1, MidPoint, 9 h, compared with the explicit run at dt = 1 s (first-cell wind 3.70 m/s):

| run | outcome | first-cell wind | max \|ΔU\| (m/s) | max \|Δθ\| (K) |
| --- | --- | --- | --- | --- |
| dt = 60 s, no drag | clean | 3.71 | 3.2e-3 | 4.2e-3 |
| dt = 60 s, implicit drag | clean | 3.60 | 0.10 | 5.6e-3 |
| dt = 120 s, no drag | aborts at step 22 | – | – | – |
| dt = 120 s, implicit drag | completes, smooth K | 2.69 | 1.0 | 0.23 |

- **At dt = 120 s** the run no longer aborts, and the runaway to 76 m/s is gone. But the first steps still swing between −3 and 35 m/s, and the first cell ends 1 m/s off.
- **At dt = 60 s** it makes the first cell about 30 times less accurate (0.10 against 3.2e-3 m/s).

**Why it falls short.** MidPoint solves only in the first stage; the second stage still applies the surface stress explicitly over the full step. Making the stress implicit in one stage but not the other breaks the stage consistency, which is where the accuracy loss at dt = 60 s comes from. A proper fix would treat the surface stress consistently in both stages, which is an integrator-side change rather than a patch to the solver row.

**Where this leaves MRF.** Option 1 is clean under MidPoint up to dt = 60 s on every case above. The dt = 120 s limit on GABLS1 is the separate first-cell drag limit, roughly dt·u*²/(dz₁·U₁) ≲ 1, which any closure can hit. YSUNew survives it on this deck only because its u* is lower.

AMLattanzi:
@hgopalan Option 1 seems to be the choice (implementing now). This approach is consistent with WRF (over-mixing) and is stable. The crux of the issue appears to be amplification from the PBL model itself. Correcting such things seems to be more akin to creating a new hyrbrid PBL model, which I don't think belongs under the MRF label.

hgopalan (approved):
Thanks Aaron

PR #4053, opened 2026-09-18
Open BC: fix advection coverage at lateral corners and at internal z splits

### LINKED ISSUE #4052: [Bug]: Open BC advection kernels write corner cell twice when 2 perpendicular faces meet, and use ghost cell to do it
### Summary

Using open boundaries on adjacent sides can lead to values blowing up, especially if there are irregularities (terrain, turbulence) nearby
 A single, minimal call to AdvectionSrcForOpenBC_Tangent_Xmom shows a uniform flow (RHS should be ~0 everywhere) producing unexpected results at the corner, sourced entirely from one irrelevant value in that unfilled ghost cell.
I created a small unit test which should demonstrate the issue and a possible fix.

### Steps to reproduce

to test:
git clone https://github.com/RaymndH/ERF.git
cd ERF && git checkout upstream-openbc-fix
git submodule update --init Submodules/AMReX Submodules/googletest
cmake -S . -B build-unit -DERF_ENABLE_UNIT_TESTS=ON -DERF_ENABLE_MPI=OFF
cmake --build build-unit --target erf_unit_tests -j <N>
./build-unit/Tests/Unit/erf_unit_tests --gtest_filter='MinimalCornerDemo.*'

or

1. Create A 4-cell domain, with x Boundary Condition Open at the high end (i=3).
2. Fill everything real with a normal, uniform flow (u=1 everywhere in-domain) - physically, the RHS should be ~0 everywhere, which it correctly is at the ordinary cell i=1.
3. Leave the single out-of-domain ghost cell (i=4) at some garbage value - I used 1.0e6
4. Call AdvectionSrcForOpenBC_Tangent_Xmom once, on the box that (pre-fix) still includes the corner.
5. Look at the corner cell. I got -2.500005e+11

### Expected behavior

corner cell (i=3) = 0

Every cell's advection RHS, including corner cells where two perpendicular Open boundaries meet, should depend only on valid domain/boundary data — never on an unfilled out-of-domain ghost cell.

### Git commit hash

9929d94a96ba40db3b1640bfc06e679ed72c34e5

### Environment

OS: Linux 5.4.0
Compiler: GCC 13.1.0 (/usr/bin/g++-13)
Build: CMake, ERF_ENABLE_UNIT_TESTS=ON, ERF_ENABLE_MPI=OFF
AMReX submodule: e60cdc187 (tag 26.09)

### Relevant logs or output

```shell
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from MinimalCornerDemo
[ RUN      ] MinimalCornerDemo.PrintTheCorruptedValue

  ordinary in-domain cell (i=1) RHS: -0
  corner cell            (i=3) RHS: -2.500005e+11
  (uniform u=1 flow -- both 'should' be near zero)

[       OK ] MinimalCornerDemo.PrintTheCorruptedValue (0 ms)
[----------] 1 test from MinimalCornerDemo (0 ms total)
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 1 test.
```

asalmgren:
Fixes #4052.

Two bugs in which open-boundary cells get the open-BC advection treatment. Both are coverage bugs — which cells each kernel runs on — and both show up as a stencil differencing across an open boundary into an out-of-domain ghost cell that is never filled.

---

## 1. Corners: two kernels write the same cell (#4052)

Where two perpendicular open lateral boundaries meet, the corner cell was written by *both* directions' open-BC advection kernels. Those kernels assign rather than accumulate, so whichever launched last won — and it applied a stencil that differenced across the open boundary it was not treating, reading an unfilled out-of-domain ghost cell. With uniform flow the corner RHS came out ~1e11 instead of ~0, as reported in the issue.

Ownership of each open-boundary cell is now unique:

- A **boundary-normal momentum** owns the corner face in its own direction — that face lies on the open boundary, so the radiation condition is what belongs there — and the perpendicular tangential boxes are trimmed to match.
- **`w` and the cell-centered state** are tangent to every lateral boundary and so have no normal owner at a corner. `OpenBCTangentPatches` enumerates the four edges trimmed clear of the corners, plus the corners themselves tagged with *both* open sides, so the kernel differences across neither open boundary there.

`AdvectionSrcForOpenBC_Tangent_Zmom` and `AdvectionSrcForOpenBC_Tangent_Cons` now take `x_side` and `y_side` (an `OpenSide` enum) in place of a single `dir`. That is what lets a corner use the one-sided open form in both directions at once — the old single-`dir` signature could not express it.

The trim fires only where the box end actually coincides with the domain end. The domain-level open flags by themselves are not a sufficient test: the boxes handed to these routines are per-box or per-tile slabs whose ends usually lie in the interior, and trimming those would drop interior cells from the open-boundary treatment and make the answer depend on the grid decomposition.

## 2. Internal z splits: no kernel writes the face

`AdvectionSrcForMom` rebuilt its own `tbx`/`tby`/`tbz` from the cell-centered box and shrank `tbz` in z **unconditionally**, where the caller (`ERF_SlowRhsPre.cpp`) shrinks only at an end that actually reaches the bottom or top of the domain. For grids A = cells [0,7] and B = [8,15], A's local `tbz` became nodes [1,7] and B's [9,15] — node 8 is dropped by *both*, so no grid gave it open-BC treatment. The caller's `bxz` does include node 8, so the ordinary operator wrote it with the interior stencil, which at a lateral open boundary differences into the unfilled ghost cell. Same failure mode as above, surviving at every internal z split. Same family as #4043, #4044 and #4046.

The fix is simpler than patching the shrink. The open-BC kernels overwrite what the advection operators just wrote, so they must run on exactly the boxes those operators covered — so use the `bxx`/`bxy`/`bxz` already passed in instead of rebuilding them. `bx` becomes unused and is removed from the signature. This also drops a second, benign inconsistency: `surroundingNodes(tilebox)` overlaps at internal *tile* boundaries in x and y where `nodaltilebox` does not. The domain-end shrink moves into `ShrinkZmomBoxAtDomainEnds` so the policy lives in one place and can be tested.

Reviving these faces requires `Omega`, `ax`, `ay`, `az` and `detJ` to be valid one cell further in z at grid edges. That was checked:

- `Omega` — 1 ghost, `FillBoundary` in `ERF_SlowRhsPre.cpp` before the loop (with a comment saying it is exactly for vertical decomposition).
- `ax`, `ay`, `az`, `detJ` — 1 ghost (`ERF_MakeNewArrays.cpp`), `FillBoundary`'d in `make_J` / `make_areas`.
- Momentum arrays — the requirement is not new: `AdvectionSrcForZMom` already runs over the un-shrunk `bxz` with the same `k-1`/`k+1` footprint, so any grid whose `bxz` reaches an internal split already reads them there. The open-BC kernel's footprint is a strict subset.

---

## Files

| File | Change |
| --- | --- |
| `Source/Advection/ERF_Advection.H` | `OpenSide`, `OpenBCPatch`, `TrimOpenBCCorner` / `OpenBCCornerBox` / `OpenBCTangentPatches`, `ShrinkZmomBoxAtDomainEnds` |
| `Source/Advection/ERF_AdvectionSrcForMom.cpp` | normal kernels own the corner faces; tangential boxes trimmed; `w` moves to the patch loop; boxes no longer rebuilt from `bx` |
| `Source/Advection/ERF_AdvectionSrcForOpenBC.cpp` | `_Tangent_Zmom` / `_Tangent_Cons` take `x_side` + `y_side` |
| `Source/Advection/ERF_AdvectionSrcForState.cpp` | state moves to the patch loop |
| `Source/EB/ERF_EBAdvectionSrcForState.cpp` | same, for the EB path |
| `Source/TimeIntegration/ERF_SlowRhsPre.cpp` | uses `ShrinkZmomBoxAtDomainEnds`; drops the `bx` argument |
| `Tests/Unit/Advection/ERF_GTestOpenBCCorners.cpp` | new — corner ownership |
| `Tests/Unit/Advection/ERF_GTestZmomBoxDecomposition.cpp` | new — z-split coverage |

## Testing

`ERF_GTestOpenBCCorners.cpp` checks that the patches cover every open-boundary cell exactly once, that corners carry both sides, that normal and tangential momentum faces stay disjoint, and that all of this is invariant under how the grid is split (five decompositions, including splits in each direction and in all three at once).

`ERF_GTestZmomBoxDecomposition.cpp` checks that the shrink fires only at the domain ends, pins the gap the unconditional shrink leaves across the whole split plane, and checks that every interior w-face is covered for every decomposition.

Verified locally in a Release build with MPI on and the EB sources compiled: full unit suite 600 tests, 597 passed, 3 skipped (the pre-existing opt-in `DumpCurvesForPlotting` / `EvalFromCsv` cases, unrelated).

Scope note: this is box-logic verification plus unit tests. No full open-BC regression case on a z-decomposed grid was run to compare answers end to end.

## Not addressed here

`Source/EB/ERF_EBAdvectionSrcForMom.cpp` has no open-BC handling at all. That is pre-existing and outside the scope of this issue.


RaymndH:
Thank you! Verified locally (GCC 13.1.0, unit tests only, ERF_ENABLE_MPI=OFF): builds clean, full suite 580/580 pass, including all 8 of your new OpenBCCorners/ZmomBoxDecomposition tests.

PR #4068, opened 2026-09-21
Add station time-series output

asalmgren:
Adds erf.station_names: named sets of geographic points at which a chosen
set of variables is written to an ASCII time series, one file per station,
for comparison against meteorological tower and surface-station
observations.  This is the ERF analog of WRF's tslist; the existing line
and plane sampling writes whole lines and planes, in index or domain
coordinates, from a hand-coded list of nine variables.

    erf.station_names  = Lake1 Forests
    erf.Lake1.field    = rain_accum
    erf.Lake1.lat      = 45.13
    erf.Lake1.long     = -122.34
    erf.Forests.field  = magvel local_helicity
    erf.Forests.lat    = 45.20 45.41
    erf.Forests.long   = -122.10 -122.02
    erf.Forests.height = 10.0 80.0

lat and long are paired positionally; .x and .y place a station in domain
coordinates for a run with no lat/lon arrays.  Heights are metres above
the local terrain.  Values are bilinearly interpolated in the horizontal
and linearly in the vertical, from the finest level whose valid region
covers the whole interpolation stencil.

The variable names accepted are exactly the names that can be written to a
3D or 2D plotfile.  To keep that true as variables are added, the per-level
fill is split out of the plotfile writers -- FillPlot3DVars and
BuildPlot3DScratch out of Write3DPlotFile, FillPlot2DVars out of
Write2DPlotFile, canonicalizePlot3DVariables and appendPlot3DVariables out
of the selection code -- and the station sampler calls them, so a station
column and the corresponding plotfile component cannot disagree.  That
split is answer-preserving: every gold file is unchanged.

An unusable station stops the run naming the key at fault, rather than
silently dropping or moving a column; a 2D diagnostic that is valid but
not computed in this run is written as the missing value it would have in
a 2D plotfile.  A restart appends to the file the earlier run wrote.

Stations initialized from wrfinput de-stagger XLAT_V/XLONG_U to the mass
point, so a station lands on the cell its coordinates name.

A level supplies a station when it covers the stencil from the bottom of the
domain -- a height is measured from the local terrain -- up through the cells
the vertical interpolation reads, so a level that refines only the lower part
of the domain is used rather than rejected.  The sampler runs at every step by
default, so it does not do the microphysics average-down that the plotfile
writer does: turning station output on cannot change the answer.  The row
buffer is flushed with every checkpoint, a restart drops any rows the earlier
run wrote past the checkpoint it restarts from, and a restart into a file whose
header describes different columns stops the run.

Tests: a unit test of the local inverse of the lat/lon map, which is
exercised end to end only by a run initialized from a WRF or metgrid file.
No regression tests are added.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


AMLattanzi:
<html>
<body>
<html><head></head><body><h1>Review: ERF PR #4068 — "Add station time-series output"</h1>
<p><strong>Repo:</strong> erf-model/ERF · <strong>Author:</strong> asalmgren · <strong>Base:</strong> <code>development</code> · <strong>Head:</strong> <code>station-time-series</code>
<strong>State:</strong> open, <code>mergeable_state: unstable</code> · 16 files, +3527 / −1823
<strong>Reviewed at:</strong> commit <code>039631e</code> (feature commit), merge head <code>4ea422c</code></p>
<hr>
<h2>What the PR does</h2>
<p>Adds <code>erf.station_names</code>: named sets of geographic points at which a chosen set of variables
is written to an ASCII time series, one file per station — the ERF analog of WRF's <code>tslist</code>.
Points are given as paired <code>lat</code>/<code>long</code> (or <code>x</code>/<code>y</code> in domain coordinates), with <code>height</code> in
metres above local terrain. Values are bilinearly interpolated in the horizontal and linearly
in the vertical, from the finest level whose valid region covers the interpolation stencil.</p>
<p>To keep station columns and plotfile components from diverging, the per-level fill is split out
of the plotfile writers:</p>

Extracted | From
-- | --
BuildPlot3DScratch, FillPlot3DVars | Write3DPlotFile
FillPlot2DVars | Write2DPlotFile
canonicalizePlot3DVariables, appendPlot3DVariables | setPlotVariables / appendPlotVariables


<p>New files: <code>Source/IO/ERF_StationSampler.{H,cpp}</code>, <code>Tests/RunExpectAbort.cmake</code>,
<code>Tests/Unit/IO/ERF_GTestStationSampler.cpp</code>, <code>Tests/test_files/StationSampling/StationSampling.i</code>.</p>
<h2>Overall assessment</h2>
<p>The design is sound and unusually well documented — inputs, header, and Sphinx docs all agree,
and aborting on an unusable station (rather than dropping or shifting a column) is the right
policy. The plotfile refactor looks answer-preserving: diffed with <code>-w</code>, it is essentially
<code>mf[lev]</code> → <code>mf_dst</code> throughout, plus the hoisting of the scratch build. The one reordering —
the "requested but not available" warning loop moving after <code>appendPlot3DVariables</code> — is benign.</p>
<p>Four issues I'd want resolved before merge, then a set of smaller items.</p>
<p><em>Caveat: this is a read of the code only. Nothing was built or run, so items 1 and 4 especially
should be confirmed empirically.</em></p>
<hr>
<h2>1. Full-column coverage test likely defeats AMR level selection</h2>
<p>In <code>resolve_station_stencils</code>, <code>covered()</code> builds</p>
<pre><code class="language-cpp">Box column(IntVect(ii,jj,dom.smallEnd(2)), IntVect(ii,jj,dom.bigEnd(2)));
return grids[lev].contains(column);
</code></pre>
<p>This requires the level to cover the <strong>entire vertical extent of its own domain</strong> at that
<code>(i,j)</code>. With partial vertical refinement the condition can never be met.</p>
<p>That is exactly what this PR's own test deck sets up. <code>StationSampling.i</code> has
<code>amr.ref_ratio_vect = 2 1 2</code> and <code>box1.in_box_hi</code> z = 3200 out of a 6400 m domain, so <code>grids[1]</code>
spans k = 0..63 of a level-1 domain running 0..127. <code>grids[1].contains(column)</code> is therefore
false for every station, and all of them should fall back to level 0 — despite the deck's comment
that the Center stations "must come from level 1, and two heights, so the vertical interpolation
runs." <code>add_test_box_parity</code> would still pass, because both runs fall back identically.</p>
<p><strong>Suggested fix:</strong> require coverage only over the k range that brackets the requested heights,
and trim the stencil box in <code>sample_stations</code> (currently <code>klo..khi</code>) to match. <strong>To confirm:</strong>
print <code>loc.lev</code> under <code>erf.v</code> and check which level the Center stations resolve to.</p>
<h2>2. <code>BuildPlot3DScratch</code> runs every step and has side effects</h2>
<p><code>sample_stations</code> calls <code>BuildPlot3DScratch</code> at every output step, and the default interval is 1.
That routine is not a pure diagnostic:</p>
<ul>
<li><code>FillPatchCrseLevel</code> on level 0 plus <code>FillPatchFineLevel</code> on every finer level</li>
<li>re-points <code>qmoist[lev][mvar]</code> at the microphysics storage</li>
<li>for Lagrangian microphysics with <code>CouplingType::TwoWay</code>, calls <code>AverageDownMicroVars</code> and
<code>AverageDownMoistStateTo</code>, which <strong>modify the coarse solution</strong></li>
</ul>
<p>At plotfile cadence this was acceptable. Per-step it is a different proposition.</p>
<p><strong>Asks:</strong> confirm the average-down is idempotent, so that enabling station output cannot change
the answer; reconsider the every-step default, or document the cost. Also note
<code>BuildPlot3DScratch</code> builds scratch on <em>all</em> levels while <code>FillPlot3DVars</code> is only called on
levels that host a station. The dominant cost is filling every requested derived variable over
whole levels in order to read a handful of points.</p>
<h2>3. Restart is only correct from the previous run's last checkpoint</h2>
<p>Rows are buffered <code>station_buffer_steps</code> (default 100) and flushed at the end of <code>Evolve</code>, but
there is <strong>no flush in the checkpoint path</strong>. Two failure modes:</p>
<ul>
<li><strong>Gap.</strong> Checkpoint at step 150, crash at 180, restart from 150 → the file holds rows through
step 100 and then jumps to 151. Rows 101–150 are silently lost.</li>
<li><strong>Overlap.</strong> If the earlier run ran past the checkpoint and flushed, the restart appends rows
at times already in the file, and the series goes non-monotonic.</li>
</ul>
<p><code>StationSampling_Restart</code> covers only the clean case (restart from the run's final checkpoint).</p>
<p><strong>Suggested fix:</strong> flush in the checkpoint path to close the gap; for the overlap case either
truncate rows beyond the restart time or document/warn. Separately, nothing checks that the
station configuration on restart matches the header already in the file — changing
<code>erf.&lt;name&gt;.field</code> appends mismatched columns silently.</p>
<h2>4. Pinned-arena MultiFabs read on the host without an explicit sync</h2>
<p><code>st3d</code>, <code>st2d</code>, <code>stzc</code>, <code>stznd</code> and <code>latlon_all</code> are <code>The_Pinned_Arena()</code> MultiFabs that get
<code>setVal</code> + <code>ParallelCopy</code> — both device-side in a CUDA build — and are then read directly on the
host with no <code>Gpu::streamSynchronize()</code>. This pattern is new in ERF; the existing
<code>The_Pinned_Arena</code> uses are all FArrayBox NetCDF staging, not MultiFab <code>ParallelCopy</code>.</p>
<p><strong>Ask:</strong> confirm AMReX guarantees completion on return from <code>ParallelCopy</code> for the
local-copy-only path, or add an explicit sync before the host-side interpolation.</p>
<hr>
<h2>Smaller items</h2>
<ul>
<li><strong><code>erf.do_station_sampling = false</code> cannot disable station output.</strong>
<code>if (!do_station) { do_station = (pp.countval("station_names") &gt; 0); }</code> — naming stations always
wins. Either honor an explicit <code>false</code> or drop the key.</li>
<li><strong>Station names are used unvalidated as path components</strong> (<code>m_dir + "/" + name + ".dat"</code>).</li>
<li><strong>Heights below the first cell centre clamp</strong> to the first cell-centre value with no
similarity-based extrapolation. <code>.height = 10.0</code> in a 100 m-dz run quietly returns the lowest
cell value. Users wanting 2 m / 10 m should be pointed at the 2D diagnostics; the docs mention
the clamp but not the trap.</li>
<li><strong>Setup-time lat/lon gather is O(nx·ny) per location on the IO rank</strong>, with the whole level-0
lat/lon array gathered onto one rank. Fine for typical domains; for a large hurricane domain
that is 100+ MB on a single rank. Worth a comment on the expected ceiling.</li>
<li><strong>Destagger keys off <code>solverChoice.init_type == InitType::WRFInput</code>.</strong> On a restart where
<code>init_type</code> is not set the same way, every station shifts half a cell with no diagnostic.
Storing a "lat/lon are staggered" flag in the checkpoint would be safer.</li>
<li><strong>Misleading abort message:</strong> the <code>|di| &gt; 1 || |dj| &gt; 1</code> branch reports the point as outside the
domain, but it also fires for a poorly conditioned map near the requested point.</li>
<li><strong>CI:</strong> GitHub reports the PR as <code>unstable</code>, so something is not green. (Rate limits blocked
reading the individual check runs.)</li>
</ul>
<hr>
<h2>What's good</h2>
<ul>
<li><code>FillPlot3DVars</code> / <code>FillPlot2DVars</code> extraction is clean and appears answer-preserving; the PR
states every gold file is unchanged.</li>
<li>Abort-with-the-offending-key on bad input, plus <code>RunExpectAbort.cmake</code> testing the <em>message</em>
and not just the exit status.</li>
<li>2D diagnostics that are valid but not computed in a run are written as the missing value a 2D
plotfile would use, rather than dropped — consistent with existing convention.</li>
<li>Unit tests for <code>invert_latlon_offset</code> cover axis-aligned, rotated/sheared, and degenerate maps,
which is the right call given there is no regression deck initialized from WRF/metgrid.</li>
<li>Test deck deliberately exercises the collapsed stencil at a non-periodic boundary, the periodic
wrap, and the 2D path.</li>
<li>Docs are thorough and match the code.</li>
</ul></body></html>
</body>
</html>

AMLattanzi:
<html>
<body>
<html><head></head><body><h1>ERF PR #4068 — "Add station time-series output"</h1>
<p><strong>Reviewed at:</strong> <code>dd9cf3a</code>, diffed against merge-base <code>b4eda42</code> (12 files, +3823 / −1825)
<strong>Caveat:</strong> code read only. Nothing was built or run, so items 2 and 3 need empirical confirmation.</p>
<hr>
<h2>Must fix before merge</h2>
<h3>1. No integration test coverage at all</h3>
<p>Commit <code>b249ffc</code> removed the box-parity test, the restart-parity test, the five
input-validation abort tests, the expect-abort runner, and <code>StationSampling.i</code>.
What remains is a unit test of <code>invert_latlon_offset</code>, a nine-line inline
function in the header. <code>grep station_names Tests/ Exec/</code> returns nothing.</p>
<p>So roughly 1550 new lines merge with no coverage of level selection, the
periodic wrap, the restart seam, the header check, the 2D path, or any abort
path. The <code>de66b32</code> message reports that 138 regression tests and 639 unit tests
pass unchanged — true, but no deck enables station output, so that says nothing
about this feature.</p>
<p><strong>Ask:</strong> restore at least the box-parity deck (same series from a single box, a
split grid and 4 ranks) and the restart test, or explain what made them
unworkable.</p>
<h3>2. Answer-invariance is asserted but not demonstrated</h3>
<p>Dropping the microphysics average-down (<code>sync_solution = false</code>) was the
important half. But <code>BuildPlot3DScratch</code> still runs, every output step at the
default interval of 1:</p>
<ul>
<li><code>FillPatchCrseLevel(0, t_new[0], {cons, xvel, yvel, zvel})</code></li>
<li><code>FillPatchFineLevel(lev, ...)</code> for every level up to <code>top_station_lev</code></li>
<li>re-points <code>qmoist[lev][mvar]</code> on every level</li>
</ul>
<p>The qmoist re-point is idempotent. The fillpatches overwrite the ghost cells of
<code>vars_new</code> at end of step with a fill at <code>t_new</code>, which is answer-neutral only if
every consumer refills ghosts before reading them. Most suspect: runs with
time-dependent lateral boundary data, where ghost values are tied to a
particular <code>m_r2d</code> read, and the surface-layer path.</p>
<p><strong>Ask:</strong> run a case with AMR + moisture + WRF boundary data with and without
<code>erf.station_names</code> and diff the plotfiles bitwise. The docs make this claim in a
<code>.. note::</code>, so it should have evidence behind it.</p>
<h3>3. The restart header check is whole-file string equality</h3>
<p><code>checkHeaderMatches</code> compares every leading <code>#</code> line, including the trailing
prose paragraph about bilinear interpolation and the coordinates printed at
<code>datprecision = 10</code>.</p>
<ul>
<li>Any future edit to the header wording makes every existing station file
un-restartable.</li>
<li>Any floating-point difference in the resolved <code>x</code>/<code>y</code> aborts with a message
naming the wrong cause (it says the configuration changed; the real cause
could be a different <code>init_type</code>, different compiler flags, or a different
inversion path).</li>
</ul>
<p><strong>Ask:</strong> compare a compact machine-readable signature (format version + hash of
the column descriptions) instead of the prose; at minimum exclude the trailing
explanatory paragraph.</p>
<p><strong>Also:</strong> the check fires at the first <code>flush()</code> — first checkpoint, or 100 rows
in. A configuration mismatch is a setup error and belongs in <code>init_stations</code>,
before the run does any work.</p>
<h3>4. Destagger still keys off run-time <code>init_type</code></h3>
<pre><code class="language-cpp">const bool destagger = (solverChoice.init_type == InitType::WRFInput);
</code></pre>
<p>On restart, <code>lat_m</code>/<code>lon_m</code> come from the checkpoint (<code>ERF_Checkpoint.cpp:1465</code>)
and hold the staggered values. A restart deck that does not set <code>init_type</code> the
same way moves every station half a cell. The header check now catches this, but
reports it as an unrelated configuration change.</p>
<p><strong>Ask:</strong> store a "lat/lon are staggered" flag in the checkpoint, or derive it
from the presence of the staggered edge.</p>
<hr>
<h2>Optional fixes (not blocking)</h2>

Item | Where | Note
-- | -- | --
Per-step stencil resolution | resolve_station_stencils | Full ParallelCopy + streamSynchronize + Bcast every output step whether or not the grids moved. Cache on a regrid counter.
Velocity all-or-nothing | canonicalizePlot3DVariables | Asking for x_velocity fills all three over whole levels. Belongs in the cost note, since it is the obvious tower request.
Sampled-level 2D vars unreachable | sample_stations | Descriptors built with an empty definition list. Docs say accepted names are "exactly the names that can be written to a 2D plotfile", which overstates it; such a name aborts with a confusing message.
Level switching at a regrid | docs | A station moving between levels steps discontinuously. Belongs next to the clamping warning, given the observation-comparison purpose.
In-place rewrite | truncate_station_file | ios::trunc on the same path, no temp-file-and-rename. A crash mid-rewrite loses the series.
t_cut arithmetic | truncate_station_file | t_first*(1-1e-12) - 1e-12 is negative at t_first == 0. Unreachable today; fragile way to say "strictly before".
Hardcoded k = 0 | sample_stations, bl_2d | Neighbouring boxes use klo. Same value in practice, inconsistent to read.
z_phys_nd without z_phys_cc | resolve_station_stencils, sample_stations | z_of_k falls back to uniform dz while z_surf uses terrain. Probably unreachable; an assert is enough.
Duplicate stencil boxes | gather BoxArray | Two locations in one cell produce identical boxes. AMReX handles it, but it is untested and easy to hit with a real network.
Visibility | ERF.H | init_stations, resolve_station_positions, resolve_station_stencils, sample_stations, flush_stations and the FillPlot* helpers are all public; they look like private output machinery.
CI | — | Reported mergeable_state: unstable. Worth reading the individual check runs.


</body>
</html>

Issue #4062, opened 2026-09-19
Follow-up items from the review of #4038 (surface-temperature contract)


asalmgren:
Follow-up items from the review of #4038 ("Normalize surface temperature handling across SurfaceLayer and radiation"), filed here rather than blocking the merge. Line numbers refer to the merged state of that PR (head `d895a00`).

All twelve findings from the two earlier review rounds on #4038 were verified as genuinely addressed on that head; what follows is new.

### A. Indeterminate `psurf` when `metgrid_surface_pressure()` fails (GPU release builds)

`Source/Initialization/ERF_InitFromMetgrid.cpp:1692-1712`

```cpp
Real psurf;                                     // :1692, uninitialized
...
const bool valid_psurf = metgrid_surface_pressure(
    metgrid_debug_psfc, flag_psfc, file_psfc, z_sfc, P00, T00, TLP, psurf);
AMREX_ALWAYS_ASSERT(valid_psurf);               // :1711
AMREX_ALWAYS_ASSERT(psurf > zero);              // :1712
```

`metgrid_surface_pressure` (`Source/Initialization/ERF_MetgridUtils.H`) writes `pressure` only on its success paths, and this call site is inside an `AMREX_GPU_DEVICE` lambda. `AMREX_ALWAYS_ASSERT` expands to `amrex::Assert` (`Submodules/AMReX/Src/Base/AMReX_BLassert.H:50`), which is compiled out **on the device** under `NDEBUG` without `AMREX_USE_ASSERTION` (`Submodules/AMReX/Src/Base/AMReX.H:363-374`) — that is, in every release GPU build.

Failure scenario: a met_em file whose `PSFC` carries a `_FillValue` or `missing_value`. #4038 newly canonicalizes those to NaN in `canonicalize_metgrid_psfc`, so the helper returns `false`, both asserts are no-ops on device, and the hydrostatic column integration below builds `rho`, `p_hse`, `r_hse` and `th_hse` for the entire column from whatever was on the stack — silently. The same applies when `flag_psfc == 0` and `z_sfc`/`TLP` fail the new validation.

Before #4038 this path was safe: `psurf` was assigned unconditionally on all three branches (see `development` prior to the merge, lines 1638/1640/1644), so an invalid PSFC produced an obviously wrong 9.97e36 Pa that at least tripped `psurf > zero` on the host.

CPU builds still abort correctly, so this is GPU-release-only.

[...]

PR #4069, opened 2026-09-21
Fix surface-temperature contract follow-ups and GPU Metgrid pressure handling

pressel:
## Summary

This PR addresses the follow-up correctness issues identified in #4062 after the surface-temperature contract changes in #4038. The patch is intentionally focused on deterministic input validation, explicit unsupported EB surface-temperature configurations, checkpoint provenance, and configured thermodynamic constants.

### Deterministic Metgrid initialization

- Make Metgrid base-state failure propagation deterministic and GPU-safe: invalid surface-pressure results are initialized explicitly, device-side failures are recorded through a safe flag, and the host aborts before hydrostatic-equilibrium construction can consume an invalid pressure.
- Canonicalize missing and default-fill values for Metgrid `PSFC`, `TT`, and `PRES` before validation and hydrostatic initialization.
- Validate the complete atmospheric column for finite, positive potential temperature before HSE construction.

### Explicit EB surface-temperature contract

Reject unsupported EB configurations before entering the SurfaceLayer path:

- planar SST/TSK providers;
- coupled SST;
- LSM `t_sfc` providers;
- LSM flux providers;
- text/file surface forcing;
- custom or file-based roughness providers.

LSM fluxes are rejected under EB because their provider is defined on a planar `k=0` surface, while the consumer may need values on arbitrary three-dimensional EB cut-cell surfaces; ERF has no mapping that can make that transfer contract-safe.

Supported native EB SurfaceLayer configurations remain unchanged.

### Checkpoint provenance and legacy compatibility

- Record and use initialization provenance from `job_info` when validating markerless legacy checkpoints containing `SST_0` or `TSK_0`.
- Reject legacy Metgrid checkpoints containing those arrays because their old values are absolute temperature under the legacy contract.
- Keep the current `surface_temperature_contract` marker path unchanged.
- Fail closed for ambiguous provenance rather than guessing the stored representation.
- Provide the explicit `erf.legacy_surface_temperature_init_type = WRFInput` escape hatch only for old provenance-less WRFInput checkpoints. This escape hatch does not claim that every old WRFInput checkpoint is readable.

### RRTMGP thermodynamic constants

Use the configured `rdOcp` consistently when converting RRTMGP heating rates back to potential-temperature tendencies. The default configuration remains unchanged.

## Tests

Focused and broad coverage includes:

- Metgrid surface-pressure validation, including deterministic invalid-input handling and device execution;
- missing/default-fill canonicalization for `PSFC`, `TT`, and `PRES`;
- EB SurfaceLayer compatibility and rejection cases;
- checkpoint provenance and legacy surface-temperature compatibility;
- RRTMGP conversion with a non-default `rdOcp`;
- multi-time Metgrid forcing and surface-temperature normalization.

Local CPU validation after merging current `upstream/development` included the complete serial unit suite, the registered MPI parallel matrix, and the full configured CPU regression label. GPU and RRTMGP CI remain the authoritative validation for those configurations.

Addresses #4062


asalmgren:
## Verification performed

Worktree at `374bc578a`, GCC/MPI Release, NetCDF on:

- **Build clean**: 0 warnings, 0 errors.
- **`erf_unit_tests`**: 648 passed, 3 skipped (the three pre-existing `DumpCurvesForPlotting`/`EvalFromCsv` skips). All 22 new/touched tests in the four modified suites pass.
- **`ctest -L regression`**: 138/138 passed. No gold files need regeneration — `solverChoice.rdOcp` defaults to `RdoCp`, so the RRTMGP change is a no-op at default settings, and the other three changes are validation-only.
- **Docs render**: `Checkpoint.rst` and `SurfaceLayer.rst` parse under docutils with no malformed-table errors (only the expected unknown `toctree`/`:ref:` Sphinx-isms).
- **New runtime inputs**: none. The branch adds no `pp.query`/`pp.get` keys, so nothing is owed to `Inputs.rst`.

## Status of the four #4062 items

**A — indeterminate `psurf` (GPU release): genuinely fixed.** `metgrid_surface_pressure` now assigns NaN on entry so every failure path leaves a defined value; the call site sets a `Gpu::DeviceScalar<int>` flag and returns instead of relying on a device assert; `init_from_metgrid` syncs, `ReduceIntMax`es across ranks, and aborts before anything downstream runs. The early `return` is the last statement of the lambda, so no partially-built column is consumed. `std::numeric_limits<Real>::quiet_NaN()` in device code is fine — AMReX passes `--expt-relaxed-constexpr` in both the CMake and GNUmake CUDA paths.

**B — unreachable EB carve-out: resolved by rejection, but over-broad.** I re-confirmed the underlying bug: under EB, `ba_flux = ba` (3-D) with `ng = {1,1,1}` (`Source/BoundaryConditions/ERF_SurfaceLayer.H:641-644`), so `growntilebox().smallEnd(2) == klo-1` and the four `growntilebox`-based fills never execute. The PR takes the "assert unsupported" option, which is legitimate. See concern 2 below about `m_has_lsm_fluxes`.

**C — guard keyed on runtime input: hole closed, with a new compatibility cost.** The scan is now content-driven and provenance is read from the checkpoint's own `job_info`. I verified the format assumption holds: `writeJobInfo` calls `ParmParse::dumpTable(os, true)`, and both the current AMReX (`pp_to_pretty_string`) and the pre-2024 `PP_entry::print` emit `erf.init_type = Metgrid` / `= "Metgrid"`, which the new parser handles, including the `\r` from CRLF files. See concern 1 below.

**`rdOcp` note: fixed.** `kokkos_buffers_to_mf` hoists `m_rdOcp` into a local before the `MFIter` (so the lambda doesn't capture `this`) and routes through the new `rrtmgp::inverse_exner`.

## Concerns

**1. Markerless checkpoints written before the provenance work cannot be restarted at all.** `job_info` has only been written into checkpoints since #3433 (`c6792fd95`, 2026-07-12). Any checkpoint older than that has no `job_info`, so `parse_legacy_init_type_from_job_info` is never called, the type stays `Unknown`, and a checkpoint containing `SST_0`/`TSK_0` aborts with `UnknownProvenance`. That includes legacy **WRFInput** checkpoints, whose SST/TSK were already stored as potential temperature and which restart fine on `development` today — `Source/Initialization/ERF_InitFromWRFInput.cpp:1069` populates `sst_lev`, so they do carry those arrays. The PR description's "Legacy WRFInput checkpoints remain readable" is only true for the ~2-month window between #3433 and #4038; `Checkpoint.rst` states the rule correctly, but the user-facing consequence is that there is no way out — no input key, no override, nothing you can edit in the checkpoint. I'd add an explicit opt-in escape hatch (e.g. `erf.legacy_surface_temperature_init_type = WRFInput`) so a user asserting provenance can restart, and document it alongside the rejection.

**2. The EB rejection also covers `m_has_lsm_fluxes`, which was not broken.** The three t_surf providers (`fill_tsurf_with_sst_and_tsk`, `get_lsm_tsurf`, `fill_tsurf_with_coupled_sst`) all use `growntilebox` and were the silent no-ops #4062 described. But `compute_sfc_params_from_lsm_fluxes` (`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:1609-1610`) keys on **`validbox`**, whose `smallEnd(2) == klo` for the ground-touching box under EB — so that path *was* running. Rejecting it is a new restriction on working configuration, not a codification of an existing one. No in-tree deck combines EB with LSM (I checked all twelve EB inputs files), so nothing in CI breaks, but if the rejection is intentional it deserves a sentence in the PR description and in `SurfaceLayer.rst` explaining that LSM *fluxes* under EB are being withdrawn, not merely documented as already-broken.

**3. The assert one line below the fix has the same defect.** `Source/Initialization/ERF_InitFromMetgrid.cpp:1729`: `AMREX_ALWAYS_ASSERT(new_data(i,j,0,RhoTheta_comp) > zero);` is still a device assert inside the same `AMREX_GPU_DEVICE` lambda, compiled out in every release GPU build — exactly the failure mode A was filed about. A met_em file with a fill-valued `TT` would sail past it and build the column from NaN. It is one line to route it through the `surface_pressure_failed` flag that the PR just added (renaming the flag to something like `column_init_failed`).

**4. Dead code left behind.** Now that EB aborts, the `m_terrain_type != TerrainType::EB &&` clauses at `Source/BoundaryConditions/ERF_SurfaceLayer.cpp:1792` and `:2117` are doubly unreachable and still read as if EB were handled there. They should go, since removing them is what makes the new abort self-documenting.

**5. Related gap the PR doesn't close.** `read_custom_roughness` (`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:2562-2564`) uses the same `growntilebox` pattern against `z_0[lev]`, which is defined on `ba_flux`. So `most.roughness_file_name` is *also* silently ignored under EB. Not a regression and not in scope, but given that this PR's stated purpose is making EB SurfaceLayer limitations explicit, it's the one remaining member of that family.

**6. `Closes #4062` will drop five open notes.** The issue's "Notes" section still contains the coupled-SST coverage-mask behavior change, the PSFC-read-at-every-forcing-time asymmetry, the albedo/emissivity clamp removal, `fill_qsurf_with_qsat` reading the moisture component unconditionally, and the `noahmp_active`/`has_lsm_t_sfc` test asymmetry. Either drop the `Closes` keyword or open a successor issue.

## Minor

- The Copilot finding about a null `job_info_chars.dataPtr()` is **not** a real bug: `ReadAndBcastFile` always resizes to `fileLength+1` and writes a terminating `'\0'`, and the early `return` that leaves the buffer empty only happens when `bExitOnError` is false (the default is `true`, so an unopenable file aborts in `FileOpenFailed`). No change needed.
- The new read duplicates work: `ReadCheckpointFile` already reads the same `job_info` ~50 lines earlier via `erf_provenance::read_job_info_file`, which does the IO-rank-only read plus broadcast rather than a per-rank `FileExists` guarding a collective. Reusing it (or having it hand back the raw text) would be cheaper and removes the theoretical rank-divergence hazard. Low priority — it matches the marker-check pattern already in the file.
- `classify_legacy_surface_temperature_checkpoint(false, true, legacy_init_type)` hardcodes its first two arguments at the only production call site, since both cases are handled by early returns above. The helper's extra parameters exist only for the unit tests.
- `planar_sources_supported_for_terrain`'s continuation lines are indented one column past the open paren.

Nothing here is a correctness bug in the new code — the three original findings are all addressed and the numerics are untouched. Items 1 and 2 are backward-compatibility decisions worth settling before merge; 3 through 6 are follow-ups.

Issue #4060, opened 2026-09-19
Follow-ups from review of #4011: inert LSM wrfinput mapping, unreliable first-step radiation guard, test coverage


asalmgren:
Follow-ups from the review of #4011 (multilevel refinement enhancements for the WRF-input pathway). The mechanical fixes from that review were pushed onto the PR branch; the three items below need a decision from the author rather than a patch, so they are collected here.

### 1. The LSM surface fields are read from a mapping that nothing populates

`MakeNewLevelFromCoarse` sizes `lsm_data[lev]` / `lsm_data_name` and calls `lsm.Define()` before the surface-only wrfinput read, with the comment that this is so the read can populate them (`Source/ERF_MakeNewLevel.cpp:505-515`). But `lsm.Define()` is an empty function for both land-surface models:

- `Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP.H:70`
- `Source/LandSurfaceModel/SLM/ERF_SLM.H:38`

The structures the read actually indexes — `LsmDataMap`, `LsmDataName`, `lsm_fab_data` — are sized in `NOAHMP::Init` (`Source/LandSurfaceModel/Noah-MP/ERF_NOAHMP_Init.cpp:55-110`), which this path deliberately defers until after `FillCoarsePatch`. So at the time of the read, `lsm.Get_DataIdx()` iterates `LsmDataName[idx]` over an empty `amrex::Vector` and `lsm_data[lev][lsm_idx]` is still `nullptr`.

That does not fire today only because `Lsm_WRFInputNames()` returns an unpopulated `wrfinput_map` — nothing in `Source/` ever fills it — which makes the LSM branch of the read dead code. The practical consequence is that **the soil and vegetation fields the PR describes as being read from `wrfinput_d02` (TSLB, SMOIS, SH2O, LAI, ZS, DZS, VEGFRA, TMN, SHDMIN, SHDMAX) are not in fact read.** Note `init_from_wrfinput` already prints a warning about this when LSM is active.

Two ways out, and it is the author's call which is intended:

- If the LSM fields are meant to come from the file, populate `wrfinput_map` and move `lsm.Init` ahead of the read so the destination MultiFabs exist.
- If they are not, drop the pre-read `Define` block and the LSM entries from the surface-only variable list, so the code stops claiming to read something it does not.

### 2. `istep[lev] == 0` is not a reliable "first step after this level was created" test

[...]

PR #4070, opened 2026-09-21
Fix radiation interpolation crash with interp_atmos_from_coarse (#4060)

adamwise95:
## Description

Fixes radiation interpolation crash and clarifies LSM initialization comments for the multilevel wrfinput pathway with `interp_atmos_from_coarse`.

### Bug Fix: Radiation Interpolation
- **Issue**: Commit 0a1b3532 removed temporary MultiFabs for radiation flux interpolation to improve performance, but this caused assertion failures in `InterpFromCoarseLevel` when using `interp_atmos_f
rom_coarse` because the coarse `rad_fluxes` lacks ghost cells in x/y directions.
- **Fix**: Restored temporary MultiFabs with ghost cells for both main `rad_fluxes` and TOA plane interpolation in `ERF_AdvanceRadiation.cpp`. Added `FillBoundary` calls to properly populate ghost cells
 before interpolation.

### Documentation: LSM Initialization
- Clarified comments in `ERF_MakeNewLevel.cpp` to explain that NoahMP reads its own soil/vegetation state from wrfinput via the Fortran noahmpio library
- The `wrfinput_map` framework hook has never been populated and is not needed for current LSM implementations

## Testing
- Tested multilevel runs with delayed wrfinput_d02 start (surface-only initialization path)
- Tested multilevel runs with same start time (correctly ignores interp_atmos_from_coarse)

## Related Issues
- Addresses all items in #4060
- Related to PR #4011 (original multilevel enhancements)


Source/TimeIntegration/ERF_AdvanceRadiation.cpp
```diff
@@ -92,9 +92,20 @@ void ERF::advance_radiation (int lev,
         if (rad_fluxes[lev] && rad_fluxes[lev-1]) {
             const int nc = rad_fluxes[lev]->nComp();
 
+            // Create temporary coarse MultiFab with ghost cells for safe interpolation.
+            // rad_fluxes is defined with IntVect(0,0,1) ghost cells (the z ghost is for TOA data),
+            // but InterpFromCoarseLevel needs ghost cells in x/y to perform interpolation near
+            // grid boundaries. Use a temporary with uniform ghost cells like the LSM code does.
+            MultiFab tmp_coarse(rad_fluxes[lev-1]->boxArray(), rad_fluxes[lev-1]->DistributionMap(),
+                                nc, IntVect(1,1,1));
+            MultiFab::Copy(tmp_coarse, *rad_fluxes[lev-1], 0, 0, nc, 0);
```
asalmgren: **`tmp_coarse`'s ghost cells are read by the interpolation, but are never initialized.**

This is the flip side of the mechanism that makes the fix necessary. Because the overload at `AMReX_FillPatchUtil_I.H:1295` builds a `PhysBCFunctUseCoarseGhost`, `AMReX_FillPatchUtil_I.H:1087-1089` sets `send_ghost = fbc.cghost` and `recv_ghost = fbc.src_ghost` -- so

```cpp
mf_crse_patch.ParallelCopy(cmf, scomp, 0, ncomp, send_ghost, recv_ghost, ...)
```

reads `tmp_coarse`'s **ghost** region, here with `send_ghost = (1,1,1)`.

But `tmp_coarse` is freshly allocated (garbage, or snan under `amrex.fpe_trap_invalid`), this `MultiFab::Copy` has `nghost = 0` so it writes only the valid region, and `FillBoundary` only fills ghosts backed by another box's valid data. Two regions therefore stay uninitialized and are then fed to `cell_cons_interp`'s slope stencil:

- ghosts outside the coarse physical domain, at any non-periodic lateral boundary where a fine grid abuts the domain edge;
- the z ghosts at `klo_c-1` / `khi_c+1`. `src_ghost` is `(1,1,2)` here, and `CellConservativeLinear::CoarseBox` grows in z whenever `ratio[2] > 1`, which is supported -- e.g. `Exec/RegTests/ParticleTests/inputs_over_hill_AMR1_box_fullz` uses `amr.ref_ratio_vect = 2 1 2`.

The result is garbage or NaN in `rad_fluxes[lev]` (plotfiles and diagnostics) rather than a clean abort. A `tmp_coarse.setVal(0.0);` before the `Copy` closes it.

Source/TimeIntegration/ERF_AdvanceRadiation.cpp
```diff
@@ -92,9 +92,20 @@ void ERF::advance_radiation (int lev,
...
+            if (!rad[lev-1]->is_nested_patch()) {
+                tmp_coarse.FillBoundary(geom[lev-1].periodicity());
```
asalmgren: **This guard does not transfer to a temporary -- it should be unconditional here.**

The `!rad[lev-1]->is_nested_patch()` test makes sense on the pre-existing `qheating_rates[lev-1]->FillBoundary(...)` at line 82, where a nested patch's ghosts were already filled by its own `InterpFromCoarseLevel`. `tmp_coarse` is a brand-new fab whose halo is **always** uninitialized, regardless of nesting.

`is_nested_patch()` means only "this level does not reach the model top" (`Source/Radiation/RRTMGP/ERF_Radiation.H:77-78`) -- it says nothing about whether a halo exchange is needed or valid.

Concrete case: `max_level = 2` with levels 1 and 2 both nested patches. `advance_radiation(2)` reaches this lambda every step, `rad[1]->is_nested_patch()` is true, so `FillBoundary` is skipped entirely and the whole `(1,1,1)` halo is uninitialized when the `ParallelCopy` reads it.

Note also that `rad_fluxes[1]` itself has no x/y ghosts, so even an unconditional `FillBoundary` cannot define the nested patch's own lateral edge -- which is why the `setVal(0)` in my other comment is the load-bearing part.

Source/TimeIntegration/ERF_AdvanceRadiation.cpp
```diff
@@ -92,9 +92,20 @@ void ERF::advance_radiation (int lev,
         if (rad_fluxes[lev] && rad_fluxes[lev-1]) {
             const int nc = rad_fluxes[lev]->nComp();
 
+            // Create temporary coarse MultiFab with ghost cells for safe interpolation.
+            // rad_fluxes is defined with IntVect(0,0,1) ghost cells (the z ghost is for TOA data),
+            // but InterpFromCoarseLevel needs ghost cells in x/y to perform interpolation near
+            // grid boundaries. Use a temporary with uniform ghost cells like the LSM code does.
+            MultiFab tmp_coarse(rad_fluxes[lev-1]->boxArray(), rad_fluxes[lev-1]->DistributionMap(),
```
asalmgren: Two things on this block.

**Consider fixing it at the definition instead.** `Source/ERF_MakeNewArrays.cpp:525` defines `rad_fluxes` with `IntVect(0,0,1)`, while `qheating_rates` one line above (520) uses one ghost everywhere and consequently needs no temporary at all. Giving `rad_fluxes` an `IntVect(1,1,1)` halo satisfies the assert directly, lets the existing `setVal(zero)` at line 527 cover the ghosts, and removes both temporaries along with both classes of uninitialized-ghost bug. It also drops the per-step allocation + copy + halo exchange of a 4-component coarse-level MultiFab -- which for a nested patch runs on *every* step, and was the cost 0a1b3532 was trying to remove.

**The comment above points at code that does the opposite.** "Use a temporary with uniform ghost cells like the LSM code does" -- the LSM block at the bottom of this same lambda passes `*lsm_coarse` **directly**, with no temporary, because it uses `pc_interp`, whose `src_ghost` is `(0,0,0)`. Also, "InterpFromCoarseLevel needs ghost cells in x/y to perform interpolation near grid boundaries" is vague enough to leave the wrong mental model in place; naming the actual precondition (`AMREX_ALWAYS_ASSERT(cghost.allGE(src_ghost_outside_domain))` in `AMReX_PhysBCFunct.H:153`) would be more durable.

Source/TimeIntegration/ERF_AdvanceRadiation.cpp
```diff
@@ -125,8 +136,10 @@ void ERF::advance_radiation (int lev,
                     return BoxArray(std::move(bl));
                 };
 
+                // The TOA planes also need ghost cells for interpolation. Create temporaries
+                // with ghost cells, similar to the main rad_fluxes interpolation above.
                 MultiFab toa_crse(top_slab(rad_fluxes[lev-1]->boxArray()),
-                                  rad_fluxes[lev-1]->DistributionMap(), nc, 0);
+                                  rad_fluxes[lev-1]->DistributionMap(), nc, IntVect(1,1,0));
```
asalmgren: `IntVect(1,1,0)` is exactly right here, and worth recording why: with `rr2d = (rx,ry,1)`, `CellConservativeLinear::CoarseBox` grows only where `ratio[dim] > 1` (`AMReX_Interpolater.cpp:818-821`), so `src_ghost = src_ghost_outside_domain = (1,1,0)`. I checked that the pre-PR version with zero ghosts fails the same assert -- so this was a second, independent abort, not just a follow-on.

Same uninitialized-halo problem as `tmp_coarse` though: the `MFIter` loop below writes only `mfi.validbox()`, so the `(1,1,0)` halo needs a `toa_crse.setVal(0.0);` here before the copy loop.

Source/TimeIntegration/ERF_AdvanceRadiation.cpp
```diff
@@ -136,6 +149,9 @@ void ERF::advance_radiation (int lev,
                     toa_crse[mfi].template copy<RunOn::Device>((*rad_fluxes[lev-1])[mfi],
                                                                sbx, 0, dbx, 0, nc);
                 }
+                if (!rad[lev-1]->is_nested_patch()) {
+                    toa_crse.FillBoundary(geom[lev-1].periodicity());
```
asalmgren: Same as the `tmp_coarse` case: `toa_crse` is a fresh fab, so its halo is always uninitialized and this `FillBoundary` should not be gated on the parent being a nested patch.

asalmgren (commented):
The diagnosis is right, and both fixes are needed. I verified it rather than reading alone: ERF calls the AMReX overload at `AMReX_FillPatchUtil_I.H:1295`, which builds a `PhysBCFunctUseCoarseGhost`, whose constructor (`AMReX_PhysBCFunct.H:139-153`) does

```cpp
cghost(cmf.nGrowVect()) ... cghost.min(src_ghost);
AMREX_ALWAYS_ASSERT(cghost.allGE(src_ghost_outside_domain));
```

I linked a small probe against this repo's `libamrex` that replays that arithmetic for every call site in the lambda:

| call site | cmf ghosts | src_ghost | src_ghost_outside_domain | assert |
|---|---|---|---|---|
| `qheating_rates` | (1,1,1) | (2,2,2) | (1,1,1) | PASS |
| `rad_fluxes` as-is | **(0,0,1)** | (1,1,2) | (1,1,1) | **FAIL** |
| PR `tmp_coarse` | (1,1,1) | (1,1,2) | (1,1,1) | PASS |
| `toa_crse` before PR | **(0,0,0)** | (1,1,0) | (1,1,0) | **FAIL** |
| PR `toa_crse` | (1,1,0) | (1,1,0) | (1,1,0) | PASS |
| `lsm_coarse` (`pc_interp`) | (0,0,0) | (0,0,0) | (0,0,0) | PASS |

So the TOA plane was a second, independent abort (good catch); `(1,1,0)` is exactly the minimum there; and the reason `qheating_rates` never crashed is just that it carries one ghost everywhere. The `ERF_MakeNewLevel.cpp` hunks are comment-only and accurate -- `SLM::Define` and `NOAHMP::Define` are both empty bodies.

The flip side of that same mechanism is my main concern, and it is in the inline comments: because this overload sets `send_ghost = cghost` and `recv_ghost = src_ghost`, the `ParallelCopy` **reads the coarse source's ghost cells**, and neither new temporary ever initializes its halo.

### Two things I could not attach inline (outside the diff hunks)

**`ERF_AdvanceRadiation.cpp:69-74` -- the surviving NOTE comment now states the opposite of what the code does, and its claim is false.** It still reads:

> the coarse MultiFabs are handed to InterpFromCoarseLevel directly rather than being copied into a ghosted temporary first. InterpFromCoarseLevel builds its own coarse patch and fills it with a ParallelCopy that takes the source's VALID region only (send_ghost = recv_ghost = 0), so a temporary carrying ghost cells contributes nothing to the result

That is exactly the misreading of `PhysBCFunctUseCoarseGhost` that produced the crash this PR fixes, and it now sits 15 lines above code that does the opposite. Left in place it invites the same regression. Please rewrite it to say that this overload *requires* the coarse source to carry -- and to have filled -- `src_ghost` ghost cells. (For context, that NOTE came in with 0a1b3532.)

**`ERF_AdvanceRadiation.cpp:129` -- `top_slab` assumes coarse boxes span the full column** (pre-existing, but this PR reworks the block). It flattens *every* coarse box to `k = khi_c`, so a z-split coarse BoxArray yields duplicate boxes, and `sbx` at `khi_c+1` falls outside a fab whose z-range ends below the top. ERF's default `amr.no_box_split_dir = 2` prevents this, so it only bites under `no_box_split_dir = -1` -- but the new `FillBoundary` on that BoxArray would be ill-defined if duplicates ever appear. Filtering `top_slab` to boxes with `bigEnd(2) == khi_c` would close it.

### Notes

No gold files are affected. Nothing under `Tests/` exercises `interp_atmos_from_coarse`, and the path needs wrfinput NetCDF files, so it has no automated coverage at all -- this is the second crash here in two commits. A CI regression test may not be practical, but it's worth saying out loud.

PR #4071, opened 2026-09-21
Init KE

AMLattanzi:
# Summary
1. Provide a pathway for `WPS` and `Metgrid` to initialize `KE` for PBL models following the established prob name convention.
2. Use `z_agl` instead of `z` absolute so `KE` is properly initialized with terrain

## Changes
This should impact any problem that initializes `KE` with terrain.


Source/Prob/ERF_InitCustomPert_KE.H
```diff
@@ -11,16 +11,22 @@
...
+                                    + z_nd(i  ,j+1,klo) + z_nd(i+1,j+1,klo) );
+        const Real z_agl = z_cc(i,j,k) - z_sfc;
         if (state_pert.nComp() > RhoKE_comp) {
             state_pert(i, j, k, RhoKE_comp) = r_hse(i,j,k) * KE_0;
             // scale initial SGS kinetic energy with height
             if (KE_decay_height > 0) {
                 state_pert(i, j, k, RhoKE_comp) *= max(
-                    std::pow(1 - min(z/KE_decay_height,one_d), KE_decay_order),
+                    std::pow(1 - min(z_agl/KE_decay_height,one_d), KE_decay_order),
                     1e-12);
```
asalmgren: **Blocker: this breaks the single-precision builds.**

`std::pow` returns `Real`, but `1e-12` is a `double` literal, so `amrex::max` cannot deduce `T` when `Real = float`. This file was dead code until now, which is why it has never been caught. Reproduced standalone against this repo's AMReX with `Real = float`:

```
error: no matching function for call to 'max(float, double)'
note: candidate: 'template<class T> constexpr const T& amrex::max(const T&, const T&)'
note:   deduced conflicting types for parameter 'const T' ('float' and 'double')
```

`gcc.yml`, `cuda-ci.yml`, `hip.yml`, `sycl.yml` and `windows-mpi.yml` all build `mesh_precision: SINGLE`, so this fails CI.

The sibling headers already get this right -- `ERF_InitCustomPert_ABL.H:200` and `ERF_InitCustomPert_TurbulentInflow.H:111` both write `Real(1e-12)`. Same fix here.

Source/Prob/ERF_InitCustomPert_ABL.H
```diff
@@ -171,7 +177,17 @@
...
         // Set an initial value for SGS KE
+        //
+        // KE_decay_height is a depth of turbulence above the ground, so the profile
+        // is a function of height above the terrain surface, not of absolute z.  Over
+        // terrain higher than KE_decay_height the latter would clamp the decay factor
+        // to its floor at the ground -- exactly backwards.  On a flat non-EB mesh
+        // z_sfc is zero and this reduces to the absolute height.  Matches the surface
+        // height used by SurfaceLayer::init_tke_from_ustar.
         if (state_pert.nComp() > RhoKE_comp) {
+            const Real z_sfc = fourth * ( z_nd(i  ,j  ,klo) + z_nd(i+1,j  ,klo)
```
asalmgren: **The `z_agl` change is backwards for EB meshes** -- i.e. in the very direction the PR sets out to fix. (Same issue in `ERF_InitCustomPert_TurbulentInflow.H:103` and `ERF_InitCustomPert_KE.H:21`.)

For `erf.terrain_type = EB`, `init_default_zphys` (`Source/Utils/ERF_TerrainMetrics.cpp:36,46`) writes

```
z_nd = k*dz - z_offset
z_cc = (k+0.5)*dz - z_offset
```

with `z_offset = eb2.plane_point[2]`, and `ERF::init_zphys` returns early for EB (`Source/ERF_MakeNewArrays.cpp:880`) so these are never overwritten. So under EB, `z_cc` is **already** height above the EB surface, and `z_nd(i,j,klo) = -z_offset`. The new expression therefore gives

```
z_agl = z_cc - z_sfc = z_cc + z_offset
```

which is height above the *mesh* bottom -- exactly the quantity the PR is trying to stop using.

`Exec/CanonicalTests/ABL/inputs_GABLS1_EB` is `prob_name = ABL` with `eb2.plane_point = 0 0 63.28125`. It happens not to set `prob.KE_decay_height` today, so nothing breaks right now, but any EB run that does would get a profile shifted by 63 m. Suggest forcing `z_sfc = 0` when `terrain_type == EB`.

Minor, same block: the comment says "On a flat non-EB mesh `z_sfc` is zero" -- that holds only when `prob_lo[2] == 0`, since `init_default_zphys` ignores `prob_lo[2]` entirely.

Source/Prob/ERF_InitCustomPert_ABL.H
```diff
@@ -102,6 +102,12 @@
     }
 #endif
 
+    // The KE profile below is measured from the surface, so each box must reach
+    // the bottom of the domain for z_nd(i,j,klo) to be in scope.
+    const int klo = geomdata.Domain().smallEnd(2);
```
asalmgren: **This assert is unconditional, and rejects configurations ERF supports.** (Same at `ERF_InitCustomPert_TurbulentInflow.H:63`.)

It runs for every ABL / `constant_rhotheta_src` / turbulent-inflow init, regardless of whether a KE profile was requested -- `KE_decay_height` defaults to `-1`, in which case `z_agl` is computed and then never used.

Level 0 is fine in practice because `add_par` sets `amr.no_box_split_dir = 2` by default (`Source/main.cpp:45`). But two previously-working setups now abort at initialization:

1. A refined level whose grids do not reach the domain bottom. The message is also misleading there -- nothing was "vertically decomposed".
2. `amr.no_box_split_dir = -1`, documented at `Source/main.cpp:42` as a supported user override (`ERF::ReadParameters` only rejects it for implicit substepping).

Minimum fix: gate the check on `KE_decay_height > 0` (and on `state_pert.nComp() > RhoKE_comp`), so it only constrains runs that actually build the profile.

Better: handle it rather than forbid it. `SurfaceLayer::init_tke_from_ustar` (`Source/BoundaryConditions/ERF_SurfaceLayer.cpp:2450-2477`) already gathers the surface slab with a reduce+broadcast for exactly this case -- and it is the routine the new comment three hunks down cites as the reference for `z_sfc`.

Source/Prob/ERF_InitCustomPert_KE.H
```diff
@@ -11,16 +11,22 @@
...
 
+    int klo = geomdata.Domain().smallEnd(2);
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(klo == bx.smallEnd(2),
+        "Vertical decomposition with KE initialization is not allowed.");
+
     ParallelFor(bx, [=,one_d=one] AMREX_GPU_DEVICE(int i, int j, int k) noexcept
     {
         // Set an initial value for SGS KE
-        const Real z = z_cc(i,j,k);
+        const Real z_sfc = fourth * ( z_nd(i  ,j  ,klo) + z_nd(i+1,j  ,klo)
```
asalmgren: Minor: `z_sfc` and `z_agl` are computed above the `if (state_pert.nComp() > RhoKE_comp)` guard here, whereas the ABL and TurbulentInflow copies compute them inside it. Harmless today since `cons` always carries `RhoKE_comp`, but it diverges from the siblings for no reason.

Exec/ERF_Prob.cpp
```diff
@@ -195,6 +195,10 @@ Problem::init_custom_pert (
     else if (my_prob_name_ci == "bellforest") {
         // No state perturbation; uniform flow is set in init_custom_pert_vels
     }
+    else if (my_prob_name_ci == "wps"   ||
```
asalmgren, line 198: Minor, but worth a thought: this branch has no opt-out. `KE_0` defaults to `0.1`, so any run with `erf.prob_name = WPS` or `Metgrid` and a TKE-carrying model now gets `r_hse * 0.1` added to `RhoKE` through the entire column up to the model top, even when the user sets none of the `prob.KE_*` knobs.

For an idealized ABL that default is conventional; for a real-data WRF/metgrid run it is a silent and fairly large change to the initial TKE field. Consider keying the branch on `pp.query("KE_0", ...)` returning true, or on `KE_decay_height > 0`.

asalmgren (commented):
Reviewed the diff and built the branch (double precision builds clean). Two substantive issues plus a few smaller ones.

**Blocker:** `Source/Prob/ERF_InitCustomPert_KE.H` was previously included nowhere, so wiring it up at `Exec/ERF_Prob.cpp:198` compiles it for the first time -- and it does not compile in single precision. Details inline.

**Correctness:** the `z_agl` change is backwards for EB meshes, where `z_cc` is *already* height above the EB surface; and the new `AMREX_ALWAYS_ASSERT` is unconditional, so it aborts runs that never ask for a KE profile.

**Gold files:** none affected. `Exec/RegTests/WPS_Test/inputs_noah_rrtmgp` is not referenced by `Tests/CTestList.cmake`, and no regression test sets `prob.KE_decay_height` (only `CanonicalTests/ABL/inputs_GABLS1_{deardorff,mynn25}` and `Canonical_LES/Stable_ABL/inputs_with_const_massflux`), so the `z_agl` change moves no stored plotfile. Note the `ABL` -> `WPS` switch in that input also drops the ABL path's `RhoScalar = A_0*exp(-10 r^2)` perturbation -- presumably intended.

Also confirmed as safe: `z_nd` is used unguarded, but `z_phys_nd[lev]` is always allocated (`Source/ERF_MakeNewArrays.cpp:55`), so the `Array4<Real const>{}` fallback at `Source/Initialization/ERF_InitCustomPertState.cpp:62` is never taken; and for WRFInput/Metgrid the terrain is built before `init_custom` runs, so the new branch does see real terrain heights.

Source/Prob/ERF_InitCustomPert_KE.H
AMLattanzi: Use `amrex::Real` to correct in last commit.

Source/Prob/ERF_InitCustomPert_ABL.H
AMLattanzi: Boolean tests on if EB, if true the z_sfc value is set to zero.

Source/Prob/ERF_InitCustomPert_ABL.H
AMLattanzi: All asserts are now inside the decay height check

Exec/ERF_Prob.cpp
AMLattanzi: Default KE_0 to zero

Source/Prob/ERF_InitCustomPert_KE.H
AMLattanzi: Relocated inside the if block

